Issue #94, opened 2020-10-09
«JS-SDK» Framework built-in gas metering


hellwolf:
# Synopsis

Since we are using call proxies, `eth-gas-reporter` cannot report gas for each agreement functions anymore. We should build such feature in the framework as a debug option, and start to use it in our test suite.

# Spec

Make a option during framework initialisation to allow the gas report per each agreement function.

# ACs

- Use the framework in test suite
- New gas report with breakdown of gas usage grouped by agreement functions

PR #161, opened 2020-12-30
Feature 94 sdk gas metering

hellwolf:
closing #94


packages/js-sdk/src/utils/gasMetering/gasMetering.js
```diff
@@ -0,0 +1,118 @@
+const { BN } = require("@openzeppelin/test-helpers");
```
pi0neerpat, line 1: Is the idea that this is only to be used in development, never in production? If so we should add this to docs

packages/ethereum-contracts/test/TestEnvironment.js
```diff
@@ -73,8 +73,11 @@ module.exports = class TestEnvironment {
             ...deployOpts
         });
 
+        this.gasReportType = process.env.ENABLE_GAS_REPORT_TYPE;
```
pi0neerpat, line 76: We should add docs  for this, as well as adding to .env.template

packages/js-sdk/package.json
```diff
@@ -11,7 +11,7 @@
     ],
     "scripts": {
         "build": "scripts/build-abi-js.sh",
-        "test": "truffle test",
+        "test": "mkdir -p reports;truffle test",
```
pi0neerpat, line 14: do we need to add this to the ethereum-contracts tests too?

packages/js-sdk/src/ConstantFlowAgreementV1Helper.js
```diff
@@ -60,6 +60,7 @@ module.exports = class ConstantFlowAgreementV1Helper {
                 }
```
pi0neerpat, line 60: We also need to add to the IDA helper as well?

packages/js-sdk/src/Framework.js
```diff
@@ -169,6 +189,26 @@ module.exports = class Framework {
         );
         return this.contracts.ISuperToken.at(wrapperAddress);
     }
+    /**
+     * @dev call to add a tx in the gas report. Does nothing if gas report type is not set.
+     * @param {tx oject} tx as returned by truffleContract action
+     * @param {str} actionName action title for row in report
+     */
+    _pushTxForGasReport(tx, actionName) {
+        this._gasMetering ? this._gasMetering.pushTx(tx, actionName) : null;
```
pi0neerpat, line 198: I like this check pattern, which makes it much cleaner when adding the function  :+1:

packages/js-sdk/package.json
hellwolf: true, but perhaps the js code itself should rather try to create the dir instead... let me write that

packages/js-sdk/src/ConstantFlowAgreementV1Helper.js
hellwolf: not now. as mentioned above, just trying to close the ticket and merge the code instead of letting it dangling around.

packages/js-sdk/src/utils/gasMetering/gasMetering.js
hellwolf: answered above

pi0neerpat (approved):

PR #199, opened 2021-01-22
[examples] update flowlottery to v0.2.2

pi0neerpat:


pi0neerpat:
I need your opinions on some things  @hellwolf @kobuta23  I am trying to streamline the dev experience a bit.

TestEnvironment.js doesn't really belong in /test folder, since its not a test. Its not a script either, so maybe a new src folder in ethereum-contracts? Part of me just wants to move it to the sdk. But that would require bloating the sdk with all the build artifacts, which is the whole reason we separated them!

Right now, there is a big distinction between how I use tools to test on ganache, vs deploy on testnet/mainnet.

#### local ganache

To test, I write some code:

```js
// Deploy SF contracts
const t = new TestEnvironment(...

// Load the SDK
sf = new SuperfluidSDK.Framework(

// Deploy my app
app = await web3tx(LotterySuperApp.new
``` 

But now I want to deploy to ganache and point my App to the right contracts. So I write a deployment script that outputs the addresses I need.

#### goerli

Now I'm ready for testnet. But wait, I have to write a _new_ deployment script, since I don't want to deploy the full `TestEnvironment`. 

How can we simplify things so I don't need to write two deployment scripts? Is this worth pursuing?

hellwolf (approved):

PR #219, opened 2021-01-26
[js-sdk] Remove `ethereum-contracts` dep for load contracts

pi0neerpat:
Currently, using `new Framework()` requires that the `ethereum-contracts` package also be installed. This should not be the case, and is due to the `loadContracts.js` script that we missed during the monorepo migration.

### TODO

- [x]  Copy `loadContracts.js` into the sdk
- [x] Remove devDependency for `ethereum-contracts` 
- [x] Test for both truffle/non truffle environments

Marking as "Ready" only to access the published packages.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -6,6 +6,9 @@ module.exports = {
         es2020: true
     },
     plugins: ["prettier"],
+    globals: {
+        artifacts: "writable"
+    },
     rules: {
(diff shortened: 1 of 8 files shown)


hellwolf (approved):

pi0neerpat (commented):
:+1:  LGTM

Issue #208, opened 2021-01-25
PVE002: [SuperUpgrader] Accomondation of approve() Idiosyncrasies


hellwolf:
![image](https://user-images.githubusercontent.com/186660/105770763-50b1d000-5f68-11eb-9dd8-75941a38c09d.png)

hellwolf:
@ngmachado actually auditor looked at SuperUpgrader code too... they think it's better to approve(0) then approve(amount) later. But I think that revert would never happen. Have a look.

hellwolf:
@ngmachado also address the same issue stated in PVE003

PR #235, opened 2021-02-02
fix issue #208

ngmachado:


packages/ethereum-contracts/contracts/ux/SuperUpgrader.sol
```diff
@@ -58,8 +61,9 @@ contract SuperUpgrader is AccessControl {
         ISuperToken superToken = ISuperToken(superTokenAddr);
         //get tokens from user
         IERC20 token = IERC20(superToken.getUnderlyingToken());
-        token.transferFrom(account, address(this), amount);
-        token.approve(address(superToken), amount);
+        token.safeTransferFrom(account, address(this), amount);
+        assert(token.approve(address(superToken), 0));
```
hellwolf, line 65: we can use token.safeApprove instead

packages/ethereum-contracts/contracts/ux/SuperUpgrader.sol
```diff
@@ -58,8 +61,9 @@ contract SuperUpgrader is AccessControl {
         ISuperToken superToken = ISuperToken(superTokenAddr);
         //get tokens from user
         IERC20 token = IERC20(superToken.getUnderlyingToken());
-        token.transferFrom(account, address(this), amount);
-        token.approve(address(superToken), amount);
+        token.safeTransferFrom(account, address(this), amount);
```
hellwolf, line 64: calculate amount by the differences of balanceOf before and after

hellwolf (approved):

PR #236, opened 2021-02-02
[js-sdk] Add webpack build

pi0neerpat:
### Changes

- Adds `umd` webpack build on `yarn build` command. This should catch any browser issues during the workflows, since the build will fail.
- Removes `web3.js` from Framework
- Fixes `fs` and `path` issue in browser environment, by preventing loading

```
    "browser": {
        "fs": false,
        "path": false
    },
```

### Breaking changes

`web3Provider` should now be a `web3.js` object already instantiated, rather than relying on the sdk to create it.  Affects:

- superfluid-protocol-docs 
- SuperAuction
- 1 more private repo


packages/js-sdk/src/Framework.js
```diff
@@ -39,7 +37,7 @@ module.exports = class Framework {
         this.chainId = chainId;
         this.version = version || "test";
         this.resolverAddress = resolverAddress;
-        this.web3 = isTruffle ? global.web3 : new Web3(web3Provider);
+        this.web3 = isTruffle ? global.web3 : web3Provider;
```
pi0neerpat, line 40: Note this change

packages/js-sdk/test/Framework.test.js
```diff
@@ -99,13 +100,14 @@ contract("Framework class", accounts => {
...
         }
 
-        it("without truffle framework", async () => {
-            const sf = new SuperfluidSDK.Framework({
-                web3Provider: web3.currentProvider
-            });
-            await sf.initialize();
-            testLoadedContracts(sf);
-        });
+        // it("without truffle framework", async () => {
```
pi0neerpat, line 102: Intentionally commenting out this test, since using external web3 provider is going to be changed in #237

packages/js-sdk/src/Framework.js
hellwolf: kinda breaking change. I will already mark this version as 0.3.0

hellwolf (approved):

packages/js-sdk/.eslintrc.json
```diff
@@ -5,7 +5,8 @@
     },
     "globals": {
         "Atomics": "readonly",
-        "SharedArrayBuffer": "readonly"
+        "SharedArrayBuffer": "readonly",
+        "webpack": "readonly"
```
pi0neerpat, line 9: Can remove

packages/js-sdk/webpack.config.js
```diff
@@ -0,0 +1,37 @@
...
+module.exports = {
+    mode: "development",
+    entry: "./src/index.js",
+    output: {
+        path: path.resolve(__dirname, "dist"),
+        filename: "index.js",
+        library: "@superfluid-finance/js-sdk",
+        libraryTarget: "umd",
+        umdNamedDefine: true
+    },
```
pi0neerpat, line 14: Needs cleanuo

Issue #237, opened 2021-02-03
[js-sdk] Ethers support


pi0neerpat:
### Changes

Adds package, so we no longer depend on the web3Provider for`this.web3.utils.sha3`

- `@ethersproject/hash` 78.7 kB Minified

### Notes

`listFlows` function may eventually break in non-truffle environment, due to call limit?

```js
await this._sf.agreements.cfa.getPastEvents("FlowUpdated", {
                        fromBlock: 0,
                        toBlock: "latest",
                        filter: {
                            token: superTokenNorm,
                            receiver: accountNorm
                        }
                    })
```

### TODO

- [x] Parallelize initialization script (mostly tokens is the slow one)
- [ ] Enable ethers support via "adapter"
- [x] Add web3 provider "mode" flag: ["truffleNative", "truffleContract", "ethers"]

  - isTruffle will still work for "truffleNative"
  - no options specified defaults to "truffleContract"

- [ ] Fix failing non-native truffle environment test in `Framework.test.js`

- [ ] Update deployment scripts + tests with new arguments

pi0neerpat:
Adds package, so we no longer depend on the web3Provider for`this.web3.utils.sha3`

-  `@ethersproject/hash` 78.7 kB Minified

pi0neerpat:
Example React app usage

```js
import SuperfluidSDK from '@superfluid-finance/js-sdk';
import { Web3Provider } from '@ethersproject/providers';

React.useEffect(() => {
    test();
  }, []);
const test = async () => {
    const walletAddress = await window.ethereum.request({
      method: 'eth_requestAccounts',
      params: [
        {
          eth_accounts: {}
        }
      ]
    });
    const sf = new SuperfluidSDK.Framework({
      ethers: new Web3Provider(window.ethereum),
      mode: 'ethers'
    });
    await sf.initialize();
    const carol = sf.user({
      address: walletAddress[0],
      token: '0x2bf02814ea0b2b155ed47b7cede18caa752940e6'
    });
    const details = await carol.details();
    console.log(details);

    // Low level method
    await sf.host.callAgreement(
      sf.agreements.cfa.address,
      sf.agreements.cfa.contract.methods
        .createFlow(walletAddress[0], '0xA8f3447922d786045CB582B0C825723B744a54df', 386, '0x')
        .encodeABI(),
      '0x',
      {
        from: walletAddress[0]
      }
    );

    // High level
[...]

PR #238, opened 2021-02-03
[js-sdk] add ethers support

pi0neerpat:
Addresses #237 
Depends on #236


packages/ethereum-contracts/package.json
```diff
@@ -47,6 +47,7 @@
     },
     "devDependencies": {
         "@decentral.ee/web3-helpers": "0.3.5",
+        "@ethersproject/contracts": "^5.0.10",
```
hellwolf, line 50: do we check compatible ethers.js versions in Framework initialization?

packages/ethereum-contracts/scripts/deploy-test-environment.js
```diff
@@ -14,49 +14,49 @@ const deploySuperToken = require("./deploy-super-token");
...
         const tokens = ["fDAI", "fUSDC", "fTUSD"];
-        for (let i = 0; i < tokens.length; ++i) {
-            console.log(`==== Deploying test token ${tokens[i]}...`);
-            await deployTestToken(errorHandler, [":", tokens[i]], {
-                isTruffle,
-                web3Provider: this.web3.currentProvider,
-                from
-            });
-            console.log(`==== Test token ${tokens[i]} deployed.`);
+        await Promise.all([
```
hellwolf, line 35: This we shouldn't do.

Transactions from the same account should not be parallized, it will create nonce conflict and will likely fail.

packages/js-sdk/src/ConstantFlowAgreementV1Helper.js
```diff
@@ -247,7 +237,26 @@ module.exports = class ConstantFlowAgreementV1Helper {
         };
     }
 
-    static getLatestFlows(flows) {
+    async getFlowEvents({ token, receiver = null, sender = null }) {
```
pi0neerpat, line 240: Not sure how this was included here?

hellwolf: not sure, git blame?

packages/js-sdk/src/InstantDistributionAgreementV1Helper.js
```diff
@@ -20,16 +20,16 @@ module.exports = class InstantDistributionAgreementV1Helper {
...
+        const tx = this._sf.host.callAgreement(
+            this._ida.address,
+            this._ida.contract.methods
+                .createIndex(superToken, indexId, "0x")
+                .encodeABI(),
+            userData,
+            { from: sender }
+        );
+        if (tx.on) await tx.on("transactionHash", onTransaction);
+        else await tx();
```
hellwolf, line 32: PMed you, I don't understand this :D

hellwolf (changes requested):

packages/ethereum-contracts/package.json
pi0neerpat: Support for using ethers in the  ethereum-contracts deployment script is incomplete. I don't plan to release docs for using the scripts with ethers. 

Some of the changes I made to `ethereum-contracts` were required, and others are just getting prepared for when the time comes.

So no need to perform this check here

As far as performing this check in the js-sdk, I think we don't need to do it for now. If we see issues then we can start to think about it more

packages/ethereum-contracts/scripts/deploy-test-environment.js
pi0neerpat: Its currently not failing in truffle environment. Where would it fail? Only place I can think is if someone called it using MetaMask and had to manually sign each tx.

I'd prefer to keep it, if it reduces test time

packages/ethereum-contracts/package.json
hellwolf: okay

packages/ethereum-contracts/scripts/deploy-test-environment.js
hellwolf: this will definitely fail for testnet/mainnet deployments.

hellwolf (approved):
good stuff

PR #246, opened 2021-02-07
a bug in agreement library causing non super app callbacks in IDA case

hellwolf:
1. due to dividends rights token are no longer part of the test suite, this
regression was not detected.
2. IDA full test suite should cover this


ngmachado (approved):

hellwolf:
@ngmachado thanks!

PR #248, opened 2021-02-08
[examples] fix flowlottery demo

santteegt:
This PR introduces some changes to the `flowlottery` demo:

* Fixes error when a user tries to join the lottery: update the list of Txs sent through `batchCall` so they match to the recent release of Superfluid contracts
* Fixes an error when calling the `leaveLottery`: match the `callAgreement` method call to the recent release of Superfluid contracts
* Minor fixes in the UI so warnings/errors disappear from the console


hellwolf:
thanks @santteegt !

hellwolf (approved):

PR #258, opened 2021-02-08
[js-sdk] Ethers API provider issue with getSigner()

pi0neerpat:
Prevents `ethers.getSigner()` from throwing an error when using with `InfuraProvider`


hellwolf (approved):
ethers to ethers, dust to dust

PR #285, opened 2021-02-22
updated IDAHelper class

hellwolf:


hellwolf:
@pi0neerpat  please review the breaking changes in IDAHelper and its updated documentation!

packages/js-sdk/test/InstantDistributionAgreementV1Helper.test.js
```diff
@@ -312,23 +312,23 @@ contract("InstantDistributionAgreementV1Helper helper class", (accounts) => {
...
                 units: units.toString(),
-                sender: publisher,
             });
 
             await sf.ida.approveSubscription({
                 superToken: superToken.address,
                 indexId,
                 publisher,
-                sender: subscriber,
+                subscriber: subscriber,
```
pi0neerpat, line 331: if the sender is always the subscriber, then this is ok

hellwolf: yes it is.

The purpose of these changes is to keep all the parameters consistent. Only some operation would require "sender" parameter to tell from which account the transaction should be (including claim and deleteSubscription)

hellwolf:
@pi0neerpat ready

pi0neerpat (approved):
LGTM I can update docs this week

PR #296, opened 2021-03-01
use same upgrade/downgrade functions

hellwolf:


hellwolf:
@samparsky would this help that we don't need separated adapters?

samparsky:
@hellwolf yes it does, it makes it much simpler

ngmachado (approved):
LGTM

packages/ethereum-contracts/contracts/tokens/SETH.sol
```diff
@@ -38,7 +43,12 @@ interface ISETHCustom {
  * @author Superfluid
  */
 // solhint-disable-next-line no-empty-blocks
-interface ISETH is ISETHCustom, ISuperToken { }
+interface ISETH is ISETHCustom, ISuperToken {
+    function getUnderlyingToken() external override(ISETHCustom, ISuperToken) view returns(address tokenAddr);
+    function upgrade(uint256 amount) external override(ISETHCustom, ISuperToken);
+    function upgradeTo(address to, uint256 amount, bytes calldata data) external override(ISETHCustom, ISuperToken);
+    function downgrade(uint256 amount) external override(ISETHCustom, ISuperToken);
```
samparsky, line 50: Is there a reason for defining this here since it's already part of the `ISuperToken` interface?

packages/ethereum-contracts/contracts/tokens/SETH.sol
```diff
@@ -25,11 +25,16 @@ interface IWETH is IERC20 {
...
     function upgradeByETH() external payable;
     function upgradeByETHTo(address to) external payable;
-    function upgradeByWETH(uint wad) external;
     function downgradeToETH(uint wad) external;
-    function downgradeToWETH(uint wad) external;
+
+    // using wrapped native token
+    function getUnderlyingToken() external view returns(address tokenAddr);
+    function upgrade(uint256 amount) external;
+    function upgradeTo(address to, uint256 amount, bytes calldata data) external;
```
samparsky, line 36: I think it's possible to merge the ISETH and ISETHCustom interfaces? or is there a reason for having them separate?

packages/ethereum-contracts/contracts/tokens/SETH.sol
hellwolf: because SETHProxy only inherits the ISETHCustom

```
contract SETHProxy is ISETHCustom, CustomSuperTokenProxyBase {
```

packages/ethereum-contracts/contracts/tokens/SETH.sol
hellwolf: same as previous answer, the proxy conotract contains the "custom" token functions

PR #303, opened 2021-03-12
👨🏻‍🚀 Subraph initial setup

pi0neerpat:
:tada: The initial subgraph is live! (warning- its untested so be careful!)

https://thegraph.com/explorer/subgraph/pi0neerpat/superfluid-xdai

Notes:

- Subgraphs don't query the current state of the chain. There's no way to tell what the absolute truth is for a flow's `sum`, since this has changed since the last time the flow updated. I solved this problem already for rDAI using a wrapper utility, and plan to do the same here.
- Right now it only tracks CFA, but plan to add IDA later.
- What features would you like to see? Let me know!


pi0neerpat:
@hellwolf ready for review! Note that I changed the root package.json

packages/js-sdk/src/Framework.js
```diff
@@ -165,6 +165,23 @@ module.exports = class Framework {
...
+        this.superTokens = {};
+        await Promise.all(
+            [
+                ...(this._options.tokens ? this._options.tokens : []),
+                ...(this._options.loadSuperNativeToken &&
+                this.config.nativeTokenSymbol
+                    ? [this.config.nativeTokenSymbol]
+                    : []),
+            ].map(this.loadToken.bind(this))
+        );
```
pi0neerpat, line 184: I don't remember adding this. Did I accidentally add it back?

hellwolf: No i did, to speed up things more.

hellwolf (approved):
looks good. We shall sooner or later converge everything to our data api story, before everything get diverged away from each other too much.

PR #308, opened 2021-03-23
use node-jq package and improve error handling of tasks/build-abi-js

hellwolf:


hellwolf:
@pi0neerpat you should like this one

pi0neerpat (approved):
Excellent!

PR #317, opened 2021-04-18
Update and rename deploy-native-supertoken.js to deploy-native-super-…

d10r:
renamed `deploy-native-supertoken.js` to `deploy-native-super-token.js` in order to be consistent with the other deploy scripts.  
Note that the link in https://docs.superfluid.finance/superfluid/docs/super-tokens#deploy-a-custom-super-token may need to be updated too.


hellwolf:
addressed in #329 329

PR #343, opened 2021-06-18
Split IDA into a library and main contract

hellwolf:
IDA is at times exceeding the deployment size limit of ethereum networks (EIP-170), it is time to split it into a deployed library (for slots mapping) and main logic contract.


ngmachado (approved):
LGTM

PR #352, opened 2021-06-28
Update create-new-app-registration-key.js

hellwolf:


hellwolf:
@ngmachado please review this small change

hellwolf:
This is to avoid printing registrationkey (which is the secret) to the public eyes

ngmachado (approved):

PR #353, opened 2021-07-01
Make selfMint requireReceptionAck=false

### LINKED ISSUE #345: selfMint requireReceptionAck is hard-coded to true
We should set it to false instead, otherwise a contract (gnosis safe e.g.) cannot self-mint super tokens (ETHx).

## Note

- Create a SelfMintTetsterApp to test it.

ngmachado:


hellwolf (approved):
looks good

PR #366, opened 2021-07-28
Framework loader and load token by address

hellwolf:


ngmachado (approved):
I'm second guessing the need for a different contract (SuperfluidLoader.sol) and not include the functionality directly on the resolver.

hellwolf:
> I'm second guessing the need for a different contract (SuperfluidLoader.sol) and not include the functionality directly on the resolver.

Could be considered when we upgrade resolver.

PR #389, opened 2021-08-12
Fix SDK browser support + add hardhat network

### LINKED ISSUE #388: Bug - Running SDK in browser and using hardhat
When initialising the superfluid framework on hardhat I receive the following error. 

Screenshot 2021-08-12 at 16 28 11

After inspecting the [getConfig.js](https://github.com/superfluid-finance/protocol-monorepo/blob/dev/packages/js-sdk/src/getConfig.js), you can see the configs for the hardhat (chainId: 31337) and localhost (chainId: 1337) are missing. 

Additionally `process.env.TEST_RESOLVER_ADDRESS` is not defined when running the SDK in the browser. 

```
    const configs = { ...DEFAULT_CONFIGS[chainId] };
    // overriding environment variables

    configs.resolverAddress =
        process.env.TEST_RESOLVER_ADDRESS || configs.resolverAddress;
```

The issue occurs because `process` is only defined when running the SDK in node, it is not defined when running in the browser. 

So there are two problems:
1. You should be able to use the SDK in the browser.
2. You should be able to use any local test network.

trozler:
Closes #388 

Simply adds support for hardhat and localhost chainIds in `getConfig.js`.

NOTE making the SDK browser compatible should not be too hard:
* Their is only a single reference to `process.env` in the js-sdk, which is found in `getConfig.js`.
* There is an additional reference to `process` in `js-sdk/src/utils/gasMetering.js`.

Approach:
1. In order to separate concerns the resolver address for the test networks will not be set in `getConfig.js`. 
This file exports a getter function and should not be used to set state - even for tests.
Instead test shall specify `resolverAdress` as an option `SuperfluidSDK.Frameworkd({....})` . 

In order to facilitate this another argument has been added to `reset()` in `TestEnvironment.js`. 
```
    async reset(deployOpts = {}, frameworkOpts = {}) {
        console.log("Aliases", this.aliases);
       .....

        // load the SDK
        this.sf = new SuperfluidSDK.Framework({
            gasReportType: this.gasReportType,
            isTruffle: this.isTruffle,
            version: process.env.RELEASE_VERSION || "test",
            ...frameworkOpts,
        });
```

2. We can safely delete the test part in `getConfig.test.js`:
```
        process.env.TEST_RESOLVER_ADDRESS = "0x42";
        const testConfig = SuperfluidSDK.getConfig(5555);
        assert.equal(testConfig.resolverAddress, "0x42");
        delete process.env.TEST_RESOLVER_ADDRESS;
```

This test is already failing as there is no entry for chainId `5555`.

TODO:
1. Resolve `getConfig.js` issue and update tests.

2. Resolve `gasMetering.js`

### DIFF
--- a/packages/ethereum-contracts/test/TestEnvironment.js
+++ b/packages/ethereum-contracts/test/TestEnvironment.js
@@ -62,7 +62,7 @@ module.exports = class TestEnvironment {
      *************************************************************************/
 
     /// reset the system
-    async reset(deployOpts = {}) {
+    async reset(deployOpts = {}, frameworkOpts = {}) {
         console.log("Aliases", this.aliases);
 
         // deploy framework
@@ -80,6 +80,7 @@ module.exports = class TestEnvironment {
             gasReportType: this.gasReportType,
             isTruffle: this.isTruffle,
             version: process.env.RELEASE_VERSION || "test",
+            ...frameworkOpts,
         });
         await this.sf.initialize();
 
--- a/packages/js-sdk/src/Framework.js
+++ b/packages/js-sdk/src/Framework.js
@@ -104,6 +104,7 @@ module.exports = class Framework {
             networkId: networkId,
         });
 
+        /** @dev resolverAddress cannot be undefined after this point. */
         const resolverAddress =
             this._options.resolverAddress || this.config.resolverAddress;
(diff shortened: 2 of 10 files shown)


trozler:
This PR does not address core of issue. Will update.

PR #390, opened 2021-08-13
Fix SDK browser support + add hardhat network

### LINKED ISSUE #388: Bug - Running SDK in browser and using hardhat

trozler:
Closes #388 

Simply adds support for running SDK in browser + hardhat use.

NOTE making the SDK browser compatible should not be too hard:
* Their is only a single reference to `process.env` in the js-sdk, which is found in `getConfig.js`.
* There is an additional reference to `process` in `js-sdk/src/utils/gasMetering.js`.
* We can safely delete the test part in `getConfig.test.js`:
```
        process.env.TEST_RESOLVER_ADDRESS = "0x42";
        const testConfig = SuperfluidSDK.getConfig(5555);
        assert.equal(testConfig.resolverAddress, "0x42");
        delete process.env.TEST_RESOLVER_ADDRESS;
```

This test is already failing as there is no entry for chainId `5555`.

- added support for hardhat and localhost chains.
- solved getconfig issue.
- remove dead test.


packages/js-sdk/test/getConfig.test.js
```diff
@@ -10,10 +10,5 @@ contract("SuperfluidSDK JS SDK", () => {
         // defaultConfig
         const defaultConfig = SuperfluidSDK.getConfig(8888);
         assert.isUndefined(defaultConfig.resolverAddress);
-        // test environment
```
hellwolf, line 13: please revert this one and see if it works

hellwolf (changes requested):

packages/js-sdk/test/getConfig.test.js
trozler: changes reverted!

hellwolf (approved):

PR #394, opened 2021-08-16
Batch liquidator contract

hellwolf:


ngmachado:
This implementation will lock the liquidation reward if perform when the account is insolvent.

We need to add at least on method to withdraw ERC20 tokens of this contract.

hellwolf:
@ngmachado yes, please add ownable

ngmachado:
We remove ownable and admin function as of this version we transfer the balance always in the end of a batch call

PR #396, opened 2021-08-18
Update yarn.lock and improve README for dev-playground

kasparkallas:
It wasn't as easy as 1-2-3 to get the dev-playground up & running.

I had to fix yarn.lock; set DATABASE_URL and use different Redwood commands than specified in the README.


examples/dev-playground/.env.template
```diff
@@ -1,4 +1,4 @@
 ETHEREUM_JWT_SECRET=
 INFURA_ENDPOINT_KEY=
 XDAI_RPC_URL=
-#DATABASE_URL=
+DATABASE_URL=file:./db.sql
```
kasparkallas, line 4: When doing the test task then I had to figure this out on my own.

examples/dev-playground/README.md
```diff
@@ -16,19 +16,19 @@
 
 #### Step 1.
 
-```sh
-yarn install
+copy `.env.template` to `.env` and add the required variables. You can use this to generate a string for `ETHEREUM_JWT_SECRET`:
```
kasparkallas, line 19: It makes more sense to set-up the environment variables before doing `yarn prisma migrate` etc.

hellwolf (approved):

PR #397, opened 2021-08-19
js-sdk fix node syntax issue

trozler:
This PR aims to fix the issue that occurs when running the SDK in the browser, in combination with webpack5. 
There is a work around where users can install node-js pollyfill, but this is not ideal. 

Instead we should move the node syntax into the tests - the SDK should be indifferent to the environment it is run in. 

The changes are simple and change how the test resolver address in `TestEnvrionment.js`.  Since `TestEnvironment` is used by both the SDK and contract, there might be concerns about this change. 
Any feedback or guidance would be much appreciated. 

The PR passes the superfluid tests + also works in a webpack5 project.


hellwolf:
thank you!

PR #404, opened 2021-08-25
Commit .idea (Webstorm) settings

kasparkallas:
Not 100% sure these are the best settings but at least it handles projects' intellisense correctly.

The **project** view now looks like this:
image

It's easy to open a project and have correct intellisense and syntax highlighting in that project. No accidental imports from outside the scope of a project.

The **project files** view now looks like this:
image

### DIFF
--- a/.idea/.gitignore
+++ b/.idea/.gitignore
@@ -1,0 +1,5 @@
+# Default ignored files
+/shelf/
+/workspace.xml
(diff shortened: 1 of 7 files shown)


d10r (approved):
looks good to me, also according to https://intellij-support.jetbrains.com/hc/en-us/articles/206544839 those files are indeed meant to be shared.

PR #419, opened 2021-09-13
Upgrade Resolver

d10r:
- [x] add Set event
- [x] rename from TestResolver to ?
- [x] move to directory `ux` (from `test`)
- [ ] deploy to same address on all networks (e.g. using [this](https://github.com/zoltu/deterministic-deployment-proxy))


d10r:
@hellwolf rename to just _Resolver_? Or _SuperfluidResolver_? Or ... ?

d10r:
The last commit also affects the external interface as it renames env vars. Not sure if we can/should do that without keeping backwards compatibility to the previously used names at least for a while.

d10r:
deployed to goerli using that deterministic deployment proxy:
```
truffle(goerli)> web3.eth.sendTransaction({from: sender, to: "0x7A0D94F55792C434d74a40883C6ed8545E406D12", data: Resolver.binary})
```
which led to a successful deployment of the Resolver at https://goerli.etherscan.io/address/0x7bdb7cc80ca59bba3b74e2d7cd544d5595ac3d97

Trying to deploy the same bytecode a 2nd time through that proxy causes a revert.

The deployer seems to exist on all testnets and mainnets currently relevant for us.

hellwolf:
>  deploy to same address on all networks (e.g. using this)

@d10r We could create a scripts/deploy-erc1820.js (which uses pre-built artifacts/ERC1820Registry.json) counter part for resolver, and have its own prebuit artifacts/Resolver.json. As long as we consider our resolver implementation is stable enough, we could pull the trigger.

hellwolf:
I stole everything to https://github.com/superfluid-finance/protocol-monorepo/pull/507

PR #423, opened 2021-09-14
Ts-sdk

JoshuaTrujillo15:
Clean branch off dev, fixed some import issues, as well as typing issue


kasparkallas:
Can we move the .tsconfig into js-sdk's package folder? Because not all packages will have exactly the same TypeScript config.

packages/js-sdk/src/ConstantFlowAgreementV1Helper.d.ts
```diff
@@ -1,6 +1,7 @@
 import { Transaction } from "web3-core";
```
0xdavinchee, line 1: can we do `import type { Transaction } from "web3-core";`?

packages/js-sdk/src/ConstantFlowAgreementV1Helper.d.ts
```diff
@@ -1,6 +1,7 @@
...
-import Framework from "./Framework";
+import type { Framework } from "./Framework";
 import type { LoadedContract } from "./loadContracts";
+import type BN from 'bn.js';
```
0xdavinchee, line 4: can we use `import type { BigNumber } from "ethers";`? 
there is a slight difference from `bn.js`: https://docs.ethers.io/v5/api/utils/bignumber/#BigNumber--BigNumber--notes--why-not-bignumber-js-bn-js-bigdecimal-etc

however, this may break something with web3, but I think if we can use `BigNumber` it is more suitable.

packages/js-sdk/src/loadContracts.d.ts
```diff
@@ -2,6 +2,32 @@ import { ethers, Contract as EthersContract, utils } from "ethers";
...
+  | 'ISuperTokenFactory'
+  | 'ISuperfluid'
+  | 'ISuperfluidGovernance'
+  | 'SuperfluidLoader'
+  | 'TestToken'
+  | 'TokenInfo'
+  | 'UUPSProxiable';
+
+export type SuperfluidContractObject = {
+  abi: any;
```
0xdavinchee, line 25: not a super useful change, but I believe we can get the type of this via:
`import type { ContractInterface } from "ethers";`
and
`abi: ContractInterface`.

again, this may break something with the tests that I am not aware of, but worth a try.

packages/js-sdk/src/loadContracts.d.ts
```diff
@@ -1,8 +1,8 @@
-import { ethers, Contract as EthersContract, utils } from "ethers"; 
+import type { ethers, Contract as EthersContract, utils } from "ethers"; 
 import type { Contract as Web3Contract } from "web3-eth-contract";
 import TruffleContract from "@truffle/contract";
 import type Web3 from "web3";
-import type { Contract } from 'ethers'
+import type { Contract, ContractInterface } from 'ethers'
```
0xdavinchee, line 5: One small comment: you are already importing ethers above and giving `Contract` an alias of `EthersContract`, I think it makes sense to remove `Contract` and just use `EthersContract` on line 27 and to move `ContractInterface` into the import in line 1.
Everything else looks great though!

0xdavinchee (approved):
looks good 👍🏼

hellwolf:
@JoshuaTrujillo15 awesome, thanks!

PR #427, opened 2021-09-20
fix up user object

0xdavinchee:
- add some docs
- add some tests
- we check if the flowRate property is empty instead of !flowRate
- we also check if the type of flowRate is not a string, we throw an error if this is the case otherwise the code may behave strangely


hellwolf (approved):
lktm

PR #444, opened 2021-10-03
[JS-SDK] Type information fixes

JoshuaTrujillo15:


packages/js-sdk/src/ConstantFlowAgreementV1Helper.d.ts
```diff
@@ -3,16 +3,64 @@ import type { Framework } from "./Framework";
...
-  inFlows: Array<Flow>;
-  outFlows: Array<Flow>;
-}>;
+// MUST include 'by' type, it's only diff between create and update
+export interface DeleteFlowOptions {
+  superToken: string;
+  sender: string;
+  receiver: string;
+  flowRate: string;
+  by: string;
```
0xdavinchee, line 47: I don't think `by` should be required, if you look at L154-155 of `CFAv1Helper.js`. It handles the case where it can be undefined (just uses the sender address for `by`).

packages/js-sdk/src/InstantDistributionAgreementV1Helper.d.ts
```diff
@@ -3,11 +3,144 @@ import type { Framework } from "./Framework";
 import type { LoadedContract } from "./loadContracts";
 import type BN from 'bn.js';
 
-export interface Subscription {
+// comes from getIndex in IDAv1 contract,
+// then passed into _sanitizeIndexData
+// then returned from getIndex
+export interface IndexData {
     exist: boolean;
     approved: boolean;
```
0xdavinchee, line 11: I think `approved` is supposed to be replaced with `indexValue`.

packages/js-sdk/src/InstantDistributionAgreementV1Helper.d.ts
```diff
@@ -3,11 +3,144 @@ import type { Framework } from "./Framework";
...
+    units: string;
+    pendingDistribution: string
+}
+
+// comes from listSubscriptions in IDAv1 contract,
+// then passed into __sanitizeSubscriptionInfo
+// then returned from listSubscriptions
+export type SubscriptionInfoList = {
+    publisher: string;
+    indexId: string;
```
0xdavinchee, line 31: `indexId` should be of type number (L570 of `InstantDistributionAgreementV1Helper.js`).

packages/js-sdk/src/User.d.ts
```diff
@@ -4,46 +4,59 @@ import type { Transaction } from 'web3';
...
-    }): Promise<void>;
+    details(): Promise<UserDetails>;
+    flow({
+        recipient,
+        flowRate,
+        ...options
+    }: UserFlowOptions): Promise<Transaction>;
+    createPool({ poolId: indexId }: CreatePoolOptions): Promise<Transaction>;
+    giveShares({ recipient, shares, poolId: indexId }: GiveSharesOptions): Promise<Transaction>;
+    distributeToPool({ poolId: indexId, amount }: DistributeToPoolOptions): Promise<void>;
```
0xdavinchee, line 61: I believe `distributeToPool` returns a `Transaction` object. Also, I messed up by not updating the typings file when I updated the `User.js` file, but all of the function which return a `Transaction` are wrapped in a try/catch and a more accurate return in the typings for all functions wrapped in a try/catch without a return for the catch would be: `Promise<Transaction | undefined>`

packages/js-sdk/src/User.d.ts
```diff
@@ -4,46 +4,59 @@ import type { Transaction } from 'web3';
 import type { Flow } from './ConstantFlowAgreementV1Helper';
 import type { Subscription } from './InstantDistributionAgreementV1Helper';
 
-export type DetailsType = {
+// params options types
+export interface UserOptions {
+    sf: Framework;
+    address: string;
+    token: string;
+    options?: any;
```
0xdavinchee, line 12: this can be removed, if you look in the `User.js` file, we removed this because it was not being used for anything.

packages/js-sdk/src/User.d.ts
```diff
@@ -4,46 +4,59 @@ import type { Transaction } from 'web3';
...
+}
+
+export interface GiveSharesOptions {
+    recipient: string;
+    shares: number | BN;
+    poolId: number;
+}
+
+export interface DistributeToPoolOptions {
+    poolid: number;
```
0xdavinchee, line 33: I believe this should be `poolId`.

packages/js-sdk/src/User.d.ts
```diff
@@ -4,46 +4,59 @@ import type { Transaction } from 'web3';
...
+    address: string;
+    token: string;
+    options?: any;
+}
+
+// specified User to avoid confusion with CFA CreateFlowOptions
+export interface UserFlowOptions {
+    recipient: string;
+    flowRate: string;
+    options?: any;
```
0xdavinchee, line 19: options can a type which looks like: 
```
{
    userData?: string;
    onTransaction?: () => void;
    by?: string;
}
```

packages/js-sdk/src/User.d.ts
JoshuaTrujillo15: Removed the 'options' from UserOptions, but I left the interface, as it types the params of the User constructor: Framework, address, and token (useful for external ts packages).

0xdavinchee (approved):
Looks good:

- lots of new interfaces provided for much better typing
- fixed up returns for existing functions (namely idahelper getter functions)

PR #455, opened 2021-10-15
Subgraph v1 tests

0xdavinchee:


packages/subgraph/src/mappings/superToken/ropstenSuperToken.ts
```diff
@@ -0,0 +1,55 @@
+import { Address } from "@graphprotocol/graph-ts";
```
hellwolf, line 1: this file should be deleted.

Also for generated file, it would be good to add a comment in the beginning showing that it is a generated file, do not edit manually.

packages/subgraph/src/mappings/superTokenFactory/ropstenSuperTokenFactory.ts
```diff
@@ -0,0 +1,33 @@
+import { Address } from "@graphprotocol/graph-ts";
```
hellwolf, line 1: ditto

.github/workflows/ci-test-subgraph.yml
```diff
@@ -0,0 +1,89 @@
+name: CI Test Subgraph Locally
+
+on:
+  workflow_dispatch:
```
hellwolf, line 4: can we make this part of the ci.feature.yml instead: https://github.com/superfluid-finance/protocol-monorepo/blob/9af63906cbdfcc43994df25e9d875319a215b423/.github/workflows/ci.feature.yml#L128

this ci-test-subgraph.yml is not something we need

packages/subgraph/src/mappings/superToken/ropstenSuperToken.ts
0xdavinchee: oops, I forgot to remove this, we just have one generated file-`addresses.ts` (which is not committed per .gitignore) which hardcodes the addresses for the different networks. it adds a few extra lines to the the package.json (9), but removes a bunch of files. thanks to @kasparkallas for the suggestion.

packages/subgraph/package.json
```diff
@@ -10,30 +10,39 @@
         "codegen": "graph codegen",
         "build": "graph build",
         "test": "truffle test --network ganache",
-        "prepare-local": "mustache config/ganache.json subgraph.template.yaml > subgraph.yaml",
+        "prepare-local": "mustache config/ganache.json test-subgraph.template.yaml > subgraph.yaml",
```
hellwolf, line 13: Please also bump version to "1.0.0"

.github/workflows/ci-test-subgraph.yml
0xdavinchee: added and working now

packages/subgraph/package.json
0xdavinchee: also updated the version in `package.json`

PR #465, opened 2021-10-27
SDK-Redux

kasparkallas:


.github/workflows/handler.build-examples.yml
```diff
@@ -1,4 +1,6 @@
 name: Build Examples
+env:
```
hellwolf, line 2: will we add sdk redux demo to the examples build matrix?

.github/workflows/handler.deploy-subgraph.yml
```diff
@@ -1,4 +1,6 @@
 name: Deploy Subgraph
+env:
+  GITHUB_TOKEN: ${{ secrets.CI_GITHUB_TOKEN }}
```
hellwolf, line 3: GITHUB_TOKEN & CI_GITHUB_TOKEN usages are now quite mixed.

There is still places using GITHUB_TOKEN, instead of the globally set CI_GITHUB_TOKEN.

To be clear:

- CI_GITHUB_TOKEN is personal token
- GITHUB_TOKEN is a automatically generated token by github

.gitignore
```diff
@@ -25,4 +25,4 @@ package-lock.json
 
 .vscode/
 *-workspace
-.npmrc
+**/*/.npmrc
```
hellwolf, line 28: what's this for?

.github/workflows/handler.deploy-subgraph.yml
kasparkallas: Yes... :disappointed:

Currently, with the automatic `yarn link` through workspaces, we don't actually need this `CI_GITHUB_TOKEN` with the `.npmrc` approach because the build doesn't have to download any @superfluid-finance packages (because they're linked from the monorepo). We could remove all of this `env` code I added...

.gitignore
kasparkallas: I had to include the root `.npmrc` to get the package published initially. I still ignored all of the subdirectory .npmrcs just in case...

Like I said in the previous comment we could still ignore all of the .npmrc's when we keep the workspaces approach.

packages/sdk-redux/README.md
```diff
@@ -0,0 +1,175 @@
+<h1 align="center">sdk-redux</h1>
+<div align="center">
+<img  width="300" padding="0 0 10px" alt="Superfluid logo" src="/sf-logo.png" />
+<p>
+  <a href="https://www.npmjs.com/package/@superfluid-finance/sdk-redux" target="_blank">
+    <img alt="Version" src="https://img.shields.io/npm/v/@superfluid-finance/sdk-core.svg">
```
hellwolf, line 6: typo here "sdk-redux"

package.json
```diff
@@ -37,7 +38,9 @@
             "packages/ethereum-contracts",
             "packages/js-sdk",
             "packages/sdk-core",
-            "packages/subgraph"
+            "packages/sdk-redux",
+            "packages/subgraph",
+            "examples/sdk-redux-react-typecript"
```
hellwolf, line 43: sdk-redux-react-typescript should not be included here, since it's an example project, in order to be consistency with other example projects.

(also maybe let's consider to rename it to a better project name, maybe `example/sdk-redux-demo/packages.json` ?)

To the practical concern about how to develop this example project going forward:

If `sdk-redux` or `sdk-core` changes are required:

A. if the changes are small, consider to use the normal process to publish a new canary package of the SDK
B. if the changes are big enough that need a bit more testing together with the example project itself
  1. Create a feature branch
  2. Change locally `pacakge.json` and `example/sdk-redux-demo/packages.json` to make it temporarily workspace package during the development process
  3. After development and testing, extract the sdk changes to a new feature branch, and use the normal process to make sure that new sdk changes are clear to everyone
  4. After the canary builds (or even a release build) of the sdk is available, revert what's done in (2), and make the example project work again as independent project.

This way we can force us to have the chance of understanding what's the impact of any sdk changes to the end users.

kasparkallas: I'm OK with it if I can do a manual `yarn link @superfluid-finance/sdk-redux` once to the example project and then change branches and the link sticks.

hellwolf (approved):
image

PR #475, opened 2021-11-02
Streaming option example

saflamini:
Hey @hellwolf - I'm going to use this in an example workshop tomorrow. If you could help me merge, that would be great so I can link it after my presentation.

I've gotten really into this one, so I will add new features, testing, etc. later on for sure


hellwolf:
GMI

PR #483, opened 2021-11-05
fixed PIC definition in interface doc

d10r:
had the definition of the abbreviation _PIC_wrong in the interface doce


hellwolf (approved):
gmi

PR #490, opened 2021-11-16
Dao global examples

bertux:
- to make it work with current versions and lock them
- to use yarn scripts
- to use .env template
- to use Etherscan validation


saflamini:
@hellwolf This looks good. I will add some additional tests and submit a PR for dev 

@bertux Thanks for cleaning some of this up - we were in a rush to get this up for the hackathon workshop.

PR #499, opened 2021-11-19
Update dependencies

hellwolf:
WIP

### DIFF
--- a/packages/ethereum-contracts/package.json
+++ b/packages/ethereum-contracts/package.json
@@ -45,7 +45,7 @@
         "@decentral.ee/web3-helpers": "0.5.3",
         "@openzeppelin/contracts": "3.4.1",
         "@superfluid-finance/js-sdk": "0.5.5",
-        "@truffle/contract": "4.3.35",
+        "@truffle/contract": "4.3.42",
         "ethereumjs-tx": "2.1.2",
         "ethereumjs-util": "7.1.0",
         "stack-trace": "0.0.10"
--- a/packages/ethereum-contracts/package.json
+++ b/packages/ethereum-contracts/package.json
@@ -39,31 +39,31 @@
         "pre-commit": "run-s pre-commit:*",
         "pre-commit:lint": "yarn lint",
         "pre-commit:check-no-focused-tests": "yarn check-no-focused-tests",
-        "check-updates": "ncu --dep prod,dev"
+        "check-updates": "ncu --target minor --dep prod,dev"
     },
     "dependencies": {
         "@decentral.ee/web3-helpers": "0.5.3",
-        "@openzeppelin/contracts": "3.4.1",
+        "@openzeppelin/contracts": "3.4.2",
         "@superfluid-finance/js-sdk": "0.5.5",
         "@truffle/contract": "4.3.42",
         "ethereumjs-tx": "2.1.2",
-        "ethereumjs-util": "7.1.0",
+        "ethereumjs-util": "7.1.3",
         "stack-trace": "0.0.10"
     },
     "devDependencies": {
-        "@openzeppelin/test-helpers": "0.5.13",
-        "@truffle/hdwallet-provider": "^1.5.0",
-        "@typechain/ethers-v5": "^7.2.0",
-        "async": "^3.2.1",
+        "@openzeppelin/test-helpers": "0.5.15",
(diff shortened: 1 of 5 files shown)


d10r:
status quo:
With no gas price explicitly set, read operations succeed.  
With a gas price set (e.g. in .env), I (still) get such results (using an unfunded throwaway account):
```
$ MUMBAI_MNEMONIC=60543b9c7c48a67d99a7e333236e6be48bf1350401fdf921e36844de0e83bb37 RELEASE_VERSION=v1 npx truffle exec scripts/info-show-protocol.js --network mumbai                   
Using network 'mumbai'.

truffle detected true
No colon arguments
use truffle native environment true
protocol release version: v1
Initializing Superfluid Framework...
version v1
networkType private
networkId 80001
chainId 80001
Using @superfluid-finance/js-sdk within a Truffle native environment.
                Truffle artifacts must be present.
Resolver at 0x8C54C83FbDe3C59e59dd6E324531FB93d4F504d3
Error: err: insufficient funds for gas * price + value: address 0xF6f0FCffD2bc8B76e41d9a531c399eB2F34c9a62 have 0 want 50000000000000000 (supplied gas 50000000)
```
It seems reasonable for calls to have a gas limit (in order to avoid endless loops), but truffle should set the gas price to 0 imho.  
The problem is that when using a script doing txs too and setting a gas price is needed, this limit of 50M gas can lead to a pretty high balance requirement - much higher than the actual txs would require.

Most of the time, not setting an explicit gas price should work just fine. It's however dependent on the connected RPC.  
In the past I've seen cases where this leads to bad outcomes, e.g. the public xdai RPC would reporting a gas price of 0, and also accept txs with that gas price, but then have them pending forever, blocking the nonce and thus any further interactions from that account. Currently not the case with https://rpc.xdaichain.com/. Also tried with an Alchemy Mumbai node and with an Infura Goerli node, transacting went smoothly.

Trials with new networks with default gas settings (nothing explicitly set):

Arbitrum Rinkeby: reasonable default settings are chosen for gas price. [Example tx](https://testnet.arbiscan.io/tx/0x60d18b7ae7e10d71b39e8097b21b60595dcffe62e524c52881a7661ed3e12b2a).  
Unlike on (all?) other networks, the gas limit is 
 (the 250M gas limit are afaik comparable to 6.7M on "normal" networks.  
What's different with Arbitrum: without gas limit set, 250M are set for all txs. On other networks, an estimation is done and the limit set to +25%.  
The high default gas limit is not per se a problem, but can require the sender account to have much more funds than actually needed for the txs done - to be taken into consideration for mainnet.

Avalanche Fuji: sound defaults set, incl. gas estimiation. [Example tx](https://testnet.snowtrace.io/tx/0x7324a326820170ddc545cfbadc0b449949b20135887acf17d3d20c07a638296c)

Optimistic Kovan: sound settings selected (I guess). [Example tx](https://kovan-optimistic.etherscan.io/tx/0x3dfe0feb7e11ae7ccb45110609da99ddfdec16e4709e691edfb65710e4323f8b)

d10r:
`deploy_framework.js` is failing:

```
$ npx truffle exec scripts/deploy-framework.js
Using network 'development'.

truffle detected true
TypeError: Cannot use 'in' operator to search for 'isTruffle' in undefined
    at /home/didi/src/superfluid/protocol-monorepo-cur/packages/ethereum-contracts/scripts/libs/truffleScriptRunnerFactory.js:39:31
    at Object.exec (/home/didi/src/superfluid/protocol-monorepo-cur/node_modules/truffle/build/webpack:/packages/require/require.js:127:1)
    at internal/util.js:308:30
    at new Promise (<anonymous>)
    at bound exec (internal/util.js:307:12)
    at Object.run (/home/didi/src/superfluid/protocol-monorepo-cur/node_modules/truffle/build/webpack:/packages/core/lib/commands/exec.js:75:1)
    at processTicksAndRejections (internal/process/task_queues.js:95:5)
    at Command.run (/home/didi/src/superfluid/protocol-monorepo-cur/node_modules/truffle/build/webpack:/packages/core/lib/command.js:183:1)
Truffle v5.4.21 (core: 5.4.21)
Node v14.17.6
```

hellwolf:
> npx truffle exec scripts/deploy-framework.js

fixed @d10r

d10r (approved):
let's go!

PR #502, opened 2021-11-22
Cleanup example projects

hellwolf:
## Checklist

- [x] removed dev-playground
- [x] removed react-ts-sdk
- [x] updated example projects' dependencies
- [x] handler.build-examples.yml to support generic build-and-test.sh as build step
- [x] include streaming-call-option
- [x] include nftbillboard-userdata


saflamini:
1) Added build-and-test.sh file to 2 additional examples
2) Added test cases to 2 additional examples
3) Fixed errors in flowlottery build script

@hellwolf all examples built and ran successfully. Please let me know if you see something I may have missed

hellwolf:
@saflamini looks good!

examples/streaming-call-option/artifacts/@openzeppelin/contracts/introspection/ERC165.sol/ERC165.dbg.json
```diff
@@ -1,4 +1,4 @@
 {
```
0xdavinchee, line 1: I think it's best to add all `artifacts` or any sort of generated files in `.gitignore`.

examples/nftbillboard-userdata/build-and-test.sh
```diff
@@ -0,0 +1,21 @@
+#!/bin/env bash
```
0xdavinchee, line 1: I think that a few of these `build-and-test.sh` files are repetitive and can be either replaced by the `testenv-ctl.sh` file in `ethereum-contracts` or if these files are different enough, to have their own `testenv-ctl.sh` file for all example projects-it seems that the projects will all have somewhat similar environment set ups unless there's an example project which exists to illustrate a new way of setting up a test environment (e.g. w/ hardhat network or off a fork, etc.)

examples/README.md
```diff
@@ -1,11 +1,6 @@
-# Example SuperApps
+TODO...
```
kasparkallas, line 1: TODO

examples/streaming-call-option/hardhat.config.js
```diff
@@ -32,13 +33,18 @@ module.exports = {
...
+  //     url: "http://127.0.0.1",
+  //     network_id: "*",
+  //     port: 8545,
+  // },
+    // rinkeby: {
+    //   url: `${process.env.RINKEBY_ALCHEMY_URL}`,
+    //   gasPrice:  1500000000,
+    //   accounts: [`0x${process.env.RINKEBY_DEPLOYER_PRIVATE_KEY}`]
+    //   },
+    // },
```
kasparkallas, line 47: Commented out code

kasparkallas (approved):
I wonder if the example frond-end apps with MetaMask when hosted on a site like https://codesandbox.io

examples/README.md
saflamini: will look into https://codesandbox.io - could be interesting to do for your redux sdk example

examples/nftbillboard-userdata/build-and-test.sh
saflamini: My one concern here is with using different test environments. I was able to use ganache + truffle within two of the hardhat projects, but may want to use hardhat network specifically for an example project in the future. 

If this is the case, would it still be best to add a` testenv-ctl.sh` file for all example projects?

hellwolf: We can leave as it is for this branch. When we have a new testenv-ctl published, we update the examples to use it.

saflamini: Sounds good

saflamini (approved):

PR #504, opened 2021-11-24
Create codecov.yml

0xdavinchee:
add a basic `codecov.yml` which tests out the carryforward flag feature to see if it solves the issue of 0% coverage ethereum-contracts when sdk-core information is uploaded to codecov first.


hellwolf (approved):
good

PR #506, opened 2021-11-24
single network deploy implemented

0xdavinchee:
- yml file changed up to be more generalized
- package.json deploy scripts changed to be more generalized


hellwolf (approved):
gmi

PR #515, opened 2021-12-01
extracted changes from sdk-redux re sdk-core

kasparkallas:


packages/sdk-core/test/setup.ts
```diff
@@ -0,0 +1,113 @@
+import { ethers } from "hardhat";
```
0xdavinchee, line 1: please delete this file

0xdavinchee (commented):
looks good, pls just delete the `test/setup.ts` file.
nice solution w/ the typedoc too btw!

packages/sdk-core/test/setup.ts
hellwolf: deleted

0xdavinchee (approved):
gmi

PR #517, opened 2021-12-03
[sdk-redux] Store SuperfluidContext in a global

kasparkallas:
Why?
`SuperfluidContext` is a _service locator_ used in the Redux reducers. Current implementation of storing it seems to have an issue with HMR runtimes like the one CodeSandbox & Next.js have.

Solution?
Use `globalThis` syntax to store `SuperfluidContext` in a global object (e.g. `window`).

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -1,4 +1,3 @@
 coverage
 coverage.json
 build
-/packages/sdk-redux/src/typings/global.d.ts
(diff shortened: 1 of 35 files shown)


.eslintignore
```diff
@@ -1,3 +1,4 @@
 coverage
 coverage.json
 build
+/packages/sdk-redux/src/typings/global.d.ts
```
hellwolf, line 4: move this to subfolder

hellwolf (approved):

PR #520, opened 2021-12-03
readme fixed

0xdavinchee:
- added updated query functions to readme
- fixed up cfa/ida/token functions to use operation syntax
- some general formatting


hellwolf (approved):
I can't find a suitable GMI meme, only this one showed up

image

PR #521, opened 2021-12-03
native coin contract verification

hellwolf:


0xdavinchee (approved):
gmi - looks like it is more generalizable to non-eth native token networks

PR #525, opened 2021-12-08
Subgraph liquidation script

0xdavinchee:
- a script to get liquidations of streams which were less than a week old and the losses were greater than $5: used the subgraph and the coingecko api


hellwolf (approved):
gmi
image

PR #530, opened 2021-12-09
Fix: type problem of createSigner

bertux:
by adding web3Provider as for the other usecases.


0xdavinchee (approved):
gmi, tyvm

0xdavinchee (approved):
gmi, looks good

PR #537, opened 2021-12-15
Sdk core web3 hardhat ethers init framework

0xdavinchee:
- `package.json` fixes for workspace issue
- injected web3, hardhat supported
- tests added for the initialization
- readme updated to show examples of injected initializations


packages/sdk-core/package.json
```diff
@@ -75,6 +75,7 @@
         "eslint-config-prettier": "^8.3.0",
         "eslint-plugin-eslint-comments": "^3.2.0",
         "eslint-plugin-import": "^2.25.3",
+        "eslint-plugin-prettier": "^4.0.0",
```
kasparkallas, line 78: Why is this necessary? Currently SDK-Core & SDK-Redux should be linted without the ESLint plugin (if you look at the lint scripts) as per prettier's recommendation: https://prettier.io/docs/en/integrating-with-linters.html#notes

`"eslint-plugin-prettier"` is defined and used at the workspace root level though... I would leave it on that level for now.

0xdavinchee: oops, not sure how that got in there

packages/sdk-core/src/Framework.ts
```diff
@@ -103,7 +109,19 @@ export default class Framework {
...
+                ? options.provider
+                : options.web3 != null
+                ? new ethers.providers.Web3Provider(
+                      options.web3.currentProvider as
+                          | ethers.providers.ExternalProvider
+                          | ethers.providers.JsonRpcFetchFunc
+                  )
+                : // NOTE: this will not be null as we check this in our validate function
+                  options.hardhatEthers!.provider;
+
```
kasparkallas, line 123: This type of code is a bit smelly because cyclomatic complexity is increasing and types are not clearly communicating. `provider`, `web3` and `hardhatEthers` have a priority order, i.e. if all three are passed in then some will get ignored. Because this is an SDK and entrypoint to the framework then it's probably OK because we will have it well-tested and well-documented. But generally in this type of situation you would want to achieve code similar to this: `provider: Web3 | HardhatProvider | EthersProvider` and this case the types communicate more clearly.

kasparkallas (approved):

PR #540, opened 2021-12-16
ethers overrides object support added

0xdavinchee:


packages/sdk-core/src/InstantDistributionAgreementV1.ts
```diff
@@ -208,7 +219,8 @@ export default class InstantDistributionAgreementV1 {
         return this.host.populateCallAgreementTxnAndReturnOperation(
             this.options.config.idaV1Address,
             callData,
-            userData
+            userData,
+            overrides
```
kasparkallas, line 223: Doesn't it break here and bunch of other places when undefined is passed?

kasparkallas (approved):

PR #541, opened 2021-12-16
Refactor SDK-Redux in preparation for better NextJS support

kasparkallas:
Why?
De-couple React from the SDK and don't include extra dependencies to enable SSR/persistance.

Main changes:
* "SuperfluidContext" is now "SdkReduxConfig"
* Instead of exposing `SdkReduxConfig` to users we expose `setFrameworkForSdkRedux` & `setSignerForSdkRedux`. It's more explicit and `SdkReduxConfig` will act as an "internal" global object (for configuration & service location).
* `sdk-core` is now a peer dependency of `sdk-redux`. We follow the rule that when a dependency's (`sdk-core`) interfaces are exposed from a library (`sdk-redux`) then that makes the dependency a peer dependency. If we didn't do it this way then it could be that our library returns a interface (e.g. `Stream` from `sdk-core`) and the consumer has `sdk-core` dependency with a different version then `Stream` from the same dependency with different version might not match anymore.
* `@reduxjs/toolkit` is now a peer dependency of `sdk-redux`. Following the same rule that dependency's interfaces are exposed (e.g. `CreateApi<T>`).
* Initialization of SDK-Redux is now lazy so that consumer can do additional configuration when creating Redux Slices. Examples of configuration: enabling SSR, enabling persistance, enabling React Hooks generation.
* Query/Mutation message object and the implementation is not code splitted into a separate file anymore. When making SDK-Redux initialization lazy then the possibility of using RTK Query's `injectEndpoints` disappeared because you can't make it work with lazy initialization. It's possible to make the code splitting work without `injectEndpoints` but it would add annoying boilerplate. In short, `addQueryEndpoints.ts` contains all the query endpoint implementations, `addMutationEndpoints.ts` contains all the mutation endpoint implementations and the messages are in `queries.ts` and `mutations.ts`.
* Other changes are small renamings


0xdavinchee (approved):
looks good

PR #542, opened 2021-12-16
Cfa wrapper library

saflamini:
CFALibraryV1 -  includes helper functions for creating, updating, and deleting streams.

CFALibraryMock - includes a contract which makes use of the helper functions, and a version of the tradeable cashflow which makes use of them as well

CFALibrary.test.js - provides light testing on the create, update, and delete functions, and also tests that the withCtx versions of each function work inside of the tradeable cashflow callbacks. 

Recent changes:
- struct to initialize the library was renamed to initData
- example for initializing the library by deriving the cfa from the host is inside of the CFALibraryMock contract. This will be explained in documentation as well.


saflamini:
Hi @hellwolf - is there anything else you need to merge the above?

packages/ethereum-contracts/contracts/libs/CFALibraryV1.sol
```diff
@@ -0,0 +1,285 @@
...
+                new bytes(0) // placeholder
+            ),
+            "0x" //empty user data
+        );
+    }
+    
+
+    //update flow with userData
+    function updateFlow(
+        InitData memory cfaLibrary,
```
hellwolf, line 90: "storage" should do, and the rest.

packages/ethereum-contracts/contracts/mocks/CFALibraryMock.sol
```diff
@@ -0,0 +1,400 @@
...
+        address receiver,
+        ISuperfluidToken token,
+        bytes memory userData
+    ) public {
+        cfaV1.deleteFlowWithCtx(ctx, sender, receiver, token, userData);
+    }
+
+}
+
+contract RedirectAll is SuperAppBase {
```
hellwolf, line 160: I understood that this is to reuse the redirectall for testing all the positive cases, right?

Though, to reduce amount of code in the repository, would it be possible if we reuse MFA.sol and use the helper library in that instead?

hellwolf (changes requested):
Great stuff @saflamini 

- Request some additional changes.
- We will need to run a coverage test (could use ".only" to focus on the CFALibraryV1.test.js file) to see if the CFALibraryV1.sol is fiully covered.

packages/ethereum-contracts/contracts/libs/CFALibraryV1.sol
saflamini: Got it - will change 'memory' to 'storage' both here and elsewhere in the lib 👍🏻

packages/ethereum-contracts/contracts/mocks/CFALibraryMock.sol
saflamini: Yes that is correct.

I think so re: MFA.sol. The goal of using redirectall was to ensure that the WithCtx versions of each function work properly inside of a Super App. To achieve the same with the MFA.sol app, I believe I'll need to create a new version of the MultiFlowApp which uses the CFA Lib to make calls to the CFA instead of the callAgreementWithContext(...) pattern. 

I can create a new version in contracts/test called MultiFlowAppLib.so which only changes the way that flows are created, updated, and deleted. Then I can reuse the tests written in ConstantFlowAgreement.test.js to ensure coverage. Does that work?

hellwolf: Okay, I see.

Using the new CFAv1Library in MultiFlowApp.sol can be a separate task for later then.

And in order to keep some consistency, I suggest we consider the following naming and file location changes:

- [ ] name to CFAv1Library, reason: it is a library for CFAv1
  - CFAv1Library.sol
  - CFAv1Library.test.js
- [ ] CFAv1Library.sol to be moved to contracts/apps/, reason:  SuperAppBase.sol is there, and it's a folder for helping to build your own super apps
  - [ ]  Fix .solcover.js, to not to skip everything from apps/ (skip apps/SuperAppBase.sol only)
- [ ] CFALibraryMock.sol, and merge CFALibraryMock, RedirectAll and TradeableCashflowMock into one CFAv1LibraryTester contract. Reason: to keep the contract strictly for "testing" purpose.

hellwolf (changes requested):

packages/ethereum-contracts/contracts/mocks/CFALibraryMock.sol
saflamini: I will change the names 👍🏻

**On merging the mocks into CFAv1LibraryTester - the current test suite works as follows inside of CFALibrary.test.js:**
1) CFALibraryMock.sol includes a list of functions for creating, updating, and deleting flows which are implemented using the CFAv1Library.

2) The first set of tests call the functions on the CFALibraryMock contract to ensure that each function opens up streams properly to another EOA.

3) The second and third set of tests also call functions on the CFALibraryMock contract, but this time they set the recipient of each flow to be the deployed TradeableCashflowMock contract's address.
Because the callbacks inside of RedirectAll.sol use the WithCtx versions of the create, update, and delete functions of the CFAv1Library, we can check whether they work as expected when a stream is created, updated, or deleted that includes the Tradeable cashflow as the receiver. To do so, we can check to see if the owner of the Tradeable Cashflow contract is receiving a flow equal to the flow being sent into the contract. 

4) Each flow that is sent from the CFALibraryMock contract to the Tradeable Cashflow contract makes use of the overloaded version of the create, update, and delete functions that takes `userData` as a final parameter. The `userData` value is then stored to a public variable inside of the associated callback in the redirectAll contract. The test suite will then check that the userData was properly included.

If I combine the two into the same contract, I will need to reformat the tests somewhat in order to maintain test coverage on the functions which include userData.

Is there another way I could check to see that userData is being passed to the overloaded functions properly? Maybe I could check w Vincent to figure this one out.

saflamini:
@hellwolf I worked with @0xdavinchee this week to get some final changes pushed here:
1) We now have 100% test coverage
2) Contract names inside of CFALibrary Mock have 'Mock' appended to the end of them so they are not included in the package.

Let me know if anything else is needed here!

Thank you,

-Sam

hellwolf (approved):
image

PR #544, opened 2021-12-19
TOGAv2: bond withdrawal in separate tx fallback

d10r:
If sending the previous bond back to the previous PIC fails for any reason, that bond is handed over to a custodial contract and can later be withdrawn from there.  
As a side effect of having more logic after the send() which redeems the bond, a OOG failure of that send could very likely not leave enough gas for the tx to succeed. The fallback is useful anyway, because the send() could fail for other reasons than OOG and it's probably a good idea to just cover all that cases with a fallback.

v2 also allows the current PIC to increase the bond without the need to outbid itself. If the current PIC sends more funds, they're simply added to the current bond. In this case, userData is ignored - won't change the exitRate. Not sure if that's the best semantic.


packages/ethereum-contracts/contracts/mocks/ERC777SenderRecipientMock.sol
```diff
@@ -170,6 +170,12 @@ contract ERC777SenderRecipientMock is Context, IERC777Sender, IERC777Recipient,
 contract ERC777RecipientReverting is IERC777Recipient, IERC1820Implementer {
```
hellwolf, line 170: Noticed a minor thing from previous code already, in package.json

```
    "files": [
        "/contracts/**/*",
        "!/contracts/mocks/*",
        "/build/contracts/*.json",
        "!/build/contracts/*Mock*.json",
        "!/build/contracts/*Tester*.json",
        "/build/abi.js",
        "artifacts/**/*",
        "scripts/**/*",
        "test/TestEnvironment.js"
    ],
```

The npm packaging rule excludes any contract builds that has name Mock or Tester in it, that leaves these testing purpose contract being packaged:

```
$ grep -hR 'contract ' contracts/mocks/   | grep -Eiv 'mock|tester'
contract ERC777RecipientReverting is IERC777Recipient, IERC1820Implementer {
contract ERC777RecipientDrainingGas is IERC777Recipient, IERC1820Implementer {
contract ExclusiveInflowTestApp is SuperAppBase {
contract NonClosableOutflowTestApp is SuperAppBase {
contract SelfDeletingFlowTestApp is SuperAppBase {
contract ClosingOnUpdateFlowTestApp is SuperAppBase {
contract FlowExchangeTestApp is SuperAppBase {
    // using immutable, otherweise the proxy contract would see zero instead
// spliting this off because the contract is getting bigger
```

No big deal, can be dealt later.

packages/ethereum-contracts/contracts/utils/TokenCustodian.sol
```diff
@@ -0,0 +1,51 @@
...
+            token.send(recipient, amount, "0x0");
+            emit CustodianWithdrawal(token, recipient, amount);
+        }
+    }
+
+    // ============ IERC777Recipient ============
+
+    // interface through which ERC777 tokens are deposited to this contract.
+    // requires userData to contain an abi-encoded address which designates the recipient for which the deposit is made.
+    function tokensReceived(
```
hellwolf, line 36: So we do allow the side effect of "parking" tokens for others using this contract.

That is okay.

packages/ethereum-contracts/contracts/mocks/ERC777SenderRecipientMock.sol
```diff
@@ -170,6 +170,12 @@ contract ERC777SenderRecipientMock is Context, IERC777Sender, IERC777Recipient,
...
     bytes32 constant private _TOKENS_RECIPIENT_INTERFACE_HASH = keccak256("ERC777TokensRecipient");
 
+    // allow to use the hook for this contract itself
+    constructor() {
```
hellwolf, line 174: FWIW, this contract is taken from a OZ implementation. WE are not keeping track of the upstream changes so far.

packages/ethereum-contracts/test/contracts/utils/TOGA.test.js
```diff
@@ -632,4 +691,129 @@ describe("TOGA", function () {
         );
         console.log(`gas used by tx: ${r1.receipt.gasUsed}`);
     });
+
+    it("#19 funds in custody can be withdrawn by legitimate owner", async () => {
```
hellwolf, line 695: :+1:

packages/ethereum-contracts/test/contracts/utils/TOGA.test.js
```diff
@@ -592,14 +615,52 @@ describe("TOGA", function () {
...
         );
+        await sendPICBid(alice, superToken, BOND_AMOUNT_1E12, EXIT_RATE_1E3);
 
-        await sendPICBid(bob, superToken, BOND_AMOUNT_1E12, EXIT_RATE_1E3);
+        await sendPICBid(bob, superToken, BOND_AMOUNT_2E12, EXIT_RATE_1E3);
         assert.equal(await toga.getCurrentPIC(superToken.address), bob);
     });
 
+    // this ensures an OOG in fallback transfer can't succeed
+    it("#17 reverting transfer to custodian reverts whole tx", async () => {
```
hellwolf, line 625: :+1:

hellwolf (approved):
image

PR #547, opened 2021-12-22
allow clearing rewardAddress via gov script

d10r:


hellwolf:
@d10r should we merge these?

hellwolf (approved):

PR #550, opened 2022-01-03
use v1 endpoints for subgraph instead of dev in sdk-core

0xdavinchee:
update avafuji resolver address
updated sdk-core, subgraph npm packages


hellwolf (approved):
image

PR #555, opened 2022-01-10
Scripts update: info-show-protocol.js + (new) resolver-grant-role.js 

d10r:
* info-show-protocol.js: don't fail if encountering a broken token deployment
* added script for granting roles on Resolver


hellwolf (approved):
image

Gazing into L2

PR #558, opened 2022-01-12
3ps

0xdavinchee:
`ConstantFlowAgreementV1.sol`
- use new `_decode3PsData` to get liquidationPeriod instead of duplicating same liquidationPeriod retrieval
- rename `signedTotalDeposit` to `signedTotalCFADeposit`: a more descriptive and accurate variable naming
- Add logic for determining `isPatricianPeriod`
- utilize the new `token.makeLiquidationPayoutsV2` function and include comments 

`SuperfluidGovernanceBase.sol`
- added event, get, set, and clear functions for ThreePs config similar to the other configs in the file

`Definitions.sol`
- `decodeThreePsConfig` function added for reuse 

`SuperfluidToken.sol`
- `makeLiquidationPayoutsV2` function implemented for new 3Ps logic

Test files
- TestEnvironment.js: added csv output code
- Refactored ConstantFlowAgreementV1.behavior.js to handle multiflow liquidations AND moved a lot of the helper functions + state storage/get/set out to ConstantFlowAgreementV1.data.js (CFADataModel class)
- Renamed liquidation functions in ConstantFlowAgreementV1.test.js 
- Added new test cases (single account multi flow liquidations)
- `MFASupport.js` class refactored out

SDK-core
- sdk-core subgraph dependent files updated + new script for generating graphql schema added

Subgraph
- new events mapping logic and schema added


hellwolf:
@ngmachado it is in governance contract

ngmachado:
> @ngmachado it is in governance contract

is implement at CFA level: https://github.com/superfluid-finance/protocol-monorepo/blob/fcbac82a77e9f4988ab05bb43cd1f6abb6abecd3/packages/ethereum-contracts/contracts/agreements/ConstantFlowAgreementV1.sol#L104

0xdavinchee:
@ngmachado the function is in the interface now

d10r (dismissed):
This PR is introducing a conflicting definition of _liquidation period_.  
Used to be a number which denotes the timeframe between a flow becoming critical and becoming insolvent (in the idealized case of no incoming flow).  
Now it's in parallel used to refer to the sub-timeframe for PIC, pleb, pirate.  
I suggest we rename the latter to be either _3Ps period_, _3Ps phase_ or _liquidation phase_ in order to avoid that naming conflict and resulting issues. 
Since names with leading digit turned out to not work in some places (e.g. for method names), we may also consider switching from 3P to TP (triple P) on the code level instead of having a mix of 3Ps and ThreePs (then it should however probably be just _TP_, not _TPS_ in order not to create confusion into another direction...)

d10r:
I'm still unhappy with some naming here - at least in the interfaces it should do the job of leading to a correct understanding without reading the implementation.  

`rewardAccount` was previously defined like this for event `AgreementLiquidated`:
```
`* @param rewardAccount Account that collects the reward or bails out insolvent accounts
```
(which btw. contains a hint that it already wasn't an optimal name, bcs. it was not just about rewards, but also about bailouts).

In the interface description of `AgreementLiquidatedV2` it's the same, but then there's also a NOTE clarifying that the rewards may actually also go somewhere else instead.

In event `AgreementLiquidatedBy` (deprecated) it was named `bondAccount` instead, which probably better reflects its nature.  
Maybe we should keep that name (bondAccount) in `AgreementLiquidatedV2` and rename `useDefaultRewardAccount` to e.g. `rewardsToBondAccount`.

`AgreementLiquidatedBy` and `makeLiquidationPayouts` aren't yet flagged as deprecated, I think they should.


As a more general question, I wonder if any good convention exists for choosing `account` vs `address` in names.

packages/ethereum-contracts/.solhint.json
```diff
@@ -7,7 +7,7 @@
         "max-line-length": ["error", 120],
         "func-param-name-mixedcase": "error",
         "modifier-name-mixedcase": "error",
-        "private-vars-leading-underscore": ["error", { "strict": true }],
+        "private-vars-leading-underscore": "off",
```
hellwolf, line 10: hmm, solhint broke this feature right?

packages/ethereum-contracts/contracts/superfluid/SuperfluidToken.sol
```diff
@@ -433,7 +433,7 @@ abstract contract SuperfluidToken is ISuperfluidToken
         if (defaultRewardAccount == address(0)) {
             rewardAccount = liquidatorAccount;
         }
-
+
```
hellwolf, line 343: Just a note, a trivial newline inserted, we don't need to update super token logic for other tokens.

packages/ethereum-contracts/package.json
```diff
@@ -24,12 +25,12 @@
...
         "testenv:deploy": "truffle exec scripts/deploy-test-environment.js : TEST",
         "pretest": "{ yarn testenv:start > /dev/null& } && sleep 5",
-        "test": "run-s test:*:*",
+        "test": "mkdir test/output && run-s test:*:*",
         "test:contracts:js": "truffle test testsuites/all-contracts.js",
         "test:contracts:sol": "truffle test test/contracts/*/*.sol",
         "test:scripts-js:deployment": "truffle test test/scripts/deployment.test.js",
         "test:scripts-sh:deployment": "test/scripts/deployment.test.sh",
-        "test-coverage": "truffle run coverage --network coverage --file testsuites/all-contracts.js",
+        "test-coverage": "mkdir test/output && truffle run coverage --network coverage --file testsuites/all-contracts.js",
```
hellwolf, line 37: `mkdir -p test/output`

But it's not needed in the "test" command

packages/ethereum-contracts/scripts/gov-set-3Ps-config.js
```diff
@@ -0,0 +1,67 @@
...
+
+/**
+ * @dev Set the 3Ps config for a super token.
+ * @param {Array} argv Overriding command line arguments
+ * @param {boolean} options.isTruffle Whether the script is used within native truffle framework
+ * @param {Web3} options.web3  Injected web3 instance
+ * @param {Address} options.from Address to deploy contracts from
+ * @param {boolean} options.protocolReleaseVersion Specify the protocol release version to be used
+ *
+ * Usage: npx truffle exec scripts/gov-set-3Ps-config.js : {TOKEN ADDRESS} {LIQUIDATION PERIOD} {PATRICIAN PERIOD}
```
hellwolf, line 17: 3P period?

hellwolf (approved):
image

hellwolf:
> This PR is introducing a conflicting definition of _liquidation period_. Used to be a number which denotes the timeframe between a flow becoming critical and becoming insolvent (in the idealized case of no incoming flow). Now it's in parallel used to refer to the sub-timeframe for PIC, pleb, pirate. I suggest we rename the latter to be either _3Ps period_, _3Ps phase_ or _liquidation phase_ in order to avoid that naming conflict and resulting issues. Since names with leading digit turned out to not work in some places (e.g. for method names), we may also consider switching from 3P to TP (triple P) on the code level instead of having a mix of 3Ps and ThreePs (then it should however probably be just _TP_, not _TPS_ in order not to create confusion into another direction...)

This should have already been reviewed, please revisit your review.

Per our last discussion:

Create Separate PRs for:
# PR 1
AgreementLiquidatedV2
- Change ISuperfluidToken.sol
- Update subgraph mapper
# PR 2
- Maybe simplify makeLiquidationPayoutsV2 by removing "ANARCHY MODE" ???
  - Remove bunch of test code

PR #560, opened 2022-01-13
Budget nft example

saflamini:
@hellwolf this is the example I referred to earlier today. We have used this in previous workshops and will use again on Thursday.

This example was built with the use case of internal DAO budgeting in mind. It is a single contract that allows the user to issueNFTs which also represent a portion of the DAOs budget. 

1) Calling `issueNFT` will mint an NFT to a specified address & open up a flow to the owner of the NFT. These flowRates are stored in a mapping inside of the contract
2) The `editNFT` function allows the contract owner to edit the `flowRate` being sent to the owner of each newly minted NFT.
3) `burnNFT` will delete the flow associated with that NFT, and burn the NFT
4) `splitStream()` and `mergeStreams()` can be used by sub NFT owners to create sub-budgets of their own
5) A test suite has been created to test all functions mentioned above, and a `build-and-test.sh` script was also added.

Please let me know if anything else is needed for this to be added to the primary examples repo.

Thanks,

-Sam

***UPDATE*** 

It looks like we had a couple of issues with the build workflow. I will ask kaspar if he can review the sdk-redux example, and I will also investigate the flow lottery. 

However, **I am receiving a 126 permission error when running the build scripts for the budget-nft in the github workflow**. It works fine locally, and adding `sudo` to the script in the git workflow is not working either. **Do you know why this could be happening?**


hellwolf (approved):
image

PR #566, opened 2022-01-17
Fix example projects

kasparkallas:


hellwolf (approved):
Thank you @kasparkallas !

PR #578, opened 2022-01-26
schema.graphql generation

0xdavinchee:
- realized that this actually isn't on dev yet, it is part of the 3Ps PR

### DIFF
--- a/packages/sdk-core/package.json
+++ b/packages/sdk-core/package.json
@@ -17,7 +17,7 @@
         "README.md"
     ],
     "scripts": {
-        "build": "yarn generate-abi-files && yarn generate-web3-types && yarn generate-graphql-types && rm -rf dist && run-p build:*",
+        "build": "run-s generate: && rm -rf dist && run-p build:*",
         "build:main": "tsc -p tsconfig.json",
         "build:module": "tsc -p tsconfig.module.json",
         "build:umd": "mkdir -p dist && browserify src/index.ts -s sdkCore -p [ tsify --noImplicitAny ] > dist/index.umd.js",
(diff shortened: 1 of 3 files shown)


kasparkallas (approved):

packages/sdk-core/package.json
```diff
@@ -35,9 +35,10 @@
         "fix:prettier": "yarn lint:prettier --write",
         "fix:eslint": "yarn lint:eslint --fix",
         "pre-commit": "run-s lint:*",
-        "generate-abi-files": "rm -rf src/abi && mkdir src/abi && tasks/build-abi-json.sh",
-        "generate-web3-types": "typechain --target=ethers-v5 --out-dir=./src/typechain \"./src/abi/**/*.json\"",
-        "generate-graphql-types": "graphql-codegen --config subgraph-codegen.yml"
+        "generate:abi-files": "rm -rf src/abi && mkdir src/abi && tasks/build-abi-json.sh",
+        "generate:web3-types": "typechain --target=ethers-v5 --out-dir=./src/typechain \"./src/abi/**/*.json\"",
+        "generate-graphql-schema": "get-graphql-schema https://api.thegraph.com/subgraphs/name/superfluid-finance/protocol-feature-matic > src/subgraph/schema.graphql",
```
kasparkallas, line 40: This is missing ":"

packages/sdk-core/package.json
```diff
@@ -17,7 +17,7 @@
         "README.md"
     ],
     "scripts": {
-        "build": "yarn generate-abi-files && yarn generate-web3-types && yarn generate-graphql-types && rm -rf dist && run-p build:*",
+        "build": "run-s generate:* && rm -rf dist && run-p build:*",
```
kasparkallas, line 20: Does sequence come from order in scripts or alphabetically?

packages/sdk-core/package.json
0xdavinchee: no, we don't want to generate this because we still have to go in and manually change stuff in other files.

kasparkallas: Actually, we don't want to necessarily run this every build so it shouldn't have it...

packages/sdk-core/package.json
hellwolf: order in the script

packages/sdk-core/package.json
hellwolf: yea, if it is one time during the development, it could be a separate step.

`yarn prepare` for example

packages/sdk-core/package.json
0xdavinchee: sequentially, from order, not alphabetically

PR #588, opened 2022-01-29
load super token by symbol

0xdavinchee:


packages/sdk-core/src/Framework.ts
```diff
@@ -250,10 +251,31 @@ export default class Framework {
...
+    loadSuperToken = async (
+        tokenAddressOrSymbol: string
+    ): Promise<SuperToken> => {
+        let address;
+        const isValidAddress = ethers.utils.isAddress(tokenAddressOrSymbol);
+
+        if (isValidAddress) {
+            address = tokenAddressOrSymbol;
+        } else {
+            const superTokenKey = `supertokens.${this.settings.protocolReleaseVersion}.${tokenAddressOrSymbol}`;
```
hellwolf, line 266: error handling

0xdavinchee: added a try/catch

packages/sdk-core/src/Framework.ts
```diff
@@ -250,10 +251,47 @@ export default class Framework {
 
     /**
      * @dev Load a `SuperToken` class from the `Framework`.
-     * @param address the `SuperToken` address
+     * @param tokenAddressOrSymbol the `SuperToken` address or symbol
```
kasparkallas, line 254: Should mention that the symbol has to be listed for this to work?

kasparkallas (approved):

packages/sdk-core/src/Framework.ts
0xdavinchee: true

PR #594, opened 2022-02-01
Subgraph tests use SDK-Core

0xdavinchee:


packages/subgraph/package.json
```diff
@@ -40,13 +40,14 @@
...
         "generate-sf-meta-local": "COMMIT_HASH=local CONFIGURATION=local BRANCH=local TAG=local ./tasks/getSFMeta.sh",
+        "generate-ethers-types": "typechain --target=ethers-v5 --out-dir=typechain './abis/*.json'",
         "check-updates": "ncu --target minor --dep prod,dev"
     },
     "dependencies": {
         "@decentral.ee/web3-helpers": "^0.5.3",
         "@graphprotocol/graph-cli": "0.25.1",
         "@graphprotocol/graph-ts": "0.24.1",
-        "@superfluid-finance/js-sdk": "0.5.10",
+        "@superfluid-finance/sdk-core": "^0.2.0",
```
hellwolf, line 50: per monorepo ethic, you should just use 0.2.0 instead of '^0.2.0'

hellwolf: how about this

hellwolf (changes requested):
Great, surprised it got migrated without many changes. Left couple of package related change requests. Otherwise good to go.

image

packages/subgraph/package.json
0xdavinchee: using 0.2.0 and removed truffle and ganache-time-traveler

PR #598, opened 2022-02-02
Self approve transfer

0xdavinchee:
- interface added for selfApproveFor and selfTransferFrom
- functions added to CustomSuperTokenMock for testing
- tests added for both revert and successful cases for selfTransferFrom and selfApproveFor


hellwolf (commented):
looks good.

@d10r please have a look too

packages/ethereum-contracts/contracts/mocks/CustomSuperTokenMock.sol
```diff
@@ -63,4 +76,23 @@ contract CustomSuperTokenProxyMock is CustomSuperTokenBaseMock {
         // this makes msg.sender to self
         ISuperToken(address(this)).selfBurn(to, amount, userData);
     }
+    
+    // this function self calls transferFrom
```
hellwolf, line 80: perhaps from grammatical reading perspective, holder should be the first parameter for "transfer from"

packages/ethereum-contracts/contracts/interfaces/superfluid/ISuperToken.sol
```diff
@@ -315,6 +315,34 @@ interface ISuperToken is ISuperfluidToken, TokenInfo, IERC20, IERC777 {
...
 
+   /**
+    * @dev Transfer tokens from the `holder` to `recipient`. If `spender`
+    * isn't the same as `holder`, checks if `holder` has allowance to 
+    * spend `spender` tokens.
+    *
+    * Modifiers:
+    *  - onlySelf
+    */
+   function selfTransferFrom(
```
hellwolf, line 326: interface is still wrong

d10r (approved):
I renamed _holder_ to _sender_ in the interface, because that's how it's named elsewhere in the interface. Unfortunately we don't have consistent naming btw. interface and implementation file (see e.g. transferFrom(), thus there's no obvious best way. Should be addressed in the future, but not a blocker here.

hellwolf:
> I renamed _holder_ to _sender_ in the interface, because that's how it's named elsewhere in the interface. Unfortunately we don't have consistent naming btw. interface and implementation file (see e.g. transferFrom(), thus there's no obvious best way. Should be addressed in the future, but not a blocker here.

indeed, naming is different in different contexts for good reason (transfers vs. approvals), but also different in different standards for bad reason (erc20 vs. erc777).

Changing interface name is fine, it shouldn't cause compatibility issues in web3 api, changing event though will affect web3 and also the subgraph.

Also note that the code is almost verbatim from openzeppellin erc777 of some point.

PR #601, opened 2022-02-03
Improve SDK-redux TypeScript performance

kasparkallas:
Why?
The latest preview release makes my laptop's fans whirl when I access Intellisense (the CPU processes go through the roof). Unfortunately, this technical nuance is an issue... because it's very annoying. There's a lot of type inference going on with RTK-Query and we're demanding a lot out of it, putting strain on the TS Language Service to figure out the types.

EDIT: It only seems to be an issue on WebStorm as the code completion is very snappy on Visual Studio Code. I opened an issue for WebStorm here: https://youtrack.jetbrains.com/issue/WEB-54675

How to solve?
Follow the guidelines here: https://github.com/microsoft/TypeScript/wiki/Performance & https://github.com/microsoft/TypeScript/wiki/Performance-Tracing

Right now I split up a file that referenced both `sfApi` & `sfSubgraph` redux slices and as I understand TS goes through the files linearly so it had to figure out `sfApi` types before getting to figure them out for `sfSubgraph`. I could be wrong... Have to test it out with the dev build.


kasparkallas:
Quick update, the issue seems to only be related to Webstorm... which has been become slow for all projects on my computer as of late. I switched to VSCode for now. I created a issue on the Webstorm's issue tracker and created a repository to reproduce the performance issue.

packages/sdk-redux/tsconfig.json
```diff
@@ -52,7 +52,7 @@
         // "experimentalDecorators": true /* Enables experimental support for ES7 decorators. */,
         // "emitDecoratorMetadata": true /* Enables experimental support for emitting type metadata for decorators. */,
         "forceConsistentCasingInFileNames": true,
-        "lib": ["dom", "es2019", "es2020.promise", "es2020.bigint", "es2020.string"],
```
kasparkallas, line 55: I'm pretty sure the DOM lib here is irrelevant. IDE or the build complained a while ago and then I added it... but it doesn't complain anymore and it doesn't make too much sense. The DOM lib is required by Redux Toolkit when generating React Hooks but that shouldn't affect our TSConfig.

hellwolf:
is it better now, do we still want to resolve it?

But perhaps putting it to draft if it's the case:

image

kasparkallas:
> 

Unfortunately hasn't helped much. I believe all the changes here are worthwhile and I would merge it for a development build.

hellwolf:
should we abandon this one @kasparkallas ?

hellwolf:
a bit outdated, we may revisit another time.

PR #604, opened 2022-02-03
Update dependencies for sdk-core sdk-redux subgraph

hellwolf:
Fix hardhat ethers version due to a bug introduced by their patch-version update


0xdavinchee (approved):
gmi

PR #606, opened 2022-02-03
Simplify make liquidation payouts v2

d10r:
proposed code simplification


0xdavinchee (approved):
this looks good to me and tests are passing - looks like it will save gas for plebs and pirates because it will only try to get the reward address if we're in the pic period

hellwolf (approved):

PR #621, opened 2022-02-07
subgraph tests sdk core refactor wip

0xdavinchee:
- removed initialization of contracts (done via framework)
- no more BN.js
- passing framework through functions a lot more

### DIFF
--- a/packages/subgraph/test/helpers/helpers.ts
+++ b/packages/subgraph/test/helpers/helpers.ts
@@ -1,12 +1,11 @@
 import { ethers } from "hardhat";
 import { BaseProvider } from "@ethersproject/providers";
 import { request, gql } from "graphql-request";
-import { Framework } from "@superfluid-finance/sdk-core";
+import { Framework, ConstantFlowAgreementV1 } from "@superfluid-finance/sdk-core";
 import { IMeta, IIndexSubscription } from "../interfaces";
 import { FlowActionType } from "./constants";
 import IResolverABI from "../../abis/IResolver.json";
 import TestTokenABI from "../../abis/TestToken.json";
-import { ConstantFlowAgreementV1 } from "../../typechain/ConstantFlowAgreementV1";
 import { Resolver, TestToken } from "../../typechain";
 import { SignerWithAddress } from "@nomiclabs/hardhat-ethers/signers";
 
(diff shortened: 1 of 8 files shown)


hellwolf:
is it still WIP or need review?

0xdavinchee:
this is good for review now - I think it's best that the refactor occurs incrementally so it's easier to review

packages/subgraph/src/mappingHelpers.ts
```diff
@@ -523,10 +523,14 @@ function updateATSBalanceAndUpdatedAt(
     let superTokenContract = SuperToken.bind(
         Address.fromString(accountTokenSnapshot.token)
     );
-    let newBalance = superTokenContract.balanceOf(
-        Address.fromString(accountTokenSnapshot.account)
+    let newBalanceResult = superTokenContract.try_realtimeBalanceOf(
```
hellwolf, line 526: why being defensive here?

is there anyway to have warning logs in subgraph, and being able to monitor them?

0xdavinchee: If we don't do the try, the whole subgraph breaks, but probably a good idea to put warning logs in the event that it doesn't work, not sure why it would revert though

0xdavinchee: Main change is also from balance of to realtime balance of

kasparkallas: If it doesn't work, it should break, no?

0xdavinchee: Probably not? We're pretty reliant on the subgraph throughout our stack currently and if a revert happens, we don't want everyone else's experience to be impacted until we notice the issue and push a change.

hellwolf: Yea, subgraph currently is built such way that there is no graceful way of warning about things. But Vincent do you experienced some address on some network actually causing the issue (once, or persistently)?

0xdavinchee: no, I don't think I've ever experienced the subgraph going down because of a revert ever since we moved it into maintenance mode-I am not sure in what case a view function would cause a revert if there isn't any require/revert/assert statements in it.

kasparkallas: There is an experimental feature for non-fatal errors: https://thegraph.com/docs/en/developer/create-subgraph-hosted/#non-fatal-errors

PR #623, opened 2022-02-07
Use proxy verification

d10r:


hellwolf:
Rosco says that packages/ethereum-contracts/tasks/etherscan-verify-proxies.sh cannot work at the moment, he will fix it.

PR #627, opened 2022-02-07
revert zero address reward account logic

0xdavinchee:


hellwolf (approved):
let's do the token upgrades again :/

PR #629, opened 2022-02-10
Add `IndexValueCurrent` to IndexSubscription query

kasparkallas:
Why?
To optimize showing `total amount distributed` in consuming applications. Right now, if we show a list of index subscriptions and want to show the total amount distributed then we would have to do _select N + 1_ queries.

Anything else?
In another pull request, I would add `tokenSymbol` to all the entities as well because consuming applications want to display the token almost everywhere but currently only the bare token ID is available which I doubt anyone wants to display.


0xdavinchee (approved):
looks good and this should be helpful

PR #630, opened 2022-02-10
Add `indexTotalUnits` to IndexSubscription query

kasparkallas:
Why?
To optimize showing index subscription's pool percentage in consuming applications. Makes it easy to display a list of subscriptions with the % of the index. Having better overview of pool percentages was a feedback from one of the Ricochet developers.


0xdavinchee (approved):
looks good

PR #634, opened 2022-02-15
add secondsCritical property

0xdavinchee:
**NOTE: requires deploying the subgraph here to feature-matic endpoint and generating the schema.graphql file locally otherwise Github actions will fail.**


0xdavinchee:
deprecated given new subgraph experiment

PR #642, opened 2022-02-18
change solidity pragma in cfa lib to support versions 0.8 & later

saflamini:
@hellwolf this is fix for the CFA Lib prior to the weekend's hackathon. Just changing the pragma statement from ^0.7.0 to > 0.7.0 as I believe using the '^' won't allow for support with 0.8.0 and above.


hellwolf (approved):
image

PR #659, opened 2022-02-28
fix publishing

samirsalem:
This should allow to publish packages again, set -x doesn't prompt any particular error during the execution of the bash scripts unfortunately.


hellwolf:
https://github.com/npm/cli/issues/3993


> As you can see, using npa returns returns GIT SSH result when subpaths are passed to it. If you add a trailing "/" at the end of your path, it should fix the issue:

!!! THIS MAKES ME FUME

hellwolf:
anyway, thanks @samirsalem we should be okay now!

PR #688, opened 2022-03-07
Workflow Improvements (loc and faster pre-commit)

hellwolf:
- loc command for ethereum-contracts
- workspace pre-commit should run only if there are local changes in each workspace package


kasparkallas (approved):
:+1:

Bash noob question, for readability, could `[ ! -z \"$(git status -s .)\" ]` be moved to "tasks/has-changes.sh" and used as `"if tasks/has-changes.sh;then run-s pre-commit:*;else true;fi"`?

kasparkallas (approved):

0xdavinchee (approved):
looks good

PR #691, opened 2022-03-09
Fix nftbillboard

saflamini:
Examples here have been fixed.

Needed to update `GANACHE_PORT` values in the `sh` script so they were hardcoded at port `8545`.
Fixed a failing test inside of the `streaming-call-option`


hellwolf (approved):

saflamini:
@JoshuaTrujillo15 can you give this an approval when you get a chance?

Fixed my bash scripts on the examples (:

PR #693, opened 2022-03-10
informative revertFromReturnedData

### LINKED ISSUE #625: [ETHEREUM-CONTRACTS] CallUtils lack support for custom errors
Solidity [custom errors](https://docs.soliditylang.org/en/v0.8.11/abi-spec.html?highlight=custom%20errors#errors) are new in solidity 0.8.4, and we believe the [`getRevertMsg`](https://github.com/superfluid-finance/protocol-monorepo/blob/3d367b98cd060f751ea5bbb2760c3f6cb052127d/packages/ethereum-contracts/contracts/libs/CallUtils.sol#L14) method does not support propagating custom error messages. 

### Current `getRevertMsg`
```
    function getRevertMsg(bytes memory res) internal pure returns (string memory) {
        // If the _res length is less than 68, then the transaction failed silently (without a revert message)
        if (res.length < 68) return "CallUtils: target reverted";
        // solhint-disable-next-line no-inline-assembly
        assembly {
            // Slice the sighash.
            res := add(res, 0x04)
        }
        return abi.decode(res, (string)); // All that remains is the revert string
    }
```

### Suggested fix

Largely based on solution from the [OZ Addressable library](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/defcf20042479d995a0688bdfd019e44c9e79d76/contracts/utils/Address.sol#L201), and the following [stack overflow comment](https://ethereum.stackexchange.com/a/111187/70970).

```
    function getRevertMsg(bytes memory returndata) internal pure returns (string memory) {
        if (returndata.length == 0) return "CallUtils: target reverted";
        assembly {
           //  Slice away first 32 bits, which store the length of the array
           returndata := add(32, returndata)
        }
        return abi.decode(returndata, (string)); // All that remains is the revert error
    }
```

hellwolf:
Four types of errors are correctly bubbled up:

- "CallUtils: target revert()"
- "CallUtils: target panicked: 0x__"
- Other error string
- Custom errors

### DIFF
--- a/packages/ethereum-contracts/contracts/libs/CallUtils.sol
+++ b/packages/ethereum-contracts/contracts/libs/CallUtils.sol
@@ -7,6 +7,50 @@ pragma solidity 0.8.12;
  */
 library CallUtils {
 
+    /// @dev Bubble up the revert from the returnedData (supports Panic, Error & Custom Errors)
+    /// @notice This is needed in order to get the human-readable revert message from a call
+    /// @param res Response of the call
+    function revertFromReturnedData(bytes memory returnedData) internal pure {
+        if (returnedData.length < 4) {
+            // case 1: catch all
+            revert("CallUtils: target revert()");
+        } else {
+            bytes4 errorSelector;
+            assembly {
+                errorSelector := mload(add(returnedData, 0x20))
+            }
+            //revert RePanic(uint256(uint32(errorSelector)));
+            if (errorSelector == bytes4(0x4e487b71) /* `seth sig "Panic(uint256)"` */) {
+                // case 2: Panics(uint256) (Defined since 0.8.0)
+                // solhint-disable-next-line max-line-length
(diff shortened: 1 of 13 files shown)


packages/ethereum-contracts/contracts/libs/CallUtils.sol
```diff
@@ -7,6 +7,50 @@ pragma solidity 0.8.12;
...
+    function revertFromReturnedData(bytes memory returnedData) internal pure {
+        if (returnedData.length < 4) {
+            // case 1: catch all
+            revert("CallUtils: target revert()");
+        } else {
+            bytes4 errorSelector;
+            assembly {
+                errorSelector := mload(add(returnedData, 0x20))
+            }
+            //revert RePanic(uint256(uint32(errorSelector)));
```
0xdavinchee, line 22: it should just be revert Panic(uint256)?

0xdavinchee (approved):
lgtm

PR #697, opened 2022-03-11
Fix canary build

hellwolf:
- New ganache-cli doesn't like chainId parameter
- Make sure if one workspace command fail, all the rest fail
- Make sure always using `mkdir -p`


0xdavinchee (approved):
lgtm

PR #703, opened 2022-03-14
Use host.getNow to fully abstract time

hellwolf:
We always knew that the governance could pause time some day, but we are not ready yet to give the governance the responsibility of it.

The main motivation for this at the moment to enable fast property based testing such as dapptools, echidna, and perhaps even allowing symbolic execution to reason things when time is a manipulable state of in a test environment.


hellwolf:
It turns out that the HEVM fuzzer or similar would actually increase the time randomly also too in testing mode. So the main intention of this feature branch is not valid, but it is still a good exercise since that is we always wanted to have.

0xdavinchee (approved):
looks good! lgtm!

PR #705, opened 2022-03-15
SDK Core QueryHandler improvements (block filtering, _meta table querying, infinity paging)

kasparkallas:
## Why
In preparation for some SDK redux changes.

### Block filtering

On the Subgraph, every possible entity/event can be queried by block number:
image

This feature enables it for all the QueryHandlers.

**How to use?**
```ts
const {
    data: pagedEvents
} = sfSubgraph.useEventsQuery(
    {
        chainId: queryChainId,
        block: {
            number_gte: 25000000
        }
    }
);
``` 

### Meta table querying
This feature creates a QueryHandler for the Subgraph's default _meta table:

image

**How to use?**
```ts
const meta = new MetaQueryHandler().get({ block: { number: 250000 } });
``` 

### Infinity paging
Iterates all the possible pages on Subgraph and returns them as one result. Definitely something that can be used to shoot oneself in the foot. But nonetheless, I'd like to be able to declaratively tell in my code to "get all the possible results" and "not having all the results can mean error".

**How to use?**
```ts
const {
    data: pagedEvents
} = sfSubgraph.useEventsQuery(
    {
        chainId: queryChainId,
        pagination: {
            take: Infinity
        }
    }
);
```


kasparkallas:
Commit history is messed up because I hadn't merged the latest dev & because of squash. I'll do another PR with cleaner history.

PR #706, opened 2022-03-15
Many small SDK Core improvements

kasparkallas:
## Why?
In preparation for some bigger SDK redux changes. 

### Block filtering

On the Subgraph, every possible entity/event can be queried by block number:
image

This feature enables it for all the QueryHandlers.

**How to use?**
```ts
const {
    data: pagedEvents
} = sfSubgraph.useEventsQuery(
    {
        chainId: queryChainId,
        block: {
            number_gte: 25000000
        }
    }
);
``` 

### Meta table querying
This feature creates a QueryHandler for the Subgraph's default _meta table:

image

**How to use?**
```ts
const meta = await new MetaQueryHandler().get(subgraphClient, { block: { number: 250000 } });
``` 

### "get everything" paging (`AllPaging`)
Iterates all the possible pages on Subgraph and returns them as one result. Definitely something that can be used to shoot oneself in the foot. But nonetheless, I'd like to be able to declaratively tell in my code to "get all the possible results" and "not having all the results can mean error".

**How to use?**
```ts
const {
    data: pagedEvents
} = sfSubgraph.useEventsQuery(
    {
        chainId: queryChainId,
        pagination: {
            take: Infinity
        }
    }
);
```

### Page number paging (`PageNumberPaging`)
Just like `SkipPaging` but with a more UI development friendly API.

**How to use?**
```ts
const {
    data: pagedEvents
} = sfSubgraph.useEventsQuery(
    {
        chainId: queryChainId,
        pagination: {
            pageNumber: 3
            take: pageSize
        }
    }
);
```

### Token symbol to entity queries on Subgraph
Currently the Subgraph queries have been made with the idea that they select minimum amount of field from sub-entities / connected entities, normally just `id`. With tokens, `id` (address) and `symbol` are by far the most used fields so it makes sense to always include them to avoid another query for the symbol.

### Export `Token` (now named `ERC20Token`)
No reason to now export the base class of `SuperToken` because this class can be initialized and used to invoke ERC-20 operations.

### Rename `PagedResult.data` to `PagedResult.items`
With sdk-redux, we get the treaded `data.data` syntax all the time. `data.items` is more readable and less error-prone.

NOTE: We left `PagedResult.data` as an option for now as well, i.e. the data reference is duplicated in both of the properties.

### Add support for `TypedDocumentNode` when using the `SubgraphClient`
Small API change, backwards-compatible, reduces generic type code when passing in a `TypedDocumentNode`.

### Remove `SubgraphClient.batchRequests`
Subgraph doesn't support it...

### Use only ESLint Prettier plugin and remove separate Prettier instance
Because the monorepo root ESLint config is using the plugin but the sdk-core project was not then these got into a bit of a conflict. Don't want to waste time on this and just moved to only using the plugin.


kasparkallas:
- [x] Reminder! Add info to the changelog.

packages/sdk-core/src/pagination.ts
```diff
@@ -27,6 +27,15 @@ export type LastIdPaging = {
     readonly lastId: string;
 };
 
+/**
+ * WARNING: Works only with number literal of {@see Number.POSITIVE_INFINITY}.
+ * WARNING: Works only with the new QueryHandlers.
+ * Recursively gets all the possible Subgraph results.
+ */
+export type InfinityPaging = {
```
hellwolf: rename to -> AllPages

hellwolf: discussion had: maybe iterator API support for querying in the future.

hellwolf (changes requested):

packages/sdk-core/src/pagination.ts
kasparkallas: Renamed to "AllPaging".

0xdavinchee (approved):
looks good! lgtm 🚀

hellwolf (approved):
image

https://xkcd.com/706/

PR #718, opened 2022-03-21
Re-Write continuous-auction dapp tests

yuliyu123:
Re-Write continuous-auction dapp tests by the sdk core & typescript & hardhat.
All functions are tested, coverage reached to 100%. Link to https://github.com/superfluid-finance/super-examples/issues/11


jtriley2p (changes requested):
# 2022-4-4

Code looks good, my only suggestion is to add `.idea` directory in the example to the local `.gitignore`.

yuliyu123:
updated

jtriley2p (approved):

PR #719, opened 2022-03-21
Re-Write Budget NFT tests using new sdk and hardhat

ananya:
Re-Write Budget NFT test using sdk core, typescript, and hardhat.

Issue: https://github.com/superfluid-finance/super-examples/issues/11


jtriley2p (changes requested):
# 2022-04-04

Code looks good, though I suggest removing the `scripts/deploy.ts` file, as it's just an example.

ananya:
updated

saflamini (approved):

jtriley2p (approved):

PR #720, opened 2022-03-22
[ISSUE #702] Access Control List

### LINKED ISSUE #702: [ETHEREUM-CONTRACTS]  CFA ACL Support
CFA ACL (access control list) allows other operators to change flows other than the standard semantics of flow sender/receiver/liquidator.

## Checklist

- [ ] Technical note to be created: "About CFA ACL"
- [x] Implement the CFA ACL per spec of protocol-forum-4
- [x] Implement test cases

## Spec protocol-forum-4

**ABI**

* `CFA.authorizeFlowOperator(token, sender, flowOperator, permission, maxFlowRate)`
* `CFA.authorizeFlowOperatorWithFullControl`  -> `CFA.authorizeFlowOperator(token, sender, ALL, MAX_INT96)`
* `CFA.revokeFlowOperatorWithFullControl` ->  `CFA.authorizeFlowOperator(token, sender, NONE, 0)`

**Event**

* `FlowOperatorUpdated(token indexed, sender indexed, flowOperator indexed, permissions, maxFlowRate, usedMaxFlowRate)`

**Storage Layout**

AgreementData/AgreementContractData of CFA

- FlowData  (256 bits)
   - Id: flowId - `keccak256(abi.encode(sender, receiver))`
   - encoding: ...
- FlowOperatorData (256 bits)
   - Id: flowOperatorId - `keccak256(abi.encode("flowOperator", sender, flowOperator))`
   - encoding: `(permissions: octal bitmask, maxFlowRate: Int96)`
   - trivial case/backward compatibility: data == 0, no permission

0xdavinchee:
# Issue: https://github.com/superfluid-finance/protocol-monorepo/issues/702

### IConstantFlowAgreementV1
* `IConstantFlowAgreementV1` interface file with explanations of new ACL functions added
  * Permission granting functions to allow change flow on behalf of sender
  * "byOperator" functions to change flow on behalf of sender
* `FlowUpdatedV2` and `FlowOperatorUpdated` events added

### ConstantFlowAgreementV1
* Implementation of new ACL functions 
* New `FlowOperatorData` struct which is stored in tokens agreement data
* General grant permissions function and wrapper grant all/revoke all functions as well
* `_createFlow`, `_updateFlow`, `_deleteFlow` internal functions created to move out shared logic between regular change flow (by sender) and "byOperator" change flow

### ConstantFlowAgreementV1.behavior.js / ConstantFlowAgreementV1.test.js
* Added test cases for newly implemented ACL functions:
  * Granting permissions
  * Change flow by operator
* Added new functions to behavior file for validating permissions granting and expecting reverts
* Modified `changeFlow` slightly to properly handle byOperator cases and new FlowUpdatedV2 event

### Others
* `FlowOperatorDefinitions` created in `Definitions.sol`
* `MAXIMUM_FLOW_RATE` and `getFlowOperatorId` added to `TestEnvironment.js`


packages/ethereum-contracts/contracts/agreements/ConstantFlowAgreementV1.sol
```diff
@@ -417,6 +333,422 @@ contract ConstantFlowAgreementV1 is
...
+        address receiver,
+        int96 flowRate,
+        bytes calldata ctx
+    ) 
+        external override
+        returns(bytes memory newCtx)
+    {
+        ISuperfluid.Context memory currentContext = AgreementLibrary.authorizeTokenAccess(token, ctx);
+        // NOTE this may not be necessary...
+        require(currentContext.msgSender != sender, "CFA: You cannot createFlowByOperator as the sender");
```
0xdavinchee: use error msg less than 32 chars

packages/ethereum-contracts/contracts/agreements/ConstantFlowAgreementV1.sol
```diff
@@ -417,6 +345,417 @@ contract ConstantFlowAgreementV1 is
...
+                    ISuperApp(flowVars.sender),
+                    SuperAppDefinitions.APP_RULE_NO_CRITICAL_SENDER_ACCOUNT);
+            }
+            // if the stream we're deleting (possibly liquidating) has a receiver that is a super app
+            // always attempt to call receiver callback
+            if (ISuperfluid(msg.sender).isApp(ISuperApp(flowVars.receiver))) {
+                newCtx = _changeFlowToApp(
+                    flowVars.receiver,
+                    flowVars.token, flowParams, oldFlowData,
+                    newCtx, currentContext, FlowChangeType.DELETE_FLOW);
```
d10r, line 536: I think we now also have to handle the case of the flow sender being a SuperApp with a callback to be invoked

packages/ethereum-contracts/contracts/agreements/ConstantFlowAgreementV1.sol
```diff
@@ -997,4 +1381,68 @@ contract ConstantFlowAgreementV1 is
...
+    (
+        FlowOperatorData memory flowOperatorData
+    )
+        internal pure
+        returns(bytes32[] memory data)
+    {
+        data = new bytes32[](1);
+        data[0] = bytes32(
+            uint256(flowOperatorData.permissions) << 128 |
+            uint256(int256(flowOperatorData.flowRateAllowance))
```
d10r, line 1417: This is potentially dangerous.
When converting negative integers to an unsigned integer, Solidity maps that value to `max_value_of_type - abs(input_value)`. E,g. `uint16(-1)` becomes 65535.  
This means that, against the intuition, you can't rely on the upper 128 bits being 0 after the type conversion to uint256.
Hence, a flowRateAllowance of -1 flips all bits to 1, incl. the permissions bits.

image

In this case, nothing bad may happen because the msg.sender has no incentive to hack themselves this way. Still, `updateFlowOperatorPermissions` doesn't prevent setting a negative flowRateAllowance, allowing for undefined states (reserved bits set). Also, a msg.sender abusing this could set permissions flags which aren't reflected in the emitted event.

Assuming that we'll never want negative values for the flowRateAllowance, it would probably be best to make it an uint in `FlowOperatorData`.

d10r (dismissed):

packages/ethereum-contracts/contracts/agreements/ConstantFlowAgreementV1.sol
hellwolf: Hmm, that's right, but that's an interesting case, but I think it wouldn't break any backward compatibility so we could skip the support of this one, and make a note about it.

packages/ethereum-contracts/contracts/agreements/ConstantFlowAgreementV1.sol
```diff
@@ -417,6 +345,417 @@ contract ConstantFlowAgreementV1 is
...
+        newCtx = _deleteFlow(flowVars, hasPermissions, ctx, currentContext);
+    }
+
+    /// @dev IConstantFlowAgreementV1.updateFlowOperatorPermissions implementation
+    function updateFlowOperatorPermissions(
+        ISuperfluidToken token, 
+        address sender,
+        address flowOperator,
+        uint8 permissions,
+        int96 flowRateAllowance, // flowRateBudget
```
hellwolf, line 681: add a sanity check of the input variable flowRateAllowance to be >= 0

packages/ethereum-contracts/contracts/agreements/ConstantFlowAgreementV1.sol
hellwolf: suggested to have a input parameter check of flowRateAllowance to be >= 0.

And here we could use assert for invariance check.

hellwolf (changes requested):

hellwolf (approved):
https://xkcd.com/720/

image

hellwolf:
@d10r any other things to resolve?

PR #731, opened 2022-04-09
Isa example

saflamini:
New Superfluid example: an employment based loan which allows an employee to create a loan contract which is tied to their salary.

NOTE: need to resolve merge conflicts


saflamini:
@kasparkallas I just pushed a new example to the repo, but it looks like one of the sdk redux examples failed the build workflow. do you mind taking a look?

also - requesting review from @JoshuaTrujillo15 on the employment-loan example. let's discuss tomorrow

jtriley2p (approved):
looks good, mr

PR #743, opened 2022-04-13
SuperToken class abstraction

0xdavinchee:
* no more "test cheating"
* SuperToken class split into three classes: `SuperTokenWithUnderlying`, `SuperTokenWithoutUnderlying`, `NativeAssetSuperToken`
* subgraph deploy scripts


packages/sdk-core/CHANGELOG.md
```diff
@@ -12,7 +12,12 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 - Added `PageNumberPaging` for UI development friendly pagination API
 - Added `AllPaging` to recursively query all results from Subgraph
 - Added support for `TypedDocumentNode` for `SubgraphClient`
+- Expose underlying ethers.js contracts for each class: CFAv1, Host, IDAv1 and `contracts` property in `Framework` class
```
kasparkallas, line 15: :heart_eyes:

packages/sdk-core/CHANGELOG.md
```diff
@@ -12,7 +12,12 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
...
 - Added new ACL function support: authorizing flow operator permissions and create/update/delete flow by operator
+- Added `nativeTokenSymbol` property to `constants.ts`
+- Split `SuperToken` class into: `SuperTokenWithUnderlying`, `SuperTokenWithoutUnderlying` and `NativeAssetSuperToken` classes
```
kasparkallas: "SuperTokenWithUnderlying" & "SuperTokenWithoutUnderlying"

Maybe we have better terminology for this?

@hellwolf

packages/sdk-core/src/Framework.ts
```diff
@@ -249,18 +264,33 @@ export default class Framework {
     };
 
     /**
-     * Load a `SuperToken` class from the `Framework`.
+     * Loads `SuperTokenWithUnderlying`, `SuperTokenWithoutUnderlying` or `NativeAssetSuperToken` class from the `Framework`.
      * @param tokenAddressOrSymbol the `SuperToken` address or symbol (if symbol, it must be on the resolver)
      * @returns `SuperToken` class
      */
-    loadSuperToken = async (
+    loadSuperToken = async <T extends SuperToken>(
```
kasparkallas: I personally think `<T extends SuperToken>` is a bit confusing.

I don't see how `await framework.loadSuperToken<SuperTokenWithUnderlying>("DAIx")` is much better than `await framework.loadSuperToken("DAIx") as SuperTokenWithUnderlying`.

If we were to make it a generic then I would expect more benefit or type-checking from it. Like it should break when I put `NativeAssetSuperToken` but it's not actually one. (It might not be possible to add any more type-checking though...) Right now it's just another way to cast the result.

But if we keep it like this then I would at least set the default type to the base class like so: `T extends SuperToken = SuperToken` because it shouldn't always be necessary to know beforehand whether it's a super token with underlying or not.

packages/subgraph/scripts/deployContractsAndToken.ts
```diff
@@ -30,9 +32,37 @@ async function main() {
...
+    // deploy native super token
+    await deployNativeSuperToken(
+        (x: any) => errorHandler("NativeSuperToken", x),
+        [":", "Mr.Token", "MR", "10000000"],
+        {
+            web3: (global as any).web3,
+            from: Deployer,
+        }
+    );
+    // TODO: figure out why this script is not properly getting the token symbol
```
kasparkallas, line 53: You mentioned that this TODO is OK, right?

packages/sdk-core/CHANGELOG.md
0xdavinchee: si mr

packages/sdk-core/src/SuperToken.ts
```diff
@@ -82,10 +83,15 @@ export default class SuperToken extends ERC20Token {
...
             config: this.settings.config,
         });
-        this.underlyingToken = new ERC20Token(settings.underlyingTokenAddress);
     }
 
-    static create = async (options: ITokenOptions) => {
+    static create = async (
+        options: ITokenOptions
+    ): Promise<
+        | NativeAssetSuperToken
```
hellwolf: should we pro-actively create a new type for this union type already?

packages/sdk-core/CHANGELOG.md
hellwolf: okay, largely reasonable, though NativeAssetSuperToken and SuperTokenWithoutUnderlying technically can overlap.

packages/sdk-core/src/Framework.ts
0xdavinchee: okay-this is fair

packages/sdk-core/CHANGELOG.md
hellwolf: in protocol, it is `createERC20Wrapper` vs `initializeCustomSuperToken`

as a reference, the current classification system is documented in https://www.notion.so/superfluidhq/Classification-of-Super-Tokens-5beace780b5c4d09a5752a3677da3dc0

packages/subgraph/scripts/deployContractsAndToken.ts
0xdavinchee: yeah this is okay

packages/sdk-core/.nycrc.json
```diff
@@ -1,16 +1 @@
...
-        "src/pagination.ts",
-        "src/utils.ts",
-        "src/validation.ts",
-        "src/CustomError.ts",
-        "src/constants.ts",
-        "src/frameworkHelpers.ts",
-        "src/mapGetAllEventsQueryEvents.ts"
-    ]
-}
+{}
```
hellwolf, line 1: grt

hellwolf (commented):
good stuff, just some comments

packages/sdk-core/src/SuperToken.ts
kasparkallas: I would just use the ` SuperToken` base class here, not the union.

packages/sdk-core/CHANGELOG.md
kasparkallas: Is it possible that I've heard the term "native super token" for super tokens that don't have an underlying? Because that's the association I had in my head before this PR.

packages/sdk-core/src/SuperToken.ts
0xdavinchee: `SuperTokenType`, also renamed the classes based on: https://www.notion.so/superfluidhq/Classification-of-Super-Tokens-5beace780b5c4d09a5752a3677da3dc0
*with some slight modifications for consistency:
- SuperTokenWithUnderlying => ERC20WrapperSuperToken => ERC20WrappedSuperToken
- NativeAssetSuperToken => CustomWrappedSuperToken
- SuperTokenWithoutUnderlying => CustomSuperToken

packages/sdk-core/CHANGELOG.md
0xdavinchee: okay, naming discussed and addressed

packages/sdk-core/src/SuperToken.ts
0xdavinchee: naming addressed

0xdavinchee:
mr @saflamini you can use this PR for reference when updating the docs :)

hellwolf:
image

kasparkallas (approved):

hellwolf (approved):
https://xkcd.com/743/

image


this one is good. down with notion.so!

PR #750, opened 2022-04-20
Governance Class added to sdk-core

### LINKED ISSUE #744: [SUBGRAPH] Map governance parameters and contract addresses to an entity
**Why?**
Right now it's not straightforward enough to query the governance parameters and contract addresses. It would be great if there was an entity on Subgraph which contained all the up-to-date information.

**How?**
Governance parameters are a result of projecting `cfav1LiquidationPeriodChangedEvent` & `pppconfigurationChangedEvent` data. NOTE: There seems to be an issue that test networks haven't had a `pppconfigurationChangedEvent` emitted, at least on Subgraph.

Contract addresses can be queried from the resolver.

0xdavinchee:
* Governance class added to getGovernanceParameter
* Tests added for Governance class
* contract moved from get() to member property


packages/sdk-core/src/Governance.ts
```diff
@@ -0,0 +1,48 @@
+import { ethers } from "ethers";
+
+import SuperfluidGovernanceIIABI from "./abi/SuperfluidGovernanceII.json";
+import {
+    IGetGovernanceParametersParams,
+    IWeb3GovernanceParams,
+} from "./interfaces";
+import { SuperfluidGovernanceII } from "./typechain/SuperfluidGovernanceII";
+
+export default class Governance {
```
kasparkallas, line 10: Please export this out of the SDK.

packages/sdk-core/src/Governance.ts
```diff
@@ -0,0 +1,48 @@
...
+        this.hostAddress = hostAddress;
+    }
+
+    /**
+     * Returns the relevant governance parameters
+     * @param providerOrSigner a provider or signer for executing a web3 call
+     * @param tokenAddress governance parameter for a specific token address
+     * @returns
+     */
+    getGovernanceParameters = async ({
```
kasparkallas: I would personally prefer to have separate methods because the user might not need all the data. We're forcing 3 sequential RPC calls. Let's leave it up to the SDK user to choose which governance parameters to get and how to optimize the calls with `Promise.all`. As a rule of thumb, I think we should keep it granular. We can always keep the aggregated call as an extra.

Thoughts?

0xdavinchee: Good point, will make this change.

packages/sdk-core/src/Governance.ts
0xdavinchee: Si

0xdavinchee:
@kasparkallas pls review again when you can

kasparkallas (approved):
Love it

PR #751, opened 2022-04-20
no zero subs for IDA + tests

0xdavinchee:
* do not allow IDA publishers to updateUnits/deleteSubscription zero address subscriber
* do not allow anyone to do a claim for zero address
* tests added to capture these cases


packages/ethereum-contracts/test/contracts/agreements/InstantDistributionAgreementV1.test.js
```diff
@@ -309,6 +310,48 @@ describe("Using InstantDistributionAgreement v1", function () {
...
+                    indexId: DEFAULT_INDEX_ID,
+                });
+
+                await expectRevert(
+                    shouldUpdateSubscription({
+                        testenv: t,
+                        superToken,
+                        publisherName: "alice",
+                        indexId: DEFAULT_INDEX_ID,
+                        subscriberAddress: ZERO_ADDRESS,
```
hellwolf: since subscriberAddress to ZERO_ADDRESS is only for expectRevert test cases, we don't need to modify the behaviour test code. I would say in these case just do the call directly through the js-sdk wrapper and expect the revert.

hellwolf (changes requested):

hellwolf (approved):

hellwolf:
https://xkcd.com/751/

image

PR #752, opened 2022-04-25
fix test subgraph ci

0xdavinchee:
* race condition was fixed in ci.feature but not ci.canary
* TODO - move to test subgraph entirely using Hardhat node - one test currently breaks for some strange reason

### DIFF
--- a/.github/workflows/ci.canary.yml
+++ b/.github/workflows/ci.canary.yml
@@ -161,6 +161,12 @@ jobs:
         uses: actions/setup-node@v1
         with:
           node-version: ${{ matrix.node-version }}
+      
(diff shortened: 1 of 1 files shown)


hellwolf:
> TODO - move to test subgraph entirely using Hardhat node - one test currently breaks for some strange reason

Not of the scope of the current PR right?

0xdavinchee:
> > TODO - move to test subgraph entirely using Hardhat node - one test currently breaks for some strange reason
> 
> Not of the scope of the current PR right?

nope, not this one, same thing w/ the sdk-core tests, they're both using ganache still

hellwolf (approved):
https://xkcd.com/752/

image

PR #755, opened 2022-04-27
[sdk-core] Use environment variable to decide which Subgraph release tag to use

kasparkallas:
Why?
With the latest sdk-core dev build, some Subgraph queries break because the deployed schema does not have all the properties of the queries.

Open for discussion or alternatives.

### DIFF
--- a/.github/workflows/ci.canary.yml
+++ b/.github/workflows/ci.canary.yml
@@ -217,6 +217,8 @@ jobs:
 
       - name: Build
         run: yarn build
+        env:
+          SUBGRAPH_RELEASE_TAG: dev
 
       - name: Setup canary package versions locally
         run: |
--- a/packages/sdk-core/src/Framework.ts
+++ b/packages/sdk-core/src/Framework.ts
@@ -39,7 +39,7 @@ type SupportedProvider =
 // instead of having to pass it in every single time
 export interface IFrameworkOptions {
     chainId?: number;
-    customSubgraphQueriesEndpoint?: "v1" | "dev" | "feature" | string;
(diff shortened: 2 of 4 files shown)


hellwolf:
originally there is an option in the design to override your subgraph query endpoint url, for testing purposes. If we don't have that, let's add that, so that the user of the sdk could override it, instead of hacking internally here.

hellwolf:
@0xdavinchee is such option being omitted so far?

hellwolf:
if knowing the full url is the problem, perhaps allowing several options in combination:

- being able to override the entire url
- being able to just override a portion of it (e.g. the staging part: v1 vs dev vs feature)

kasparkallas:
There is an option to override the query endpoint but it's pretty annoying to do if the SDK is in use for all the networks with the default.

> if knowing the full url is the problem, perhaps allowing several options in combination:
> 
>     * being able to override the entire url
>     * being able to just override a portion of it (e.g. the staging part: v1 vs dev vs feature)

Yes, I agree with these. 

What's you opinion on doing _dev_ SDK builds with upcoming Subgraph features, available only on "-dev-" Subgraph endpoints, but we publish the SDK with "v1" as the default?

hellwolf (approved):
Good solution, so the canary builds of sdk* would use also dev endpoints of subgraph-v1.

https://xkcd.com/755/

image

kasparkallas:
Okay, I was a bit confused here -- it's a runtime environment variable, not a build time one, so this solution is not going to work...

The confusion came from the fact that with CRA/Next.js environment variables like `process.env.SUBGRAPH_RELEASE_TAG` get embedded into the code...

PR #757, opened 2022-04-27
[sdk-core] Prioritize Wrapper Super Token creation over Native Asset one

kasparkallas:
Why?
For "MATICx" & "xDAIx", `NativeAssetSuperToken` was being created when it's actually a `WrapperSuperToken`.

Solution?
The solution is to change the order of object creation in the `create` method. The first thing is to check whether the token has an underlying token and when it does that means it's a `WrapperSuperToken`.

Todo?
- [ ] Write a test for this case.

### DIFF
--- a/packages/sdk-core/src/SuperToken.ts
+++ b/packages/sdk-core/src/SuperToken.ts
@@ -129,9 +129,13 @@ export default abstract class SuperToken extends ERC20Token {
                 networkName,
             };
 
-            const tokenSymbol = await superToken
-                .connect(options.provider)
-                .symbol();
+            if (underlyingTokenAddress !== ethers.constants.AddressZero) {
+                return new WrapperSuperToken(options, {
+                    ...settings,
+                    underlyingTokenAddress,
+                });
+            }
+
             const resolverData = chainIdToResolverDataMap.get(chainId) || {
                 subgraphAPIEndpoint: "",
                 resolverAddress: "",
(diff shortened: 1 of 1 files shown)


packages/sdk-core/src/SuperToken.ts
```diff
@@ -141,6 +145,9 @@ export default abstract class SuperToken extends ERC20Token {
             const nativeTokenSymbol = resolverData.nativeTokenSymbol || "ETH";
             const nativeSuperTokenSymbol = nativeTokenSymbol + "x";
 
+            const tokenSymbol = await superToken
+                .connect(options.provider)
+                .symbol();
             if (nativeSuperTokenSymbol === tokenSymbol) {
```
kasparkallas, line 151: Not a fan of this condition for checking whether it's a `NativeAssetSuperToken` -- I wish it was more concrete than just checking the token symbol and making assumptions.

0xdavinchee:
can you pls write a test for this - other than that it looks good.

packages/sdk-core/src/SuperToken.ts
0xdavinchee: one other check I can think of is checking if a function unique to `NativeAssetSuperToken`'s exist: e.g. `upgradeByETH` - it requires casting the superToken contract as any and I'm not sure if it actually works.

kasparkallas:
We're gonna stick with keeping these types of tokens as `NativeAssetSuperToken` as the _wrapped_ part is deprecated.

Application developers should discard the part about the underlying token when querying from the Subgraph.

PR #760, opened 2022-04-28
[sdk-core] Embed default Subgraph release tag at Canary build time

kasparkallas:
Follow-up to: https://github.com/superfluid-finance/protocol-monorepo/pull/755#issuecomment-1111872919


0xdavinchee (approved):
lgtm

PR #763, opened 2022-04-28
ACL redundant sender param cleanup

0xdavinchee:
* removed redundant 
* modify test behavior cases
* remove redundant test case (same as setting themselves as operator)
* sdk core api and test cases cleanup


hellwolf (approved):
https://xkcd.com/763/

image

PR #766, opened 2022-05-02
A separated ISuperAppRegistry and a  SuperAppRegistryDelegate

### LINKED ISSUE #672: App whitelisting by trusted deployer
We should be able to trust a deploy for up to a period of time, this way we could lightening our overhead of app whitelisting process without bigger compromise in its setup.

## TODOs

- [ ] Make utils/SuperAppFactoryDelegate.sol part of the full flow (deploy-framework.js).
  - Register register resolver key "SuperAppFactoryDelegate-v1"
  - Piggyback: register resolver key "BatchLiquidator-v1"
  - Design point: should we allow registerAppDelegate to be called outside of app's constructor
- [ ] Update test code and documentations to use the new paradigm
- [ ] ...?

--

## Deprecated Notes

- Extend `Superfluid.registerApp` with new logic that it will check if the tx.origin(the deployer) is whitelisted and within its expiration date.
- Extend `SuperfluidBaseGoverance` with deployer whitelisting with expiration date logic.

hellwolf:
To have a smoother transition from white-listed Super App framework to no whitelisted ones, this can be a solution to unify the super app deployment workflow:

On testnets:

- Superfluid implements ISuperAppRegistry, and the apps can continue to use it directly,
- But they are also going to be asked to add new code to select the app registry during its constructor. For reason that will be evident below.

On mainnets:

- A SuperAppRegistryDelegate is created for each host contract,
- Such contract may white-list deployer with an expiration date,
- And Super App register their app via this delegate instead.

### DIFF
--- a/packages/ethereum-contracts/contracts/interfaces/superfluid/ISuperAppRegistry.sol
+++ b/packages/ethereum-contracts/contracts/interfaces/superfluid/ISuperAppRegistry.sol
@@ -1,0 +1,81 @@
+// SPDX-License-Identifier: AGPLv3
+pragma solidity >= 0.8.0;
+import { ISuperApp } from "./ISuperApp.sol";
+
+
+/**
+ * @title Super App Registry
+ * @author Superfluid
+ *
+ * NOTE:
+ * The superfluid host implements the Super App Registry interface, but it might have a white-list.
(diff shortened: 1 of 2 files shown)


hellwolf:
continue from #769 instead

PR #768, opened 2022-05-02
add missing providers (GH env) for L2s

d10r:
added missing keys for GH actions


hellwolf (approved):
https://xkcd.com/768/

image

PR #769, opened 2022-05-02
don't disable registration key on use

d10r:
Proposed alternative to solving the problem of app registration currently being too cumbersome.  

Rationale:
We eventually want to make registration permissionless, but aren't yet confident enough (reentrancy...) to do so.  
Also, the initial mechanism (a gov action for every app deployment) has become too burdensome. 
We thus considered adding a registration method whitelisting a _deployer_.  
Handing someone a one-time registration already has trust assumption. If they were malicious and wanted to exploit the protocol with a SuperApp, a single deployment would suffice.
Thus proposed solution: allow reuse of registration keys. That is effectively the same as whitelisting a deployer. That way we have minimal code changes and don't spend more time dealing with a temporary mechanism.
If we don't want to give prolonged permission to a deployer, gov can still revoke the key after use.

A more thorough application of this change would involve removal of the actual key and renaming of the interfaces to better reflect what it's doing (whitelist deployer). But since this mechanism isn't expected to be around long term, it's probably not worth it.

**Update**:

status quo:
interfaces unchanged. Semantics of registration key changed to be reusable, but with expiration date.  
Expiration date is set to 90 days from now by default (when using `scripts/gov-create-new-app-registration-key.js` without argument `EXPIRATION_TS`).
The convenience function `gov.whiteListNewApp` isn't used, should be deprecated / removed on next gov contract update. We could instead change and add a param for expiration date, but not really needed (we can keep using the generic `setConfig` instead).


packages/ethereum-contracts/scripts/gov-create-new-app-registration-key.js
```diff
@@ -14,17 +16,34 @@ const {
  * @param {Address} options.from Address to deploy contracts from
  * @param {boolean} options.protocolReleaseVersion Specify the protocol release version to be used
  *
- * Usage: npx truffle exec scripts/create-new-app-registration-key.js : {DEPLOYER} {REGISTRATION_KEY}
+ * Note: the key itself doesn't have much meaning, it could be "stolen" from a broadcast tx anyway.
+ * But since it's bound to a deployer address, that doesn't really matter.
+ *
+ * Usage: npx truffle exec scripts/create-new-app-registration-key.js : {DEPLOYER} {REGISTRATION_KEY} [EXPIRATION_TS]
```
hellwolf, line 22: REGISTRATION_KEY -> APP_KEY?

packages/ethereum-contracts/scripts/gov-create-new-app-registration-key.js
```diff
@@ -54,8 +73,12 @@ module.exports = eval(`(${S.toString()})({
...
     );
     console.log("App key", appKey);
+    console.log("Expiration date", new Date(expirationTs * 1000)); // print human readable
 
+    // Note that we are NOT using gov.whiteListNewApp here, because it doesn't support setting
+    // an expiration date and will eventually be deprecated.
+    // Instead we use the lower level setConfig directly.
     await sendGovernanceAction(sf, (gov) =>
-        gov.whiteListNewApp(sf.host.address, appKey)
+        gov.setConfig(sf.host.address, ZERO_ADDRESS, appKey, expirationTs)
```
hellwolf, line 82: what do we do with the `host.whiteListNewApp`?

hellwolf: >  The convenience function gov.whiteListNewApp isn't used, should be deprecated / removed on next gov contract update. We could instead change and add a param for expiration date, but not really needed (we can keep using the generic setConfig instead).

ah.

hellwolf: we could already remove it, and next scheduled update will get it removed.

packages/ethereum-contracts/scripts/gov-create-new-app-registration-key.js
d10r: I don't think so. The appKey is derived from it, as before:
```
    const appKey = web3.utils.sha3(
        web3.eth.abi.encodeParameters(
            ["string", "address", "string"],
            [
                "org.superfluid-finance.superfluid.appWhiteListing.registrationKey",
                deployer,
                registrationkey,
            ]
        )
    );
```

hellwolf: yea, my bad.

hellwolf (approved):
https://xkcd.com/769/

image

packages/ethereum-contracts/scripts/gov-create-new-app-registration-key.js
d10r: done

hellwolf:
```
  1) Superfluid Ownable Governance Contract
       #2 configurations
         #2.5 whiteListNewApp:
     TypeError: governance.whiteListNewApp is not a function
      at Context.<anonymous> (test/contracts/gov/SuperfluidGovernanceII.test.js:465:28)
      at processImmediate (node:internal/timers:466:21)
```

hellwolf:
@d10r 

```

  1) Superfluid Ownable Governance Contract
       #2 configurations
         #2.5 whiteListNewApp:
     TypeError: governance.whiteListNewApp is not a function
      at Context.<anonymous> (test/contracts/gov/SuperfluidGovernanceII.test.js:465:28)
      at processImmediate (node:internal/timers:466:21)
```

PR #771, opened 2022-05-03
Reapply createProviderForOpenEthereum hack

hellwolf:


d10r (approved):
is unfortunate, but gud

PR #782, opened 2022-05-06
SDK-Core Version 0.4.0

0xdavinchee:
* CHANGELOG.md contains all the changes
* Deleted a lot of the README.md information - best route is to just redirect them to docs instead of having to duplicate information in two places and sometimes forgetting to update readme


packages/sdk-core/CHANGELOG.md
```diff
@@ -29,6 +31,19 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 ### Internal
 - Use `eslint-plugin-prettier` over separate `prettier` instance
 
+### Breaking
+- The `SuperToken` class is now an abstract base class and no longer contains the functions `upgrade` and `downgrade`.
+- `underlyingToken` is possibly undefined on `SuperToken`: `WrapperSuperToken` has `underlyingToken`, but `PureSuperToken` and `NativeAssetSuperToken` do not.
+> NOTE: These changes are due to the split of `SuperToken` into `WrapperSuperToken`, `PureSuperToken` and `NativeAssetSuperToken` classes.
+  - Migration: if you are using TypeScript, you will need to specify the type of token you are attempting to load:
```
kasparkallas: No, you don't...? Most cases, I'm fine with the base `SuperToken` type.

packages/sdk-core/CHANGELOG.md
```diff
@@ -29,6 +31,19 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
...
+  ```
+  import { NativeAssetSuperToken } from "@superfluid-finance/sdk-core";
+  ...
+  const nativeAssetSuperToken = await framework.loadSuperToken<NativeAssetSuperToken>("ETHx");
```
kasparkallas: I still dislike that this doesn't do any runtime check that it's actually of that type. It's actually pretty bad... It's easy for consumers to make expectations of the type here and run into issues later when it's not actually of that type.

packages/sdk-redux/package.json
```diff
@@ -48,7 +48,7 @@
     },
     "peerDependencies": {
         "@reduxjs/toolkit": "^1.6.0 || ^1.7.0",
-        "@superfluid-finance/sdk-core": "~0.3.2"
+        "@superfluid-finance/sdk-core": "~0.4.0"
```
kasparkallas, line 51: :heart:

packages/sdk-core/CHANGELOG.md
0xdavinchee: I had this issue in testing

packages/sdk-core/CHANGELOG.md
0xdavinchee: pls elaborate

packages/sdk-core/CHANGELOG.md
kasparkallas: What issue? In `sdk-redux`, we never know which type it is so we do `instanceof` instead. Or not do at all if we _only_ need to use the base type methods.

packages/sdk-core/CHANGELOG.md
kasparkallas: This will not fail:
```ts
const nativeAssetSuperToken = await framework.loadSuperToken<NativeAssetSuperToken>("USDCx");
```

It should.

hellwolf:
can you please piggy back this change in:

```
$ git diff packages/ethereum-contracts/package.json
diff --git a/packages/ethereum-contracts/package.json b/packages/ethereum-contracts/package.json
index 9e6b1314..45016e95 100644
--- a/packages/ethereum-contracts/package.json
+++ b/packages/ethereum-contracts/package.json
@@ -56,7 +56,7 @@
     "dependencies": {
         "@decentral.ee/web3-helpers": "0.5.3",
         "@openzeppelin/contracts": "4.5.0",
-        "@superfluid-finance/js-sdk": "0.5.12",
+        "@superfluid-finance/js-sdk": "0.5.13",
         "@truffle/contract": "4.5.5",
         "ethereumjs-tx": "2.1.2",
         "ethereumjs-util": "7.1.4",
```

And update the yarn.lock. @msoni89 let's remember to update monorepo packages correctly next time, I forgot to review this one carefully.

packages/sdk-core/package.json
```diff
@@ -1,6 +1,6 @@
 {
     "name": "@superfluid-finance/sdk-core",
-    "version": "0.3.2",
+    "version": "0.4.0",
```
hellwolf, line 3: no CHANGELOG?

packages/sdk-core/CHANGELOG.md
0xdavinchee: the issue is that it complains because underlyingToken might be undefined, so when I access it I get a TypeError

packages/sdk-core/CHANGELOG.md
0xdavinchee: what would the fix be then?

kasparkallas: A:
```ts
const nativeAssetSuperToken = await framework.loadSuperToken("USDCx", SuperTokenType.NativeAdded); // And do some TS magic here.
```
B:
```ts
const nativeAssetSuperToken: NativeAssetSuperToken = await framework.loadNativeAssetSuperToken("USDCx"); // Throws if not native asset super token
const wrapperSuperToken: WrapperSuperToken = await framework.loadWrapperSuperToken("USDCx"); // Throws if not wrapper super token
const pureSuperToken: pureSuperToken = await framework.loadPureSuperToken("USDCx"); // Throws if not pure super token
const superToken: SuperToken = await framework.loadSuperToken("USDCx"); // This one should know how to delegate to other calls, i.e. you can always use this if you don't know the token type upfront.
```

I know you did something similar to B at first but lost the general one which we definitely shouldn't do. But if the general one delegated to the other calls and we re-use the code then I think this is a good solution.

kasparkallas (approved):

packages/sdk-core/CHANGELOG.md
kasparkallas: We went with option B.

hellwolf (approved):
https://xkcd.com/782/

image

PR #784, opened 2022-05-06
SDK-Core and Subgraph Release Cleanup

0xdavinchee:
# Subgraph Cleanup
* make `FlowOperatorUpdatedEvent` entity immutable
* fix a comment regarding the `AccountTokenSnapshot` entity ID

# SDK-Core Cleanup
* add new explicit initialization functions for different SuperToken class types
* add tests for new initialization functions including expected fail cases

### DIFF
--- a/packages/sdk-core/src/Framework.ts
+++ b/packages/sdk-core/src/Framework.ts
@@ -11,7 +11,11 @@ import InstantDistributionAgreementV1 from "./InstantDistributionAgreementV1";
 import Operation from "./Operation";
 import Query from "./Query";
 import { SFError } from "./SFError";
-import SuperToken from "./SuperToken";
+import SuperToken, {
+    NativeAssetSuperToken,
+    PureSuperToken,
+    WrapperSuperToken,
+} from "./SuperToken";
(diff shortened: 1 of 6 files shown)


0xdavinchee:
@saflamini
there are three new super token initialization functions (in addition to loadSuperToken which loads the generic SuperToken base class):
- `loadWrapperSuperToken`
- `loadNativeAssetSuperToken`
- `loadPureSuperToken`

to load super tokens explicitly based on the type of SuperToken being loaded, it might be worth adding this to docs, thoughts?

kasparkallas (approved):

saflamini:
> @saflamini there are three new super token initialization functions (in addition to loadSuperToken which loads the generic SuperToken base class):
> 
> * `loadWrapperSuperToken`
> * `loadNativeAssetSuperToken`
> * `loadPureSuperToken`
> 
> to load super tokens explicitly based on the type of SuperToken being loaded, it might be worth adding this to docs, thoughts?

Agree. I like this approach better tbh as it accounts for different super token types.

We really need to start converting naming conventions everywhere over to this Native v Pure terminology 😅

PR #786, opened 2022-05-08
patch lend function

saflamini:
change lend() function such that 
- calling lend automatically transfers funds from lender to borrower
- the check on whether or not enough collateral is `>=` instead of `==`


hellwolf:
@JoshuaTrujillo15 @ngmachado ?

jtriley2p (approved):

PR #788, opened 2022-05-09
Unify app registration

d10r:
skip registration key check on networks with app whitelisting disabled


packages/ethereum-contracts/contracts/interfaces/superfluid/ISuperfluid.sol
```diff
@@ -196,15 +199,16 @@ interface ISuperfluid {
     event AppRegistered(ISuperApp indexed app);
```
hellwolf, line 198: we should mark registerApp as deprecated in the comments?

per https://docs.soliditylang.org/en/develop/natspec-format.html

maybe `custom: deprecated` is the closed we could use to mark it as deprecated

hellwolf: also then we have to start to use registerAppWithKey in all our examples, and communicate this change to the devx folks.

hellwolf (changes requested):

packages/ethereum-contracts/contracts/interfaces/superfluid/ISuperfluid.sol
d10r: I know we wanted to deprecate it.
However working on it my thinking went like: `registerWithKey` won't make much sense anymore once (if) we switch to permissionless App registration. Thus it's probably better to just leave `registerApp` in place, but tell current App devs they can use the same register method on both testnet and mainnet (could be with key or by factory).

hellwolf: but it makes sense "right now", because otherwise the app would need to use registerApp for testnet deployment, and registerWithKey on mainnet deployment. Since solidity doesn't have pre compiler, app would need to fiddle with this. And we don't really have examples for that.

d10r: done, I also updated the docs and created https://github.com/superfluid-finance/protocol-monorepo/wiki/Super-App-White-listing-Guide

packages/subgraph/scripts/deployContractsAndToken.ts
```diff
@@ -2,7 +2,7 @@ import {ethers} from "hardhat";
 import deployFramework from "@superfluid-finance/ethereum-contracts/scripts/deploy-framework";
 import deployTestToken from "@superfluid-finance/ethereum-contracts/scripts/deploy-test-token";
 import deploySuperToken from "@superfluid-finance/ethereum-contracts/scripts/deploy-super-token";
-import deployNativeSuperToken from "@superfluid-finance/ethereum-contracts/scripts/deploy-unlisted-native-super-token";
+import deployNativeSuperToken from "@superfluid-finance/ethereum-contracts/scripts/deploy-unlisted-pure-super-token";
```
hellwolf, line 5: a minor fix here: deployPureSuperToken

packages/ethereum-contracts/contracts/interfaces/superfluid/ISuperfluid.sol
```diff
@@ -1,5 +1,5 @@
 // SPDX-License-Identifier: AGPLv3
-pragma solidity >= 0.8.0;
+pragma solidity >= 0.8.2;
```
hellwolf, line 2: we might want to do a replaceAll for interface files

hellwolf (approved):
some minor fixes suggested pending, then it's good to go

PR #790, opened 2022-05-09
[778] SlotsBitmapLibrary Property Test Fuzzing

### LINKED ISSUE #778: [ETHEREUM-CONTRACTS] Add propreties test for SlotsBitmapLibrary
SlotsBitmapLibrary is a standalone EVM library, mainly used by distribution agreement(s) in general for managing a device called "subscription bitmap" where up to 256 subscriptions can be approved for each subscriber. Test this using foundry property testing and integrate this to CI, check "ConstantFlowAgreementV1.prop.sol" for inspiration.

0xdavinchee:
* fuzz test SlotsBitmapLibrary contract
* basic tests which check invariants without fuzzing
* fuzzing test which checks basic invariants in an expected way (add + check invariants then remove + check invariants)
* randomized sequence fuzzing test which checks basic invariants in a randomized way (add/remove + check invariants then add/remove + check invariants)


packages/ethereum-contracts/test/foundry/libs/SlotsBitmapLibrary.prop.sol
```diff
@@ -0,0 +1,204 @@
...
+            // findEmptySlotAndFill when number is greater than 0 AND the slotIds isn't maxed out
+            if (
+                instructions[uint256(i)] > 0 && slotIds.length < type(uint8).max
+            ) {
+                // we find an empty slot and fill it and length should be one more now
+                _findEmptySlotAndFill(superToken, subscriber, fakeId);
+                (slotIds, ) = _listData(superToken, subscriber);
+                assertEq(slotIds.length, currentLength + 1);
+            }
+        }
```
hellwolf, line 254: should add an invariance check

packages/ethereum-contracts/test/foundry/libs/SlotsBitmapLibrary.prop.sol
```diff
@@ -0,0 +1,204 @@
...
+    /// @dev Subscriber state slot id for storing subs bitmap
+    uint256 private constant _SUBSCRIBER_SUBS_BITMAP_STATE_SLOT_ID = 0;
+    /// @dev Subscriber state slot id starting point for subscription data
+    uint256 private constant _SUBSCRIBER_SUB_DATA_STATE_SLOT_ID_START =
+        1 << 128;
+
+    constructor() {
+        vm.startPrank(subscriber);
+        // Deploy ERC1820
+        vm.etch(
```
hellwolf: use the new lib re ERC1820

hellwolf (changes requested):
some requests

packages/ethereum-contracts/test/foundry/libs/SlotsBitmapLibrary.prop.sol
hellwolf: or put another way, what is the expected return result from _listData? I would refactor the test to check exactly what is the expected values. E.g. (pseudo code, not solidity) :

```
_listData(superToken, subscriber, [0, 1, 2, 4] /* expected slotIds */ )
```

0xdavinchee: okay, from what I understand, this will require recreation of the expected slot ids array using the same or similar logic as listData though (without assembly) as the `listData` function filters out any "empty" slots.

is this ok?

hellwolf: Ah yea, you would need to use solidity to recreate that, too cumbersome.

Let's try another trick to reuse the functions to create other type of invariances, e.g.

reverse(reverse(x)) == x

and in this example, a clever invariance could be:

- with the slots list returned:
- clear all of them
- and list again, the slots list should now be empty
- an additional good assert to make this even better is to add an assert in `clearSlot` so that if the slot was not set, it assets:

```
        assert(subsBitmap & (1 << uint256(slotId)) != 0); // PLEASE double check if this is statement correct, I didn't test it
        slotData[0] = bytes32(subsBitmap & ~(1 << uint256(slotId)));
```

hellwolf: This is also a good "property" of slotsbitmap in pseudo code:

```
l1 <- listSlots
foreach i  in l1: clearSlot i
l2 <- listSlots
assert l2 is empty
```

Does it make sense?

0xdavinchee: yeah, makes sense

packages/ethereum-contracts/test/foundry/libs/SlotsBitmapLibrary.prop.sol
0xdavinchee: fixed

packages/ethereum-contracts/test/foundry/libs/SlotsBitmapLibrary.prop.sol
0xdavinchee: added the property to test, are we sure we want to add that assert in `clearSlot` - our fuzzing doesn't break when adding it in.

hellwolf: yes we do, it's better to asserts and burn a bit gas than letting bad cases survive in the wild.

hellwolf: okay.

I would try a bit more:

1. make testRandomSlotsBitmapLibraryBehavior a shared function _createRandomSlots
2. combine _createRandomSlots + testCreateSlotsAndClearAll
3. a new test case: combine _createRandomSlots + createTooManySlots, and the last creation should revert

Not sure about the purpose of those non "property" single test cases , but no harm to keep them neither.

hellwolf:
```
    function testRandomSlotsBitmapLibraryBehavior(
        uint8 numSlots,
        int256[] memory instructions
    ) public {
 ```

numSlots is not used, instead I would do assume(instructions.length > 0)

hellwolf (approved):
https://xkcd.com/790/

image

PR #794, opened 2022-05-09
Subgraph 1.3.2 patch

0xdavinchee:
* Subgraph 1.3.1 was incorrectly handling ATS creation in a specific case leading to ATS with null addresses. This was because ATS was being created even though the fromAddress was the zero address
* This lets us know that no zero address account entities are being added (expected behavior) - **ACTUALLY THIS ISN'T TRUE ON POLYGON** - this is because there are events where the subscriber was the zero address and we are creating `IndexSubscription` and therefore `Account` entities for zero address subscribers

### DIFF
--- a/packages/subgraph/package.json
+++ b/packages/subgraph/package.json
@@ -1,6 +1,6 @@
 {
     "name": "@superfluid-finance/subgraph",
-    "version": "1.3.1",
+    "version": "1.3.2",
     "description": "Subgraph for the Superfluid Ethereum contracts.",
     "homepage": "https://github.com/superfluid-finance/protocol-monorepo/tree/dev/packages/subgraph",
     "repository": {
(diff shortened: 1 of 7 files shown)


hellwolf:
A test case?

0xdavinchee:
> A test case?

redeployed to optimism this query is working now:
```
{
  accountTokenSnapshots {
    account { id }
  }
}

```

https://thegraph.com/hosted-service/subgraph/superfluid-finance/protocol-dev-optimism-mainnet 

but this def makes me think there are some subgraph invariants to be added

hellwolf:
have a test case in our testsuite for no future regression.

kasparkallas (approved):

kasparkallas (commented):
We got discussing why make an exception for zero addresses anyway. Seems to just add complexity.

PR #798, opened 2022-05-10
Subgraph Fix Pt2

### LINKED ISSUE #796: [SUBGRAPH] Add test case which queries all Events/HOLs for ILightEntity
* A bug was recently discovered in the Subgraph after pushing 1.3.1 where the `Account` property was not being properly added
   * The expectation was that the subgraph would not be able to index to completion when attempting to create an entity without a link, but it's possible to create an entity pointing to an address, but if the entity doesn't exist on the other side, it breaks

Proposal: Add a single test case which queries all event/HOL/aggregate entities and try to get the ILightEntity of each just to ensure nothing is null.

e.g.

```
query GetFlowUpdatedEvents {
    flowUpdatedEvents {
        stream {
             id
        } 
    }
}
```

0xdavinchee:
The immutable tag seemed to be breaking the global `events` query. 
If I had to guess, this was likely due to our `FlowUpdatedEvent` entity not containing this tag (due to the requirement of going back and updating it again in a subsequent event: `FlowUpdatedExtension`).
Removing this tag allows the global `events` query to be called without breaking.

### Learnings
Upon further investigation, it is clear that when you have multiple entities implementing an interface, in our case the `Event` interface. **If some of the entities are immutable and a single one is not, it will break when trying to query for all events.** - Tested by keeping the immutable tags, but removing the `implements Event` for `FlowUpdatedEvent` which is not an immutable event.

Another bug (previous issue) is that you can create a **non-nullable** property on an entity which links to another entity via the ID of the entity (e.g. `account: Account!` on `AccountTokenSnapshot` BUT this other entity doesn't need to exist and the Graph will successfully index and this error will only arise when you try to make a nested query.

### DIFF
--- a/packages/subgraph/test/helpers/helpers.ts
+++ b/packages/subgraph/test/helpers/helpers.ts
@@ -9,9 +9,8 @@ import {
     ITestModifyFlowData,
 } from "../interfaces";
 import { FlowActionType } from "./constants";
-import IResolverABI from "../../abis/IResolver.json";
 import TestTokenABI from "../../abis/TestToken.json";
-import { Resolver, TestToken } from "../../typechain";
+import { TestToken } from "../../typechain";
 import { SignerWithAddress } from "@nomiclabs/hardhat-ethers/signers";
 import { BigNumber } from "ethers";
 
--- a/packages/subgraph/test/helpers/helpers.ts
+++ b/packages/subgraph/test/helpers/helpers.ts
@@ -7,6 +7,7 @@ import {
     IIndexSubscription,
     ITestUpdateFlowOperatorData,
     ITestModifyFlowData,
+    ISubgraphErrors,
 } from "../interfaces";
 import { FlowActionType } from "./constants";
 import TestTokenABI from "../../abis/TestToken.json";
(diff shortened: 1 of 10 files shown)


hellwolf:
immutable bag would be nice too

kasparkallas (approved):
Created an issue for The Graph team here: https://github.com/graphprotocol/graph-node/issues/3553

packages/subgraph/test/helpers/helpers.ts
```diff
@@ -136,11 +137,20 @@ export const subgraphRequest = async <T>(
...
-        const response = await request<T>(
+        const response = await request<T & ISubgraphErrors>(
             "http://localhost:8000/subgraphs/name/superfluid-test",
             query,
             variables
         );
+
+        if (response.errors) {
+            throw new Error(
+                `Failed call to subgraph with query ${query}, error thrown: ${JSON.stringify(
```
kasparkallas, line 148: The `query` is not always a string, it's often a GraphQL AST object which is JSON basically. Changing this to "Failed call to subgraph with query ${JSON.stringify(query)}..." is a simple fix.

PR #801, opened 2022-05-10
fix query syntax

0xdavinchee:
* this query actually worked as it was previously which is why the tests passed, but this syntax makes more sense


0xdavinchee:
seems better to piggyback other subgraph fixes, this pr seems too small to merge by itself

hellwolf (approved):
https://xkcd.com/801/

image

PR #803, opened 2022-05-10
CfAv1Library ACL

jtriley2p:


d10r:
@JoshuaTrujillo15 why remove the flow functions?

hellwolf (commented):
many now have some vested interest in the flow function, we will gather today briefly to talk about it

jtriley2p:
As far as I know, it's used in Degendogs and in two test files in the monorepo. Since I likely won't be at this discussion, the following is my case in favor of removal:

---

The original `flow` function only covered two permutations, one that uses neither userdata nor ctx, the other being one that uses ctx but not userdata. In order to cover all permutations, we should have 4 `flow` functions and ideally we should also have 4 `flowByOperator` functions to cover all cases in the new ACL update.

Aside from number of functions, the current implementation neither checks nor indicates when the `desiredFlowRate` and the `existingFlowRate` are both zero. This leads to a case where if both are zero, `flow` just tries to call `deleteFlow` and fails at the CFA contract. If the decision is made to keep the flow functions, I would like to ask that we do not attempt to call `deleteFlow` if both flow rates are zero and instead update the `flow` function to return a `flowRateDelta` so developers can choose whether or not their intended behavior is to revert when the `flowRate` doesn't change.

kobuta23:
Let's please merge this today. 

Flow function is inherently opinionated and an extra thing to maintain (8 permutations due to userdata+context = lots of test cases). On top of this it can revert unexpectedly, which is problematic as it could be used in a termination callback. 

We'll move it somewhere in the docs or repo so developers can copy paste it if they want, but keep it out of the library.

hellwolf:
Flow doesn't have to have all the permutations, since/and cfalib can be opinionated.  But more importantly, it will break some people's code, though we can deal with it.

I will just merge as it is in interest of time. But be prepare to deal with backward compatibility of things we release more often going forward.

hellwolf (approved):
https://xkcd.com/803/

image

PR #804, opened 2022-05-10
create folder structure for examples

saflamini:
1) Add new examples to examples folder readme
2) Add note about JS SDK in examples readme
3) Reorganize examples folder such that examples which have implementations in 2 dev environments are placed into a single folder


saflamini:
cc @JoshuaTrujillo15 this has to be something simple that I'm missing, no?

Getting the following error when this action runs: 
`tradeable-cashflow/tradeable-cashflow-truffle: No such file or directory`

I've added the paths to each example in the build-examples github workflow file, and it looks like the script does indeed try to `cd` into the right directory before running `sh build-and-test.sh`

But maybe I'm missing something?

hellwolf:
I made a fix to the ci script

```shell
          cd "$(basename ${{ matrix.example }})"
```
this was required.

saflamini:
> cc @JoshuaTrujillo15 this has to be something simple that I'm missing, no?
> 
> Getting the following error when this action runs: `tradeable-cashflow/tradeable-cashflow-truffle: No such file or directory`
> 
> I've added the paths to each example in the build-examples github workflow file, and it looks like the script does indeed try to `cd` into the right directory before running `sh build-and-test.sh`
> 
> But maybe I'm missing something?

UPDATE
- Permissions have been fixed so that the budget nft can be included in the github action
- The build examples script for the github action was fixed by @hellwolf 
- I have temporarily removed the continuous auction example due to a mysterious failed test. It is a complex example that is not used often. It will be better to remove it for now and revisit its structure later.

@JoshuaTrujillo15 can you review when you have time?

ngmachado (approved):
LGTM. 

Worry about duplication of code and maintaining always two versions.

saflamini:
> LGTM.
> 
> Worry about duplication of code and maintaining always two versions.

@ngmachado - agree. I have moved several examples into an archive folder with this in mind, and kept only a single live truffle example which is part of our build workflow.

I have also:
-  re-written the tradeable cashflow truffle example in the sdk core
- re-written the rewards distribution token using the IDAv1 Library

It looks like I need one more review for this to be merged. @hellwolf, @0xdavinchee or @JoshuaTrujillo15 would you be able to review this?

hellwolf (approved):
i will approve for now, but the main question is:

- Having duplicated contracts code between hardhat/truffle flavor is not ideal, and error prone if we intend to update some of the contract.
   - the contracts should be largely the same between truffle/hardhat environment.
   - test cases can be written differently.
   - consider to have a way of getting them in sync or reduce number project having dual versions
   - (I prefer a bit more this) or have two configurations in the same project, but let each development framework has own test suite. Downside of this is that there is slightly more information load to people cloning the project, since they likely need to prune something out afterwards (I still don't think it's a big deal, if people intend to take it forward more seriously after cloning)

Other side questions:

- What's with the archive thing, e.g. is flow lottery not useful?

saflamini:
> i will approve for now, but the main question is:
> 
> * Having duplicated contracts code between hardhat/truffle flavor is not ideal, and error prone if we intend to update some of the contract.
>   
>   * the contracts should be largely the same between truffle/hardhat environment.
>   * test cases can be written differently.
>   * consider to have a way of getting them in sync or reduce number project having dual versions
>   * (I prefer a bit more this) or have two configurations in the same project, but let each development framework has own test suite. Downside of this is that there is slightly more information load to people cloning the project, since they likely need to prune something out afterwards (I still don't think it's a big deal, if people intend to take it forward more seriously after cloning)
> 
> Other side questions:
> 
> * What's with the archive thing, e.g. is flow lottery not useful?

Thank you @hellwolf 

> * What's with the archive thing, e.g. is flow lottery not useful?

The archive is for several projects which are not often used or referenced. The /examples folder was getting unwieldy with many different examples. The flowLottery, streaming option, etc haven't been used often and can be a bit more complicated to understand. 
I also temporarily removed these projects from the standard github action workflow. I did this in part because I couldn't figure out how to make sure that they're run properly in the action when they're in a separate folder, and also because we are not intending to maintain these projects at the same level as the examples which are not in the archive folder.

On this point:
> * Having duplicated contracts code between hardhat/truffle flavor is not ideal, and error prone if we intend to update some of the contract.

I agree, so I actually ended up either removing the duplicate projects or moving them into the 'archive,' with the exception of the tradable cashflow. Many of the older Truffle examples also used the js sdk for testing, which isn't something we want to highlight. In the case of the tradable cashflow example, we could technically include both truffle and hardhat into a single project directory. 
However, I do think that this could cause confusion for early career devs who need to navigate a development environment which contains two frameworks in one. You can, in theory, always subtract items when you clone the repo and fork the project, but I think that the better UX is to split them in two (especially if it's only for a single example).

PR #805, opened 2022-05-11
[RFR] 654 added a property to order events chronologically

### LINKED ISSUE #654: [SUBGRAPH] Add a property to order events chronologically
**Why?**

I'd like to do a simple query and retrieve events in chronologically correct order. Currently ordering events by block number or timestamp does not give perfect results because many events can be included in the same block. Log index is used to create ordering inside the block but just adding `logIndex` to events won't suffice because Subgraph doesn't support ordering by more than 1 field.

**How?**

Add a string field to events that combines both block number and log index. Should be pretty simple change to common mapping logic of events in `/packages/subgraph`. For example: `“${blocknumber}-${logIndex}”`.

**Notes**

* The log index needs to be zero padded in the front to avoid "${blocknumber}-12" being ordered before "${blocknumber}-2". So it should be "${blocknumber}-012" and "${blocknumber}-002" respectively.
* How much zero padding? https://ethereum.stackexchange.com/questions/99910/whats-the-maximum-number-of-logs-per-transaction

msoni89:


packages/subgraph/src/utils.ts
```diff
@@ -1,27 +1,17 @@
...
+import {ISuperToken as SuperToken} from "../generated/templates/SuperToken/ISuperToken";
+import {Resolver} from "../generated/ResolverV1/Resolver";
+import {IndexSubscription, StreamRevision, Token, TokenStatistic,} from "../generated/schema";
 
 /**************************************************************************
  * Constants
  *************************************************************************/
-
 export const BIG_INT_ZERO = BigInt.fromI32(0);
+export const PADDING_LEFT_LENGTH = BigInt.fromI32(6);
```
0xdavinchee: this is unused - can be removed

packages/subgraph/test/helpers/helpers.ts
```diff
@@ -371,8 +371,11 @@ export const modifyFlowAndReturnCreatedFlowData = async (
     const block = await data.provider.getBlock(txnResponse.blockNumber);
     const timestamp = block.timestamp;
     await waitUntilBlockIndexed(txnResponse.blockNumber);
-
-    const { flowRate, deposit } = await data.framework.cfaV1.getFlow({
+    const transactionReciept = await txnResponse.wait();
```
0xdavinchee: small typo: `transactionReciept` => `transactionReceipt`

packages/subgraph/test/helpers/helpers.ts
```diff
@@ -426,7 +430,11 @@ export const updateFlowOperatorPermissions = async (
     const block = await data.provider.getBlock(txnResponse.blockNumber);
     const timestamp = block.timestamp;
     await waitUntilBlockIndexed(txnResponse.blockNumber);
-    return { timestamp, txnResponse };
+    const transactionReciept = await txnResponse.wait();
```
0xdavinchee: ditto above

packages/subgraph/test/helpers/testers.ts
```diff
@@ -632,11 +638,18 @@ async function executeIDATransactionByTypeAndWaitForIndexer(
 
     const block = await provider.getBlock(txnResponse.blockNumber);
     await waitUntilBlockIndexed(txnResponse.blockNumber);
-
+    const transactionReciept = await txnResponse.wait();
```
0xdavinchee: ditto above

packages/subgraph/test/helpers/helpers.ts
msoni89: resolved

packages/subgraph/test/helpers/helpers.ts
```diff
@@ -455,7 +456,11 @@ export const clipDepositNumber = (deposit: BigNumber, roundingDown = false) => {
...
         : deposit.and(toBN(0xffffffff)).isZero()
-        ? 0
-        : 1;
+            ? 0
+            : 1;
     return deposit.shr(32).add(toBN(rounding)).shl(32);
 };
+
+export const getOrder = (blockNumber?: number, logIndex?: number) => {
+    return blockNumber! * 10000 + logIndex!
```
0xdavinchee: I think it will be good to add a comment that this number is defined as `ORDER_MULTIPLIER` in `packages/subgraph/src/utils.ts`

0xdavinchee (approved):
LGTM

PR #806, opened 2022-05-12
Fix codecov source files

0xdavinchee:
Note that the CI passed flag doesn't appear, but source files are visible now.


hellwolf (approved):
https://xkcd.com/806/


image

PR #819, opened 2022-05-15
Avalanche Support Added

0xdavinchee:
* Avalanche support added for ethereum-contracts (hardhat), sdk-core in constants.ts and subgraph (config/, networks.json, addresses.template.ts)
* subgraph `README.md` cleanup
* deploy-all-networks variable renaming


packages/ethereum-contracts/hardhat.config.ts
```diff
@@ -36,6 +36,10 @@ const config: HardhatUserConfig = {
             url: "http://0.0.0.0:8545/",
             chainId: 1337,
         },
+        avalanche: {
```
hellwolf: what are these for actually, is hardhat used for anything network related?

0xdavinchee: well, I guess I sometimes use it when trying to do something quick with hardhat console - but probably nothing other than that

hellwolf: ah right, makes sense. The alias system though is only implemented in truffle-config.js as a technical debt. Since one doesn't necessarily define those environment variables in alias.

packages/sdk-core/src/constants.ts
```diff
@@ -36,18 +51,19 @@ export const AUTHORIZE_FULL_CONTROL =
...
-    10, // OPTIMISM MAINNET
-    42, // KOVAN
-    69, // OPTIMISM KOVAN
-    100, // GNOSIS
-    137, // MATIC
-    42161, // ARBITRUM ONE
-    43113, // AVALANCHE FUJI
-    80001, // MUMBAI
-    421611, // ARBITRUM RINKEBY
+    ROPSTEN_CHAIN_ID, // ROPSTEN
```
hellwolf: Start to use superfluid flavor of canonical network names.

hellwolf (approved):
ts starting to smell like verbose java, ewww, soon there will be ts-beans.

PR #821, opened 2022-05-16
Production deployment script fixes

hellwolf:


d10r:
TODO: check what happens on update runs of deploy-framework.sh with mainnets if `disableTestGovernance` isn't set.

hellwolf:
In the script top comment session, I would like to have a list of "process.env|options" switches used in the deployment script and their purposes.

hellwolf:
@d10r is this one sufficient for BNB deployment, it was a patch given before by Betrand I believe

d10r:
> TODO: check what happens on update runs of deploy-framework.sh with mainnets if `disableTestGovernance` isn't set.

would deploy an instance of `TestGovernance` and register it. Not an issue beyond wasting some gas.  
After a deployment, we update config files (with resolver addrs) anyway, that's when `disableTestGovernance` has to be set.

@hellwolf this should be good for BNB deployment, yes

> In the script top comment session, I would like to have a list of "process.env|options" switches used in the deployment script and their purposes.

Is there a script specifically where you miss it?
I noticed that only `deploy-framework` accepts all config items both with `options` object and (overriding) env var. Is that a limitation somewhere?

d10r (approved):

0xdavinchee (approved):

PR #822, opened 2022-05-16
[RFR] Added criticalAtTimestamp entity to subgraph v1

### LINKED ISSUE #774: [SUBGRAPH] Add optimistic liquidation estimation property to `AccountTokenSnapshot`
Add a new property in the `AccountTokenSnapshot` entity, called "maybeCriticalAtTimestamp" which should tell us the timestamp the user will run out of funds in the specific token. This can be calculated based on balance, netFlow, updatedAtTimestamp, and needs to be recalculated every time these values change.

The "maybeCriticalAtTimestamp" value can be used to construct fairly simple queries showing all critical accounts:

```{
  accountTokenSnapshots(where: {
    criticalAtTimestamp_lt:"1635258762"
  }){
    account {
      id
    }
  }
}```

While there might be some false positives (the claim function isn't currently indexed) this should still be a very good start to create any solvency related dashboard / alert system, without relying on client-side processing or web3

### PR DESCRIPTION
**Pull request by msoni89**

### CODE REVIEW

File: packages/subgraph/src/utils.ts [hunk 1]
```diff
@@ -11,7 +11,7 @@ export const BIG_INT_ONE = BigInt.fromI32(1);
 export const ZERO_ADDRESS = Address.zero();
 export let MAX_FLOW_RATE = BigInt.fromI32(2).pow(95).minus(BigInt.fromI32(1));
 export const ORDER_MULTIPLIER = BigInt.fromI32(10000);
-
+export const MAX_SAFE_MILLISECOND = BigInt.fromI64(8640000000000); //In seconds, https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date#the_ecmascript_epoch_and_timestamps
```
0xdavinchee: maybe this should be called `MAX_SAFE_SECONDS` as the value is in seconds, then the comment can point out that the max safe milliseconds for the date object is: 8640000000000000 and include the link

msoni89: Done

0xdavinchee:
LGTM

0xdavinchee (approved):
LGTM

PR #826, opened 2022-05-17
sdk-core patch

0xdavinchee:
* patch schema.graphql file pointing at v1-matic
* remove all instances of `logIndex` and `order` from files they were added to per https://github.com/superfluid-finance/protocol-monorepo/wiki/Contributor-Modifications-Guide
* added getAllEvents query test
* bumped sdk-core version


packages/sdk-core/test/5_subgraph.test.ts
```diff
@@ -0,0 +1,20 @@
...
+        const subgraphEndpoint =
+            "https://api.thegraph.com/subgraphs/name/superfluid-finance/protocol-v1-matic";
+        const { frameworkClass } = await setup({
+            subgraphEndpoint,
+        });
+        framework = frameworkClass;
+    });
+
+    it("Should be able to make the getAllEvents query", async () => {
+        const events = await framework.query.listEvents({});
```
kasparkallas: Maybe select just like 10 and assert that the list is not empty?

kasparkallas (approved):

packages/sdk-core/test/5_subgraph.test.ts
```diff
@@ -0,0 +1,21 @@
...
+        const { frameworkClass } = await setup({
+            subgraphEndpoint,
+        });
+        framework = frameworkClass;
+    });
+
+    it("Should be able to make the getAllEvents query", async () => {
+        const events = await framework.query.listEvents({ }, { take: 10 });
+        expect(events.data.length).to.be.greaterThan(0);
+    });
```
kasparkallas, line 20: I guess there it could be a bit more clear why this test is necessary, like "events table is the most common table to break on Subgraph changes" or something.

0xdavinchee: ok true - I also just tested this test on dev and it broke as expected - when using dev-xdai (which is now rolled backed, this broke as the subgraph is lagging behind the sdk-core) [CURRENT SITUATION]
I also tried testing on this branch, but instead hitting an endpoint which still has the order/logIndex properties [NEW SUBGRAPH ENDPOINT w/ OLD SDK-CORE RELEASE CASE]

PR #827, opened 2022-05-17
[Subgraph] Get correct Subgraph schema for dev & stable release builds

kasparkallas:
Why?
We had a situation where stable SDK release got published with wrong Subgraph schema, leading to a Subgraph query failing. The purpose of this PR is to automatically get the correct Subgraph schema as a build step which will hopefully lead to either build or tests failing.

### DIFF
--- a/.github/workflows/handler.publish-release-packages.yml
+++ b/.github/workflows/handler.publish-release-packages.yml
@@ -42,7 +42,11 @@ jobs:
       - name: Build package
         run: |
           yarn install --frozen-lockfile
+          yarn --cwd packages/sdk-core set-default-subgraph-release-tag
+          yarn --cwd packages/sdk-core generate-graphql-schema:v1
           yarn build
+        env:
+          SUBGRAPH_RELEASE_TAG: v1
 
       - name: Publish ethereum-contracts package
         if: env.PUBLISH_ETHEREUM_CONTRACTS == 1
--- a/.github/workflows/ci.canary.yml
+++ b/.github/workflows/ci.canary.yml
@@ -262,6 +262,7 @@ jobs:
       - name: Build
         run: |
           yarn --cwd packages/sdk-core set-default-subgraph-release-tag
+          yarn --cwd packages/sdk-core generate-graphql-schema:dev
           yarn build
         env:
           SUBGRAPH_RELEASE_TAG: dev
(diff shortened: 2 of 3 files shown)


.github/workflows/handler.publish-release-packages.yml
```diff
@@ -42,7 +42,11 @@ jobs:
       - name: Build package
         run: |
           yarn install --frozen-lockfile
+          yarn --cwd packages/sdk-core set-default-subgraph-release-tag
+          yarn --cwd packages/sdk-core generate-graphql-schema:v1
           yarn build
+        env:
+          SUBGRAPH_RELEASE_TAG: v1
```
kasparkallas, line 49: This is not necessary but I made it explicit.

.github/workflows/handler.publish-release-packages.yml
```diff
@@ -42,7 +42,11 @@ jobs:
       - name: Build package
         run: |
           yarn install --frozen-lockfile
+          yarn --cwd packages/sdk-core set-default-subgraph-release-tag
```
kasparkallas, line 45: This solution is not optimal because tests will not be run after this step. It will make the build fail though (i.e. it would have prevented the case we had).

hellwolf:
This is a bad idea. Package built should be the package tested. Instead of fetching schema before publishing, one could validate against the "correct" version as a test step instead.

hellwolf (changes requested):
This is a bad idea. Package built should be the package tested. Instead of fetching schema before publishing, one could validate against the "correct" version as a test step instead.

kasparkallas:
Miao, the canary build _does_ run the tests against the correct built version after fetching the schema.

The stable release currently doesn't because I didn't find a place to put it where I _know_ it's the stable release build and it gets tested afterwards. That's why I commented it myself, that's why I set this PR as a draft. Open for feedback or help here.

So... is it a bad idea to get the schema altogether as a build step? Or you just meant that the stable release is currently not optimally handled and we shouldn't do half of a solution?

.github/workflows/handler.publish-release-packages.yml
kasparkallas: As a clarification, the code generation would break.

hellwolf:
replaced by #835

PR #830, opened 2022-05-18
[RFR] create account token snapshot log for each account token snapshot updates

### LINKED ISSUE #775: [SUBGRAPH] Create AccountTokenSnapshotLog for each AccountTokenSnapshot update
Add `AccountTokenSnapshotLog` entity every time we update `AccountTokenSnapshot` entity.
This allows us to easily query historical changes of an account-token pair.
Can choose to add other changing values to the entity, not just the balance. 
I would omit reverse lookups (except AccountTokenSnapshot)

## Note

This is a workaround for that we cannot access time machines of AccountTokenSnapshot as queries. https://github.com/graphprotocol/graph-node/issues/3480

msoni89:


packages/subgraph/schema.graphql
```diff
@@ -1260,6 +1266,86 @@ type AccountTokenSnapshot @entity {
...
+    """
+    totalNumberOfActiveStreamsSoFar: Int!
+
+    """
+    The number of all-time closed streams.
+    """
+    totalNumberOfClosedStreamsSoFar: Int!
+
+    """
+    The current (as of updatedAt) number of subscriptions with units allocated to them tied to this `account`.
```
0xdavinchee, line 1301: as of timestamp* same w/ some of the other comments below

0xdavinchee: as of timestamp* same w/ some of the other comments below

packages/subgraph/schema.graphql
```diff
@@ -1196,6 +1196,12 @@ type AccountTokenSnapshot @entity {
     updatedAtTimestamp: BigInt!
     updatedAtBlockNumber: BigInt!
     # ---------------------------------- state ----------------------------------
+
+    """
+    isLiquidationEstimateOptimistic, If they have `0` subcription with no units it return true or else false.
```
0xdavinchee: I think currently the mapping logic is the other way around, if subscriptionWithUnits > 0, it is true. This is fine and I think it makes sense especially since we want the default value to be false.
It would be good to mention here that optimistic (if this boolean is true) means that it is the earliest time the user will be liquidated as they may receive ongoing distributions which aren't tracked by the subgraph.
Also small typo: `subcription` => `subscription`

packages/subgraph/src/mappings/idav1.ts
```diff
@@ -520,6 +529,8 @@ export function handleSubscriptionUnitsUpdated(
     subscription.save();
 
     createSubscriptionUnitsUpdatedEntity(event, subscription.id, oldUnits);
+    createAccountTokenSnapshotLogEntity(event, event.params.subscriber, event.params.token, "SubscriptionUnitsUpdated");
+    createAccountTokenSnapshotLogEntity(event, event.params.subscriber, event.params.token, "SubscriptionUnitsUpdated");
```
0xdavinchee, line 533: I believe this is duplicated here? I think one of them is meant for the publisher.

packages/subgraph/src/utils.ts
```diff
@@ -277,3 +277,22 @@ export function getOrder(
     return blockNumber.times(ORDER_MULTIPLIER).plus(logIndex);
 }
 
+/**************************************************************************
+ * Log entities util functions
+ *************************************************************************/
+
+export function createLogID(
+    logPrefix: string,
+    aggregateId: string,
```
0xdavinchee: I think it would be good to rename this `accountTokenSnapshotId`

packages/subgraph/test/interfaces.ts
```diff
@@ -324,6 +323,30 @@ export interface IAccountTokenSnapshot extends IBaseAggregateEntity {
...
+    readonly totalNumberOfClosedStreamsSoFar: number;
+    readonly totalSubscriptionsWithUnitsSoFar: number;
+    readonly totalAmountStreamedSoFar: string;
+    readonly totalAmountTransferredSoFar: string;
+    readonly totalDepositSoFar: string;
+    readonly totalInflowRateSoFar: string;
+    readonly totalNetFlowRateSoFar: string;
+    readonly totalOutflowRateSoFar: string;
+    readonly maybeCriticalAtTimestampSoFar: string;
+    readonly  account: ILightEntity;
```
0xdavinchee: extra space, sorry my OCD 😅

packages/subgraph/test/validation/aggregateValidators.ts
```diff
@@ -96,6 +98,68 @@ export const validateATSEntity = (
     ).to.equal(expectedTotalAmountTransferredUntilUpdatedAt);
 };
 
+export const validateASTEntityAndItsLogEntry = (
+    graphATSData: IAccountTokenSnapshot
+) => {
+    const accountTokenSnapshotLogs: IAccountTokenSnapshotLog[] = graphATSData.accountTokenSnapshotLogs;
```
0xdavinchee: this can maybe be squished into one line:
` const accountTokenSnapshotLog: IAccountTokenSnapshotLog = graphATSData.accountTokenSnapshotLogs[0];`

0xdavinchee (changes requested):
looks dope overall, just a few changes

0xdavinchee (approved):
looks good LGTM

packages/subgraph/src/utils.ts
```diff
@@ -277,3 +277,22 @@ export function getOrder(
     return blockNumber.times(ORDER_MULTIPLIER).plus(logIndex);
 }
 
+/**************************************************************************
+ * Log entities util functions
+ *************************************************************************/
+
+export function createLogID(
```
kasparkallas, line 284: It would have been neat if this incremented based on account updates...

packages/subgraph/schema.graphql
```diff
@@ -1260,6 +1266,86 @@ type AccountTokenSnapshot @entity {
     account: Account!
     token: Token!
     flowOperators: [FlowOperator!]! @derivedFrom(field: "accountTokenSnapshot")
+    accountTokenSnapshotLogs: [AccountTokenSnapshotLog!]! @derivedFrom(field: "accountTokenSnapshot")
+}
+
+"""
+AccountTokenSnapshotLog: Historical entries of AccountTokenSnapshot updates.
+"""
+type AccountTokenSnapshotLog @entity {
```
kasparkallas, line 1275: Shouldn't this be marked immutable somewhere? @0xdavinchee

0xdavinchee: this is not super necessary tbh

packages/subgraph/src/utils.ts
0xdavinchee: I guess this may be useful so the IDs can be sortable?

packages/subgraph/schema.graphql
kasparkallas: But it's supposed to be an immutable entity ;)

0xdavinchee: let's add this to an issue and give the HOL entities an interface to implement and do this once the graph fixes the issue w/ mixing immutable/non-immutable

packages/subgraph/src/utils.ts
kasparkallas: The database would automatically order the ID-s a bit better, yes, but it depends on how the database is set up, regarding how it orders "-5" vs "-12". If it prioritizes shorter length then the ID-s would be perfectly sorted at the time of insertion, yes. This can have some query performance implications.

The other small advantage is that it's easy to see how many updates the account has had. This might make planning for queries easier in some cases... Also could be useful information in general.

I'm not saying this is a big deal at all... Just discussing and saying an indexer based ID would have felt more natural to me. :)

msoni89: What do you mean by indexer-based ID?

We could append an order id as part of the ID (generated based on https://github.com/superfluid-finance/protocol-monorepo/pull/805).

kasparkallas: I mean like id-1, id-2, id-3, id-4, id-5... etc after every change.

msoni89: Do you want it to store in memory or a database?

packages/subgraph/schema.graphql
kasparkallas: Apparently it got fixed: https://github.com/graphprotocol/graph-node/issues/3553 ;) Maybe hasn't shown up in hosted service yet.

packages/subgraph/schema.graphql
```diff
@@ -1260,6 +1266,86 @@ type AccountTokenSnapshot @entity {
...
+    id: ID!
+    timestamp: BigInt!
+    blockNumber: BigInt!
+
+    transactionHash: Bytes!
+    logIndex: BigInt!
+    order: BigInt!
+    triggeredByEventName: String!
```
kasparkallas, line 1283: @0xdavinchee Could we use `triggeredByEvent: Event!` here?

0xdavinchee: sure, but it may be useful to have both properties, one for filtering and the other for the information

packages/subgraph/src/utils.ts
kasparkallas: I was thinking just having a `version` or `updateCounter` on AccountTokenSnapshot entity which gets put into the ID of the log entry.

Do you guys think it could be useful? It feels natural to me...

packages/subgraph/schema.graphql
kasparkallas: I guess... although in this case, the event ID actually contains all the same information, except the order, but of course it's not as explicit.

By the way, it seems that the Graph team is quite close to implementing filtering by child entity: https://github.com/graphprotocol/graph-node/pull/3184

Once that gets merged, we can have a more normalized schema, i.e. less duplicated fields just for filtering.

packages/subgraph/src/utils.ts
msoni89: Yes, why not. 

Sure let's add one more property to track updates count, please create a separate ticket. Also regarding `triggered event: Event!` Please mention what properties you want to keep in the HOL entity.

msoni89: One problem is, that the code is pretty much scattered updating from multiple places and due to which maybe be counter or version possibly need to pass a parameter.

PR #832, opened 2022-05-18
tradeable_cashflow hot-fuzz integration

hellwolf:


d10r (approved):
At some point we should probably make an npm package bcs if including the hot-fuzz package from git dev branch as per instructions, `yarn install` tends to complain about failed integrity check.

PR #840, opened 2022-05-23
Last CI fix 

0xdavinchee:
- should be the last CI fix


hellwolf (approved):
Clean commit history in "dev":

image
image

CI workflow fixes also in "dev":

image
image

.github/workflows/call.test-subgraph-on-previous-sdk-core-versions.yml
```diff
@@ -33,14 +33,17 @@ jobs:
           node-version: 16.x
           
       - name: "Install contract dependencies"
+        if: inputs.local-subgraph-url != ''
```
kasparkallas: Why is this everywhere? Why is `local-subgraph-url` needed for CI?

EDIT: It got removed as I posted. :smile:

PR #841, opened 2022-05-23
initialize logic contracts on deployment

d10r:
as recommended by OZ: https://github.com/OpenZeppelin/openzeppelin-contracts/blob/release-v4.5/contracts/proxy/utils/Initializable.sol#L31


packages/ethereum-contracts/contracts/upgradability/UUPSProxiable.sol
```diff
@@ -9,6 +9,11 @@ import { Initializable } from "@openzeppelin/contracts/proxy/utils/Initializable
  */
 abstract contract UUPSProxiable is Initializable {
 
+    // This constructor makes sure logic contracts deriving from it are initialized on deployment,
+    // as recommended by the openzeppelin docs of `Initializable`
+    // solhint-disable-next-line no-empty-blocks
+    constructor() initializer {}
```
hellwolf: can we also set code address to address(0)?

d10r: It's already zero by default

hellwolf:
I still find OZ'z [onlyProxy](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/proxy/utils/UUPSUpgradeable.sol):

```
    /**
     * @dev Check that the execution is being performed through a delegatecall call and that the execution context is
     * a proxy contract with an implementation (as defined in ERC1967) pointing to self. This should only be the case
     * for UUPS and transparent proxies that are using the current contract as their implementation. Execution of a
     * function through ERC1167 minimal proxies (clones) would not normally pass this test, but is not guaranteed to
     * fail.
     */
    modifier onlyProxy() {
        require(address(this) != __self, "Function must be called through delegatecall");
        require(_getImplementation() == __self, "Function must be called through active proxy");
        _;
    }
```

could provide a bit more actual value in our case.

I am not quite certain what cases the initializer in logic contract could help with.

d10r:
`onlyProxy` is useful for flagging specific functions.  
What I was looking for is a solution which removes access to all restricted methods (e.g. `onlyGov` flagged ones) in the logic contracts by default, thereby minimizing the attack surface.  
The `initializer` modified on the Proxiable constructor would perfectly achieve that goal.
The problem is that some of our logic contracts are supposed to work in non-proxied mode too, e.g. for testing/local dev.

The 2nd best solution I can currently think of which also takes that into account is to run `initialize` on the logic contracts in the deploy script. It's a bit less elegant and involves the theoretical risk of somebody DoSing deployment runs (by frontrunning the initialize txs), but since there's nothing to be gained from doing that, it's imo a good solution too.

hellwolf:
I see.

Also alternatively, `UUPSProxiable._updateCodeAddress` could be improved:

1. the logic contract (which is assumed to be a `UUPSProxiable`) has a `castrate` instead of initializing in the constructor. 
2. `castrate` is called in the `_updateCodeAddress` to castrate the logic contract from being ever used.

Choice of name here is serious, not being melodramatic.

d10r:
Wasted more hours on this without finding a satisfying solution.
Problem with `_updateCodeAddress`: doesn't help with new deployments, only with updates of existing ones.
Lets see if we can tackle it in Prague.

d10r:
Finally found a way which - while not as implicit as initially attempted - afaik does the job in a rather simple and non-invasive way.
For the castration of the governance logic contract, I'll add the call to a deploy script which is on another branch.

packages/ethereum-contracts/contracts/gov/SuperfluidGovernanceBase.sol
```diff
@@ -63,12 +63,21 @@ abstract contract SuperfluidGovernanceBase is ISuperfluidGovernance
...
+            UUPSProxiable(address(hostNewLogic)).castrate();
         }
         for (uint i = 0; i < agreementClassNewLogics.length; ++i) {
             host.updateAgreementClass(ISuperAgreement(agreementClassNewLogics[i]));
+            UUPSProxiable(address(agreementClassNewLogics[i])).castrate();
         }
         if (superTokenFactoryNewLogic != address(0)) {
             host.updateSuperTokenFactory(ISuperTokenFactory(superTokenFactoryNewLogic));
+
+            // the factory logic can be updated for non-upgradable hosts too,
```
hellwolf, line 75: hmm, should we stick to this case at all? Not sure anymore why it was like this in the first place, and I don't think it had practical reason

d10r: My understanding is that it's used for testing only.
Also checking that code I wonder what the point of using a proxy for the factory is. Since it's always upgradable in the host and carries no state beyond a reference to a SuperToken logic contract - which changes with a new factory anyway - it could imo always be used without proxy. Only difference to consider then is that its address would change on updates, to be supported e.g. by the subgraph.

d10r: The ability to update it for non-upgradable hosts is needed at the moment bcs it's the only way a factory can be attached to the host.

hellwolf: > bcs it's the only way a factory can be attached to the host.

is it for a good reason or it is something we can change?

hellwolf: ah, never mind, I read the code again, it's clear

packages/ethereum-contracts/contracts/gov/SuperfluidGovernanceBase.sol
```diff
@@ -63,12 +63,21 @@ abstract contract SuperfluidGovernanceBase is ISuperfluidGovernance
     {
         if (hostNewLogic != address(0)) {
             UUPSProxiable(address(host)).updateCode(hostNewLogic);
+            UUPSProxiable(address(hostNewLogic)).castrate();
```
hellwolf, line 66: I paused to think if we could brick the host upgrade by embedding the `castrate()` step. I don't think I can find a case this can be a problem.

If @d10r you think it's not a problem neither, I think we are good to go.

d10r: no, I can't think of any such case

hellwolf (approved):

PR #842, opened 2022-05-23
Pre-Release | Merge Dev Subgraph into `release-subgraph-v1`

0xdavinchee:


hellwolf:
wait, why don't we merge from "dev"?

0xdavinchee:
> wait, why don't we merge from "dev"?

the new process requires a PR (previous was a merge from dev, but a direct push)

hellwolf:
> > wait, why don't we merge from "dev"?
> 
> the new process requires a PR (previous was a merge from dev, but a direct push)

yes, but it can/should be from "dev" directly.

image

PR #843, opened 2022-05-23
CI Pre Release Draft Fix + CI Canary Cleanup

0xdavinchee:
* remove ".github/workflows/call.test-subgraph-on-previous-sdk-core-versions.yml" from paths
* generate dev schema-graphql + include SUBGRAPH_RELEASE_TAG env variable
* use `dev` as `subgraph-release` in pre-release ci
* only run `build-and-test-tbd-subgraph-with-previous-release` for v1 subgraph endpoint deploy

### DIFF
--- a/.github/workflows/ci.pre-release-draft-test-sdk-core-subgraph-integration.yml
+++ b/.github/workflows/ci.pre-release-draft-test-sdk-core-subgraph-integration.yml
@@ -36,7 +36,9 @@ jobs:
     name: Build and Test Subgraph (Release Branch)
     with:
       local-subgraph-url: http://localhost:8000/subgraphs/name/superfluid-test
-      subgraph-release: v1
+      # we must use dev endpoint here because v1 endpoints won't be deployed yet
+      # we want to test what is to be deployed
+      subgraph-release: dev
 
   # ensure subgraph indexing is complete on all v1 endpoints before allowing sdk-core release
   check-subgraph-indexing-completeness:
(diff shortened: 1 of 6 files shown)


.github/workflows/ci.pre-release-draft-test-sdk-core-subgraph-integration.yml
```diff
@@ -60,4 +60,7 @@ jobs:
     name: Build and test current subgraph release with previous sdk-core versions
```
hellwolf: since you are at it, what could be a less hideous file name? Maybe `ci.pre-release-sdk-core` says it all? And it could be a general pattern going forward for any `pre-release-${package}` testing?

hellwolf (approved):
some feedback

.github/workflows/ci.pre-release-draft-test-sdk-core-subgraph-integration.yml
0xdavinchee: yeah, it's kinda for both subgraph/sdk-core though, but I can split it into two files and that's a good pattern for naming convention

hellwolf: okay, yes. It can be more clear. e.g. `  check-subgraph-indexing-completeness:` seems only relevant for the sdk release?

0xdavinchee: yeah I agree, there's some jobs that only are run depending on if it's for sdk-core/subgraph release

PR #854, opened 2022-05-24
set `newCtx` for callAppAfterCallback in IDA

0xdavinchee:


hellwolf (approved):
good pattern to stick to without worrying about future-proof-ness

PR #855, opened 2022-05-25
Missing transfer events for liquidation

### LINKED ISSUE #445: [ETHEREUM-CONTRACTS] Add missing events in the protocol
As a protocol user, there are some events in the protocol missing that cause some use cases harder to implement, having a list of them and add them back in the next upgrades and available from day one for future new network deployments would be very helpful.

## List

- [x] ~~ERC20 Transfer events for CFA createFlow~~ it can only be possible when we have token-centric interface.
- [x] Claimed event for IDA.claim
  - This causes subgraph entity for subscribers not having the most up-to-date data snapshots (while technically still correct).
- [ ] #855

0xdavinchee:
### Contracts
* EventsEmitter library created to emit unaccessible ERC20 Transfer events in `SuperfluidToken.sol`'s `makeLiquidationPaymentsV2` function
* Test behavior logic changed: expect transfer events when liquidations occur
* `makeLiquidationPayoutsV2` logic refactored (same logical behavior, but easier to reason about with new branch logic check - checking the sign of targetAccountBalanceDelta instead of checking useDefaultRewardAccount)
  * add asserts
  * use nicer assignment operator syntax here
  * change `rewardAccount` => `rewardAmountReceiver`

### Subgraph
* [SUBGRAPH FIX] - tokenStats.totalAmountStreamedUntilUpdated updater function was not updating updatedAtTimestamp, updatedAtBlockNumber
* [SUBGRAPH FIX] - totalAmountStreamedUntilUpdated was missing updates in some places 
* Test behavior logic changed
   * liquidation transferred amount tracking for ATS
   * remove old claim specific logic
   * handle no event emit case for distributions
* change `rewardAccount` => `rewardAmountReceiver`

### SDK-Core/SDK-Redux
* change `rewardAccount` => `rewardAmountReceiver`

### DIFF
--- a/packages/ethereum-contracts/contracts/superfluid/SuperfluidToken.sol
+++ b/packages/ethereum-contracts/contracts/superfluid/SuperfluidToken.sol
@@ -336,6 +336,7 @@ abstract contract SuperfluidToken is ISuperfluidToken
         if (useDefaultRewardAccount) {
             _balances[rewardAccount] = _balances[rewardAccount]
                 + rewardAmount.toInt256();
+            _assemblyEmitTransfer();
         } else {
             _balances[liquidatorAccount] = _balances[liquidatorAccount]
                 + rewardAmount.toInt256();
@@ -363,6 +364,37 @@ abstract contract SuperfluidToken is ISuperfluidToken
             liquidationTypeData
         );
     }
+
+    function _assemblyEmitTransfer() internal {
+            assembly {
+                function decodeAsAddress(offset) -> v {
+                    v := decodeAsUint(offset)
+                    if iszero(iszero(and(v, not(0xffffffffffffffffffffffffffffffffffffffff)))) {
+                        revert(0, 0)
+                    }
+                }
+                function decodeAsUint(offset) -> v {
+                    let pos := add(4, mul(offset, 0x20))
+                    if lt(calldatasize(), add(pos, 0x20)) {
+                        revert(0, 0)
+                    }
+                    v := calldataload(pos)
+                }
+                function emitTransfer(from, to, amount) {
+                    // keccak256(Transfer(address,address,uint256))
+                    let signatureHash := 0xddf252ad1be2c89b69c2b068fc378daa952ba7f163c4a11628f55a4df523b3ef
+                    emitEvent(signatureHash, from, to, amount)
+                }
+
+                function emitEvent(signatureHash, indexed1, indexed2, nonIndexed) {
+                    mstore(0, nonIndexed)
+                    log3(0, 0x20, signatureHash, indexed1, indexed2)
+                }
+
+                emitTransfer(decodeAsAddress(0), decodeAsAddress(1), decodeAsUint(0))
+            }
+    }
+
     /**************************************************************************
     * Modifiers
     *************************************************************************/
--- a/packages/ethereum-contracts/contracts/superfluid/SuperfluidToken.sol
+++ b/packages/ethereum-contracts/contracts/superfluid/SuperfluidToken.sol
@@ -346,7 +346,7 @@ abstract contract SuperfluidToken is ISuperfluidToken
             // NOTE: useDefaultRewardAccount being true is undefined behavior
             // because the default reward account isn't receiving the rewardAmount by default
             assert(!useDefaultRewardAccount);
-            _balances[rewardAccount] -= rewardAmount.toInt256() - targetAccountBalanceDelta;
+            _balances[rewardAccount] -= (rewardAmount.toInt256() + targetAccountBalanceDelta);
             _balances[liquidatorAccount] += rewardAmount.toInt256();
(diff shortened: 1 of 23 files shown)


packages/ethereum-contracts/contracts/superfluid/SuperfluidToken.sol
```diff
@@ -308,61 +309,61 @@ abstract contract SuperfluidToken is ISuperfluidToken
         _balances[account] = _balances[account] + delta;
     }
 
+    function _getRewardAccount() internal view returns (address rewardAccount) {
+        ISuperfluidGovernance gov = _host.getGovernance();
+        rewardAccount = gov.getConfigAsAddress(_host, this, _REWARD_ADDRESS_CONFIG_KEY);
+    }
+
     /// @dev ISuperfluidToken.makeLiquidationPayoutsV2 implementation
     function makeLiquidationPayoutsV2(
```
hellwolf: @d10r you had comments regarding this function before, now we refactored it, please review it again.

packages/ethereum-contracts/contracts/superfluid/SuperfluidToken.sol
```diff
@@ -308,61 +309,61 @@ abstract contract SuperfluidToken is ISuperfluidToken
...
-        if (useDefaultRewardAccount) {
-            _balances[rewardAccount] = _balances[rewardAccount]
-                + rewardAmount.toInt256();
-        } else {
-            _balances[liquidatorAccount] = _balances[liquidatorAccount]
-                + rewardAmount.toInt256();
+        if (targetAccountBalanceDelta <= 0) {
+            // LIKELY BRANCH: target account pays penalty to rewarded account
+            assert(rewardAmount.toInt256() == -targetAccountBalanceDelta);
+            address rewardedAccount = useDefaultRewardAccount ? rewardAccount : liquidatorAccount;
```
d10r: This will reset the rewardAccount to the zero address in anarchy mode.
Also I suggest we don't shadow variables, but pick a different name. E.g. here it could be `actualRewardAccount`.

packages/ethereum-contracts/contracts/superfluid/SuperfluidToken.sol
```diff
@@ -308,61 +309,61 @@ abstract contract SuperfluidToken is ISuperfluidToken
...
+            _balances[targetAccount] = _balances[targetAccount] + targetAccountBalanceDelta;
+            EventsEmitter.emitTransfer(targetAccount, rewardedAccount, rewardAmount);
+        } else {
+            // LESS LIKELY BRANCH: target account is bailed out
+            // NOTE: useDefaultRewardAccount is ignored
             _balances[rewardAccount] = _balances[rewardAccount]
                 - rewardAmount.toInt256()
                 - targetAccountBalanceDelta;
+            _balances[liquidatorAccount] = _balances[liquidatorAccount] + rewardAmount.toInt256();
+            _balances[targetAccount] = _balances[targetAccount] + targetAccountBalanceDelta;
```
d10r: operators `+=` and `-=` (should now exists according to the [docs](https://docs.soliditylang.org/en/v0.8.13/cheatsheet.html#order-of-precedence-of-operators)) would make this more readable

d10r (changes requested):

packages/ethereum-contracts/contracts/superfluid/SuperfluidToken.sol
0xdavinchee: haven't pushed some changes yet, but `rewardedAccount` => `rewardAmountReceiver`

Also, we should be safe here with regards to the rewardAccount not being the zero address, it is set above to liquidator account so rewardedAccount will just be liquidator account in that case.

packages/ethereum-contracts/contracts/superfluid/SuperfluidToken.sol
0xdavinchee: agree, will make this change before pushing changes

packages/ethereum-contracts/contracts/superfluid/SuperfluidToken.sol
d10r: you're right

hellwolf:
@0xdavinchee sdk-core test broken.

@d10r otherwise, is it alright now?

0xdavinchee:
> @0xdavinchee sdk-core test broken.
> 
> @d10r otherwise, is it alright now?

SDK-Core tests are gonna be broken until feature matic endpoint is indexed - I will switch it so that feature endpoint waits on goerli instead of matic. It might make sense to do the same with v1/dev as waiting for matic indexing is a pretty huge blocker. 

We have to wait for all networks to be indexed anyways before releasing to v1.

d10r (approved):

hellwolf (approved):
NICE

PR #858, opened 2022-05-26
fixed rewards dist token

saflamini:
@hellwolf this PR contains a fixed version of the rewards distribution contract example

1) Several aspects of the test suite needed to be redone
2) The build script and associated package.json files have been patched.

Please let me know if you have any other questions which would require you to approve.

Thank you,

-Sam


0xdavinchee:
hey @saflamini made a few small changes:
- deleted and gitignoring `typechain-types`: created on `yarn build`
- removed unused code in `hardhat.config.ts`
- cleaned up `DividendRightsToken.test.ts` a bit

d10r (changes requested):
Having an untainted view on this, I tried to follow the readme. Some stumbling blocks:
* doesn't mention to do a `yarn install` first
* `npx hardhat accounts` fails (_Unrecognized task_). Not sure why. If not available anymore, should be removed from readme.
* `npx hardhat run scripts/deploy.ts` and `TS_NODE_FILES=true npx ts-node scripts/deploy.ts` fail with `Artifact for contract "DividendRightsToken__factory" not found.`
* `npx eslint '**/*.{js,ts}'` fails with `ESLint couldn't find the config "standard" to extend from.`
* `npx prettier '**/*.{json,sol,md}' --check` fails with `contracts/DividendRightsToken.sol[error] No parser could be inferred for file: contracts/DividendRightsToken.sol`

Other observations:
* `npx hardhat verify --network goerli DEPLOYED_CONTRACT_ADDRESS "Hello, Hardhat!"` haven't tried this one (deploy failed), but doubt if it works. As far as I understand, the last argument is an optional constructor parameter which seems not to fit here.

My changes:
I pushed a commit with the following changes:
* replaced Ropsten with Görli bcs Ropsten may be deprecated soonish
* Uncommented the user config in `hardhat.config.ts`. If we don't want that ucommented, the README should mention that the hardhat config needs to be edited for some of the commands to work
* modified `.env.example` with inline comments, removed `API_KEY` which seems to be unused / not clear what it is

d10r:
Also, it may be nice to include instructions for forking a mainnet deployment into a local dev env. But don't know if this is the right place for that.

saflamini:
> Also, it may be nice to include instructions for forking a mainnet deployment into a local dev env. But don't know if this is the right place for that.

@d10r - I have fixed the deploy workflow, and added more clear steps to the README. 
I've also made the RPC url template more generalized by adding an option for just the RPC_URL in the .env.template

@d10r is there anything else you'd add before we merge this?

d10r:
> @d10r is there anything else you'd add before we merge this?

I spent some more time on it and improved the deploy script to not require manual adaptations on network change and to output the cmd needed for etherscan verification.

The commands `npx eslint...` and `npx prettier...` mentioned in the README still don't work (only for me?). Imo we should either remove them from the list or make sure they work.

Something I just noticed:  
There's an inconsistency in the deploy script: it will take the RPC url directly from env while relying on the signer set by hardhat.  
That can e.g. result in a situation where it will running without `--network` parameter will connect to the RPC configured in .env, but ignore the PRIVATE_KEY set there unless the default network in hardhat config happens to use it.  
When no network is specified in hardhat config, it defaults to the "hardhat" network (similar to ganache in truffle) with a signer account with address 0xf39Fd6e51aad88F6F4ce6aB8827279cffFb92266.  
I'm not familiar enough with hardhat to know what the most idiomatic way is.  
It's also not something I think needs to further delay this PR, just wanted to mention in this context.

d10r (approved):

PR #866, opened 2022-05-30
SDK-Core Subgraph Tests

0xdavinchee:
* Use MATIC for v1/release
* Use GOERLI for feature and dev

### DIFF
--- a/packages/sdk-core/previous-versions-testing/queryTests.ts
+++ b/packages/sdk-core/previous-versions-testing/queryTests.ts
@@ -1,5 +1,20 @@
 import { Query } from "@superfluid-finance/sdk-core/src";
 import { expect } from "chai";
+import {
+    ETH_GOERLI_CHAIN_ID,
+    MATIC_CHAIN_ID,
+} from "@superfluid-finance/sdk-core/src/constants";
+
+/**
+ * We only use matic network endpoints for v1 release tests
+ * otherwise, we use goerli
+ * @returns chainId
+ */
+export const getChainId = () => {
+    return process.env.SUBGRAPH_RELEASE_TAG == "v1"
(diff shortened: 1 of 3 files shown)


packages/sdk-core/previous-versions-testing/runQueryTests.ts
```diff
@@ -1,9 +1,7 @@
 import { Query } from "@superfluid-finance/sdk-core/src";
+import { chainIdToResolverDataMap } from "@superfluid-finance/sdk-core/src/constants";
```
kasparkallas, line 2: I don't see this import being used.

packages/sdk-core/previous-versions-testing/queryTests.ts
```diff
@@ -1,5 +1,20 @@
 import { Query } from "@superfluid-finance/sdk-core/src";
 import { expect } from "chai";
+import {
+    ETH_GOERLI_CHAIN_ID,
+    MATIC_CHAIN_ID,
```
kasparkallas, line 5: These imports look ugly cosmetically. :smile: Would have preferred an enum or something.

kasparkallas (approved):

packages/sdk-core/previous-versions-testing/runQueryTests.ts
0xdavinchee: it's used on line 13 by `getSubgraphEndpoint`

packages/sdk-core/previous-versions-testing/queryTests.ts
0xdavinchee: true, can be something to refactor for SDK-core in the future, but there were some strange type issues w/ enums in the past that were kinda annoying

PR #870, opened 2022-05-31
SDK-Core 0.4.4

0xdavinchee:


kasparkallas (approved):
Didn't I already approve this?

hellwolf:
test fails, by design?

0xdavinchee:
Yes, it should be failing because V1 isn't indexed yet on matic

hellwolf:
but under which condition it fails:

can you make it more clear, with current compatibility matrix e.g.:

| subgraph endpoint \ sdk-core | < 0.4.3 | 0.4.3 |
|--|--|--|
| v1 | ? | ? |
| dev | ? | ? |
| feature | ? | ? |

👍 - always works
⛔ - always fail
ℹ️ - conditional, in certain case it's broken.

hellwolf:
okay, so https://github.com/superfluid-finance/protocol-monorepo/blob/2a678e07530f5d331344fdc8cb48e5c1d05e53f2/.github/workflows/ci.pre-release-sdk-core.yml needs to be re-rewritten to originally how it was planned. otherwise it doesn't give any useful decision making points here, what we want is:

| subgraph endpoint \ sdk-core | < NEW-VERSION | NEW-VERSION |
|--|--|--|
| v1 (to be synced with dev) | 👍 or   ℹ️ |  👍 or  ℹ️ |
| dev | 👍 or   ℹ️ | 👍 (b) |

👍 - always works
⛔ - always fail **AND THIS IS A NG for the release**
ℹ️ - conditional, in certain breaking-change case it's broken (impact analysis needed, and to be documented in README/CHANGELOG).

(a) If the tests cannot tell this information, then they are not useful for release gate-keeping.
(b) **This is important, since it tells that new-version is fully working with dev endpoint**.

hellwolf (dismissed):
make test result useful for decision making.

hellwolf:
Make this a template for future sdk pre-release report.

| subgraph endpoint \ sdk-core | < NEW-VERSION | NEW-VERSION |
|--|--|--|
| dev | 👍 or   ℹ️ | 👍 (b) |

👍 - always works
⛔ - always fail **AND THIS IS A NG for the release**
ℹ️ - conditional, in certain breaking-change case it's broken (impact analysis needed, and to be documented in README/CHANGELOG).

(b) This is important, since it tells that new-version is fully working with dev endpoint.

0xdavinchee:
@hellwolf I made some changes and added comments to the workflows here: https://github.com/superfluid-finance/protocol-monorepo/pull/872

I think it will be good to do a 1.4.1 patch for the subgraph, that is, revert the breaking name change by keeping the previous property and adding a new one (the renamed one we have on ethereum-contracts)

PR #872, opened 2022-06-01
Workflow Edits + Subgraph 1.4.1 Patch

0xdavinchee:
* workflow comments added to jobs in pre-release CI files
* `generate-graphql-schema:local` script to test local subgraph endpoints without dependency on deployed endpoint
* Subgraph v1.4.1 patch keep rewardAccount in subgraph to not break previous + current versions of sdk-core
   * essentially reverting breaking change to previous SDK-core versions

> NOTE: after some discussions with @kasparkallas this morning, it makes sense not to break previous SDK-core versions as a result of subgraph updates. It does seem to make the most sense to keep these kinds of breaking changes (changing the name of a property) for a brand new v2 subgraph endpoint which can be released/supported with a new SDK-core version and then deprecating the v1 endpoint completely.

### DIFF
--- a/.github/workflows/call.setup-deploy-and-test-local-subgraph.yml
+++ b/.github/workflows/call.setup-deploy-and-test-local-subgraph.yml
@@ -53,7 +53,7 @@ jobs:
       - name: "Docker compose"
         run: docker-compose up &
         working-directory: ./graph-node/docker
-      #
+      #                                                                     #
 
       - name: "Run subgraph integration test suite"
         if: inputs.local-subgraph-url == ''
@@ -73,5 +73,5 @@ jobs:
           yarn generate
           npx hardhat test --network localhost
         working-directory: ./packages/sdk-core
-        env: 
+        env:
           LOCAL_SUBGRAPH_URL: ${{ inputs.local-subgraph-url }}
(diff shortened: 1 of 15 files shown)


hellwolf:
I find it incredible that there is no way of introduce a localized breaking change without breaking previously released SDKs completely. It is against common sense, and there must be something wrong the way we do it, we can connect to thegraph team again to address this. This is important for us to figure out, we don't want to be shackled because of this not being able to address technical debt.

hellwolf (changes requested):
need to review the backward compatibility strategy again

0xdavinchee:
@hellwolf @kasparkallas following our discussion this morning - we will move forward by patching the current subgraph release and any future modifications to the subgraph in this less-than-ideal manner so that the subgraph endpoints are fully backwards compatible with previous sdk-core versions until we have a suitable solution: https://github.com/superfluid-finance/protocol-monorepo/issues/875 

I've added the deprecated tag to the schema.graphql file in the subgraph package-mostly as internal record keeping though.

0xdavinchee:
@hellwolf is there anything else left, if not, can we please unblock this?

hellwolf (approved):
:sad panda:

PR #882, opened 2022-06-03
[SDK-CORE] Map flow update type to FlowUpdatedEvent

kasparkallas:
Why? We need it for better mapping of activities to UI.


0xdavinchee (approved):
looks good - LGTM

PR #885, opened 2022-06-06
calculateMaybeCriticalAtTimestamp is calculating for "insolvency time" instead of "critical time"

msoni89:
https://github.com/superfluid-finance/user-dashboard/issues/1201

### DIFF
--- a/packages/subgraph/src/utils.ts
+++ b/packages/subgraph/src/utils.ts
@@ -252,9 +252,10 @@ export function calculateMaybeCriticalAtTimestamp(
     updatedAtTimestamp: BigInt,
     balanceUntilUpdatedAt: BigInt,
     totalNetFlowRate: BigInt
-): BigInt {
-    if (balanceUntilUpdatedAt.le(BIG_INT_ZERO)) return BIG_INT_ZERO;
-    if (totalNetFlowRate.ge(BIG_INT_ZERO)) return BIG_INT_ZERO;
+): BigInt | null {
+    if (totalNetFlowRate.ge(BIG_INT_ZERO)) return null;
+    if (balanceUntilUpdatedAt.equals(BIG_INT_ZERO)) return BIG_INT_ZERO;
+    if (balanceUntilUpdatedAt.lt(BIG_INT_ZERO)) return updatedAtTimestamp;
     const secondsUntilCritical = balanceUntilUpdatedAt.div(totalNetFlowRate.abs());
     const calculatedCriticalTimestamp = secondsUntilCritical.plus(updatedAtTimestamp);
     if (calculatedCriticalTimestamp.gt(MAX_SAFE_SECONDS)) {
--- a/packages/subgraph/src/utils.ts
+++ b/packages/subgraph/src/utils.ts
@@ -242,22 +242,20 @@ export function getAmountStreamedSinceLastUpdatedAt(
 }
 
 /**
- * calculateMaybeCriticalAtTimestamp will return optimistic date based on updatedAtTimestamp, totalDeposit, balanceUntilUpdatedAt and totalNetFlowRate.
+ * calculateMaybeCriticalAtTimestamp will return optimistic date based on updatedAtTimestamp, balanceUntilUpdatedAt and totalNetFlowRate.
  * @param updatedAtTimestamp
  * @param balanceUntilUpdatedAt
- * @param totalDeposit
  * @param totalNetFlowRate
  */
 
 export function calculateMaybeCriticalAtTimestamp(
     updatedAtTimestamp: BigInt,
     balanceUntilUpdatedAt: BigInt,
-    totalDeposit: BigInt,
     totalNetFlowRate: BigInt
 ): BigInt {
     if (balanceUntilUpdatedAt.le(BIG_INT_ZERO)) return BIG_INT_ZERO;
     if (totalNetFlowRate.ge(BIG_INT_ZERO)) return BIG_INT_ZERO;
-    const criticalTimestamp = balanceUntilUpdatedAt.plus(totalDeposit).div(totalNetFlowRate.abs());
+    const criticalTimestamp = balanceUntilUpdatedAt.div(totalNetFlowRate.abs());
     const calculatedCriticalTimestamp = criticalTimestamp.plus(updatedAtTimestamp);
     if (calculatedCriticalTimestamp.gt(MAX_SAFE_SECONDS)) {
         return MAX_SAFE_SECONDS;
--- a/packages/subgraph/src/utils.ts
+++ b/packages/subgraph/src/utils.ts
@@ -246,16 +246,18 @@ export function getAmountStreamedSinceLastUpdatedAt(
  * @param updatedAtTimestamp
  * @param balanceUntilUpdatedAt
  * @param totalNetFlowRate
+ * @param previousMaybeCriticalAtTimestamp
  */
 
 export function calculateMaybeCriticalAtTimestamp(
     updatedAtTimestamp: BigInt,
     balanceUntilUpdatedAt: BigInt,
-    totalNetFlowRate: BigInt
+    totalNetFlowRate: BigInt,
+    previousMaybeCriticalAtTimestamp: BigInt | null,
 ): BigInt | null {
     if (totalNetFlowRate.ge(BIG_INT_ZERO)) return null;
     if (balanceUntilUpdatedAt.equals(BIG_INT_ZERO)) return BIG_INT_ZERO;
-    if (balanceUntilUpdatedAt.lt(BIG_INT_ZERO)) return updatedAtTimestamp;
+    if (balanceUntilUpdatedAt.lt(BIG_INT_ZERO)) return previousMaybeCriticalAtTimestamp;
     const secondsUntilCritical = balanceUntilUpdatedAt.div(totalNetFlowRate.abs());
     const calculatedCriticalTimestamp = secondsUntilCritical.plus(updatedAtTimestamp);
     if (calculatedCriticalTimestamp.gt(MAX_SAFE_SECONDS)) {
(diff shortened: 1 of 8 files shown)


msoni89:
@kasparkallas 
Please review

packages/subgraph/src/utils.ts
```diff
@@ -242,22 +242,20 @@ export function getAmountStreamedSinceLastUpdatedAt(
...
- * @param totalDeposit
  * @param totalNetFlowRate
  */
 
 export function calculateMaybeCriticalAtTimestamp(
     updatedAtTimestamp: BigInt,
     balanceUntilUpdatedAt: BigInt,
-    totalDeposit: BigInt,
     totalNetFlowRate: BigInt
 ): BigInt {
```
kasparkallas: There is a small debate whether to use number 0 or NULL to represent no timestamp.

packages/subgraph/src/utils.ts
```diff
@@ -242,22 +242,20 @@ export function getAmountStreamedSinceLastUpdatedAt(
...
     if (balanceUntilUpdatedAt.le(BIG_INT_ZERO)) return BIG_INT_ZERO;
```
kasparkallas: Not sure I like this because when an account becomes critical then many streams can be liquidated which means that after the first liquidation this currently becomes zero when it's still critical and there are more streams to be liquidated.

I'd rather return the `updatedAtTimestamp` and only nullify this when the flow rate is zero or positive.

EDIT: Actually, I might use "0" to point to entities _already_ critical and use NULL to point to entities without a critical timestamp. Because we already used "0" for something else then I'd put "1" or something.

Thoughts? @0xdavinchee @msoni89

packages/subgraph/src/utils.ts
```diff
@@ -242,22 +242,20 @@ export function getAmountStreamedSinceLastUpdatedAt(
...
     if (totalNetFlowRate.ge(BIG_INT_ZERO)) return BIG_INT_ZERO;
-    const criticalTimestamp = balanceUntilUpdatedAt.plus(totalDeposit).div(totalNetFlowRate.abs());
```
kasparkallas, line 260: I think `secondsUntilCritical` would be a better name.

hellwolf:
@0xdavinchee please check how do you want to mange the release of this, be a hotfix or fit to the next release (which is still in the pr).

packages/subgraph/test/validation/aggregateValidators.ts
```diff
@@ -53,7 +53,7 @@ export const validateATSEntity = (
         totalDeposit: expectedTotalDeposit,
     } = expectedATSData;
 
-    const calculatedCriticalTimestamp = calculateMaybeCriticalAtTimestamp(expectedATSData.updatedAtTimestamp, expectedBalanceUntilUpdatedAt, expectedTotalDeposit, expectedTotalNetFlowRate);
+    const calculatedCriticalTimestamp = calculateMaybeCriticalAtTimestamp(expectedATSData.updatedAtTimestamp, expectedBalanceUntilUpdatedAt, expectedTotalNetFlowRate);
```
kasparkallas, line 56: Just saying, it's slippery slope to effectively use _the implementation_ to assert this... but I'm not familiar with the tests and there might not be a much better way to do this.

packages/subgraph/src/utils.ts
kasparkallas: So ideally the timestamp would convey 3 states:
1. Already critical
2. Will be critical
3. Not possible to be critical

Right now the solution does not convey 1. very well.

0xdavinchee:
> @0xdavinchee please check how do you want to mange the release of this, be a hotfix or fit to the next release (which is still in the pr).

I think we should fit this with the next release instead of another patch. Will need to deploy on feature/dev and wait for indexing again

packages/subgraph/src/utils.ts
0xdavinchee: @kasparkallas

I think it makes sense to do this:
1. Already critical, criticalAtTimestamp = 0 or `updatedAtTimestamp` (but this may not illustrate that the account was critical at a prior time)
2. Will be critical, criticalAtTimestamp = calculatedCriticalTimestamp
3. Not possible to be critical = null

I think 1. should be 0 or `updatedAtTimestamp`, maybe it's better if it's 0 because we can't guarantee that the liquidations are always happening on the same block that the account is critical and so we may have some incorrect information there

packages/subgraph/src/utils.ts
msoni89: DONE

packages/subgraph/test/validation/aggregateValidators.ts
0xdavinchee: This is the pattern currently present in both the ethereum contracts/subgraph.
We have an initial local state, then we actually modify it (send txn), then we update our local state based on expected behavior logic and compare the local state to the actual state after the modification.
We have expected behavior logic which we use to calculate the information again, this relies on the assumption that our expected behavior logic is correct.

packages/subgraph/src/utils.ts
kasparkallas: Another option would be to keep the initial `maybeCriticalAtTimestamp` for 1.

msoni89: handled 3rd returning null when `Not possible to be critical`

0xdavinchee (approved):
looks good - LGTM

packages/subgraph/src/utils.ts
```diff
@@ -242,23 +242,24 @@ export function getAmountStreamedSinceLastUpdatedAt(
...
-): BigInt {
-    if (balanceUntilUpdatedAt.le(BIG_INT_ZERO)) return BIG_INT_ZERO;
-    if (totalNetFlowRate.ge(BIG_INT_ZERO)) return BIG_INT_ZERO;
-    const criticalTimestamp = balanceUntilUpdatedAt.plus(totalDeposit).div(totalNetFlowRate.abs());
-    const calculatedCriticalTimestamp = criticalTimestamp.plus(updatedAtTimestamp);
+    totalNetFlowRate: BigInt,
+    previousMaybeCriticalAtTimestamp: BigInt | null,
+): BigInt | null {
+    if (totalNetFlowRate.ge(BIG_INT_ZERO)) return null;
+    if (balanceUntilUpdatedAt.equals(BIG_INT_ZERO)) return BIG_INT_ZERO;
```
kasparkallas: I don't understand this one: `if (balanceUntilUpdatedAt.equals(BIG_INT_ZERO)) return BIG_INT_ZERO;` because when balance is zero then the liquidation date shouldn't be 50 years in the past (i.e. "0").

This is the code I envision:
```ts
export function calculateMaybeCriticalAtTimestamp(
    updatedAtTimestamp: BigInt,
    balanceUntilUpdatedAt: BigInt,
    totalNetFlowRate: BigInt,
    previousMaybeCriticalAtTimestamp: BigInt | null,
): BigInt | null {
    // When the flow rate is not negative then there's no way to have a critical balance timestamp anymore.
    if (totalNetFlowRate.ge(BIG_INT_ZERO)) return null;

    // When there's no balance then that either means:
    // 1. account is already critical and we keep the existing timestamp when the liquidations supposedly started
    // 2. it's a new account without a critical balance timestamp to begin with
    if (balanceUntilUpdatedAt.lte(BIG_INT_ZERO)) return previousMaybeCriticalAtTimestamp;

    const secondsUntilCritical = balanceUntilUpdatedAt.div(totalNetFlowRate.abs());
    const calculatedCriticalTimestamp = updatedAtTimestamp.plus(secondsUntilCritical);
    if (calculatedCriticalTimestamp.gt(MAX_SAFE_SECONDS)) {
        return MAX_SAFE_SECONDS;
    }
    return calculatedCriticalTimestamp;
}
```

msoni89: DONE

packages/subgraph/test/helpers/helpers.ts
```diff
@@ -468,13 +468,13 @@ export const getOrder = (blockNumber?: number, logIndex?: number) => {
...
-    totalNetFlowRate: string
+    totalNetFlowRate: string,
 ) {
-    if (BigNumber.from(balanceUntilUpdatedAt).lte(BigNumber.from("0"))) return "0";
-    if (BigNumber.from(totalNetFlowRate).gte(BigNumber.from("0"))) return "0";
-    const criticalTimestamp = BigNumber.from(balanceUntilUpdatedAt).add(BigNumber.from(totalDeposit)).div(BigNumber.from(totalNetFlowRate).abs());
-    const calculatedCriticalTimestamp = criticalTimestamp.add(BigNumber.from(updatedAtTimestamp));
+    if (toBN(totalNetFlowRate).gte(toBN("0"))) return null;
+    if (toBN(balanceUntilUpdatedAt).eq(toBN("0"))) return "0";
+    if (toBN(balanceUntilUpdatedAt).lt(toBN("0"))) return updatedAtTimestamp;
```
kasparkallas: This implementation is not the same as the other one now.

kasparkallas (commented):
@0xdavinchee When are the changes supposed to hit `sdk-core`. I'm guessing in this same PR?

packages/subgraph/src/utils.ts
0xdavinchee: initialized maybeCriticalAtTimestamp = null

packages/subgraph/test/helpers/helpers.ts
0xdavinchee: The only change in sdk-core would be to set the possibility that you will be returned null. Shouldn't be breaking to do this as the queries don't change

msoni89: Done

kasparkallas (approved):

PR #890, opened 2022-06-13
Simple IDA Example

joeljesu:
Creating the new IDA example mentioned in superfluid-finance/super-examples#10


saflamini:
@sunnyjaycer 

Contract looks good for the most part. You could consider implementing some form of access control as seen [here](https://github.com/superfluid-finance/protocol-monorepo/blob/money_router_example/examples/money-router/contracts/MoneyRouter.sol#L24), but it's not a requirement

A couple of small changes I'd recommend though in your deploy script:
1) Make sure that `gorli` is spelled `goerli`. This might trip some people up
2) In your deploy script, I'd recommend initializing the framework and getting the host and daix address from the Framework object instead of hard coding them

joeljesu:
Sounds good! I'll make those changes. On access control, just cutting it out to make this as bare bones as possible.

hellwolf:
nftbillboard fails now

saflamini (approved):

saflamini (approved):

saflamini:
> nftbillboard fails now

@hellwolf - made change that fixed the broken deployment when the bash script runs

@sunnyjaycer - this looks good.

PR #901, opened 2022-06-16
Make SDK-core test cases independant from each other

### LINKED ISSUE #758: [SDK-CORE] Tests Should Work Independently
Tests in the SDK-Core currently rely on the previous test and cannot be run independently from one another, we should ensure that this is not the case and each test case is independent from others.

elvijsTDL:


0xdavinchee (approved):
Looks good - LGTM

PR #910, opened 2022-06-21
Remove `networkId` and `dataMode` from Framework Initialization

### LINKED ISSUE #838: [SDK-Core] Framework initialization Cleanup
## Problems
- **always requires a provider (SUBGRAPH_ONLY mode too)**: As an SDK core user, I would like to use SUBGRAPH_ONLY mode without having to pass in a provider, otherwise SUBGRAPH_ONLY mode seems rather redundant. Given the way that framework creation currently works, it may also make more sense to simply remove SUBGRAPH_ONLY mode option and to ask users to use the query class directly instead of through framework initialization. 
- Remove `networkName` field from SDK-core initialization, Devx has relayed feedback from developers that this is redundant and confusing since the `chainId` property already exists

## Todos

- [x] Remove `dataMode` option from framework #910
- [x] Remove `networkName` property from framework initialization  #910

0xdavinchee:
* lots of code/tests removed for handling these two options

### DIFF
--- a/packages/sdk-core/CHANGELOG.md
+++ b/packages/sdk-core/CHANGELOG.md
@@ -5,6 +5,10 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 
 ## [Unreleased]
 
+### Breaking
+- `chainId` is a required property for framework initialization
+- `networkId` and `dataMode` no longer exist as properties for framework initialization
+
 ## [0.4.3] - 2022-05-31
 
 ### Breaking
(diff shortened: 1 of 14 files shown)


0xdavinchee:
**NOTE: Once this is merged to dev it will be added: #870 and the breaking changes will be released in version 0.4.3**

hellwolf:
How many breakage do you expect there can be by requiring chainId? @0xdavinchee

0xdavinchee:
> How many breakage do you expect there can be by requiring chainId? @0xdavinchee

it's hard to say - it depends on how many people are using chainId vs. networkName for initialization, but the fix is very simple.

hellwolf (approved):
let's go

PR #913, opened 2022-06-22
Token spreader fixes

joeljesu:
Added tutorial to README file and added in demo scripts


d10r (changes requested):
Imo we should add a lockfile to the beginner examples, otherwise upstream changes of random packages may break builds of the monorepo at random times.
Didn't do it myself bcs I'm not sure if you want to use npm here (as used in the README) or yarn (as used in the build scripts)

joeljesu:
Oh true @d10r I didn't realize I omitted that

d10r (approved):

PR #930, opened 2022-06-30
Package.json updates

0xdavinchee:
- `package.json` package updates for monorepo, ethereum-contracts, hot-fuzz, js-sdk, sdk-core, sdk-redux, and subgraph
- did not update js-sdk packages: `auto-bind` and `node-fetch`
- did not update subgraph package: `@graphprotocol/graph-cli`


kasparkallas:
The `peerDependencies` make this a breaking change.

packages/sdk-core/package.json
```diff
@@ -57,21 +57,25 @@
     "dependencies": {
         "ajv": "^8.11.0",
         "browserify": "^17.0.0",
-        "ethers": "^5.6.2",
-        "graphql-request": "^3.7.0",
+        "ethers": "^5.6.9",
```
kasparkallas: When peerDependency then this is supposed to be removed. Can be moved into a devDependency though.

0xdavinchee: done

packages/ethereum-contracts/foundry.toml
```diff
@@ -4,6 +4,7 @@ libs = ['lib']
 src = 'packages/ethereum-contracts'
 remappings = [
     '@superfluid-finance/ethereum-contracts/contracts/=packages/ethereum-contracts/contracts/',
+    'hardhat=packages/ethereum-contracts/node_modules/hardhat/',
```
hellwolf: hmm, what shady business is happening here : )??

package.json
```diff
@@ -22,47 +22,47 @@
...
-        "eslint-plugin-prettier": "^3.4.1",
+        "eslint-plugin-prettier": "^4.2.1",
         "ethereum-waffle": "^3.4.4",
-        "hardhat": "^2.9.3",
-        "husky": "^5.2.0",
-        "lerna": "^3.22.1",
-        "node-jq": "^1.12.0",
-        "nodemon": "^2.0.15",
+        "hardhat": "^2.9.9",
+        "husky": "^8.0.1",
```
hellwolf, line 51: any breaking change re husky?

packages/ethereum-contracts/foundry.toml
0xdavinchee: good question :p 
it fixes this build: https://github.com/superfluid-finance/protocol-monorepo/runs/7127843899?check_suite_focus=true

the build broke because of the upgrade to hardhat v2.9.9: https://github.com/superfluid-finance/protocol-monorepo/commit/cbc062e6c9b5d2300378ef29b7402188edfa1ceb

they added example projects in the build and this broke the foundry tests

hellwolf: huh, why is it building `node_modules/hardhat/sample-projects/basic/contracts/Greeter.sol`

package.json
0xdavinchee: yes, the only breaking change is in v6: _husky init has been moved to its own package (npx husky-init)_

but it doesn't impact our precommit hooks.

https://github.com/typicode/husky/releases

packages/ethereum-contracts/foundry.toml
hellwolf: 😕

0xdavinchee: This is because we're calling forge test from `packages/ethereum-contracts`, I will add -`-match-path  test/foundry` and run an altered workflow to test it out

package.json
hellwolf: cool

packages/ethereum-contracts/package.json
```diff
@@ -37,7 +37,7 @@
         "pretest": "{ yarn testenv:start > /dev/null& } && sleep 5",
         "test": "run-s test:*:*",
         "test:contracts:hardhat": "yarn run-hardhat test testsuites/all-contracts.js --parallel",
-        "test:contracts:foundry": "yarn run-forge test",
+        "test:contracts:foundry": "yarn run-forge test --match-path test/foundry/**/*.sol",
```
hellwolf: you should do a single quote around so that shell doesn't expand it for you `--match-path 'test/foundry/**/*.sol'`

0xdavinchee: using the `--hh` flag worked quite well, if this is acceptable, this PR can be merged after the coverage tests are complete

hellwolf: ah well, not a fan of this kind of option, but so be it.

Though it doesn't pass the urbandictionary test, https://www.urbandictionary.com/define.php?term=HH, so let's use `--hardhat` instead.

hellwolf (approved):
Good.

But let's use `--hardhat` instead of `--hh`

packages/ethereum-contracts/package.json
0xdavinchee: damnit, almost sneaked this by :trollface:

PR #939, opened 2022-06-30
Add hardhat deployer

jtriley2p:


jtriley2p:
Hardhat network not working in GH actions. Will update.

Issue #860, opened 2022-05-26
[SDK-Core] Improved Error Handling


0xdavinchee:
Currently, for web3 calls we simply use a try/catch and throw a custom error type (SFError) which includes:
1. type: one of many error types we have hardcoded
2. message: a custom error string based on the type and a custom message written where the error is to be caught
3. error: an error object, which returns the caught error

The first and second properties are fine for the time being, but the third property has room for further improvement on how to process and and display this error object in a way that makes it easier for developers/end users to pinpoint what went wrong.

## TODO
- [x] Remove layered try/catch, only have it at the level where catching the error is actionable

mikeghen:
fwiw, here's how I get to the real error message:
<img width="1051" alt="Screen Shot 2022-05-28 at 4 43 30 PM" src="https://user-images.githubusercontent.com/5077880/170842256-d2b08523-6e38-4d7a-b42f-1491ccdc5af0.png">

kasparkallas:
> fwiw, here's how I get to the real error message:

Thanks!

Added couple of comments:
<img width="1035" alt="image" src="https://user-images.githubusercontent.com/10894666/170946074-e2ebc1c7-6485-4c95-86d8-a06a17110514.png">

mikeghen:
Awesome, this problem came up today while I was testing some changes to reduce my contract sizes, so I'm back to hacking this file to get the error message again, getting this error message: 
```
     Execute Transaction Error - There was an error executing the transaction: {}
```
Very informative 😅

0xdavinchee:
@mikeghen do you have a repo or a code snippet which can replicate this empty error object?
I find it quite odd that nothing is being returned even though an error is being caught.

hellwolf:
bunch of improvements had been done, closing it now.

PR #960, opened 2022-07-11
[SDK-core] Refactor SFError to follow more conventional Error object structure

kasparkallas:
Why?
Simply said, always disliked how we were wrapping a meaningful internal error into our own non-Error type object with a huge message because it contained the serialized version of the internal error. It looks horrible everywhere you display it (starting from the browser's developer console).

Let's just keep it simple and follow the conventions on how to implement a user-defined error and how to wrap an internal error.

Read about Error: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error 
Read about the new "cause" property: https://v8.dev/features/error-cause & https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error/cause


0xdavinchee:
I like this PR - I think it is a good step towards addressing: #860

0xdavinchee (approved):
Looks good - let's merge this once the subgraph is fully indexed on V1 so we can appropriately date the changelog release date for v0.4.5

hellwolf:
good to merge?

PR #979, opened 2022-07-19
fix Invalid workflow file error

0xdavinchee:


.github/workflows/handler.placeholder-for-testing.yml
```diff
@@ -9,7 +9,11 @@ on:
     inputs:
```
hellwolf, line 9: is this one still needed?

0xdavinchee: just kept it in because I assumed it is part of the placeholder file

hellwolf: ah, that's DIdi's stuff, okay

hellwolf (approved):

PR #982, opened 2022-07-19
CFA Agreement Forwarder

### LINKED ISSUE #919: [ETHEREUM-CONTRACTS] Agreement Forwarder Contract(s)
## Problem

When users use Superfluid, they need to use host.callAgreement to create/update/delete an agreement, however because of the current design, when users go to the block explorer they can decode the inputs to see 3 parameters:
- agreementClass
- callData
- userData

Most of the information is in callData, however this is just a jumble of bytes and expecting most users/developers to spend excessive time to break this down is an unrealistic expectation.

## Solution & Reasoning

Create an Agreement Forwarder Contract as an intermediate step before refactoring the core contracts to adhere to a token centric contract interface. This contract forwards your calls to the agreement contracts and will provide more visibility at the block explorer level on the parameters being passed for the agreement.

## Todos

- [x] Implement an Agreement Forwarder Contract for CFAv1
- [x] ~~Implement an Agreement Forwarder Contract for IDAv1~~
- [x] Deployment & Tests

d10r:


packages/ethereum-contracts/contracts/agreements/AgreementForwarder.sol
```diff
@@ -0,0 +1,79 @@
...
+    IConstantFlowAgreementV1 internal _cfa;
+
+    constructor(ISuperfluid host) {
+        _host = host;
+        _cfa = IConstantFlowAgreementV1(address(_host.getAgreementClass(
+            keccak256("org.superfluid-finance.agreements.ConstantFlowAgreement.v1")
+        )));
+    }
+
+    function createConstantFlow(
```
hellwolf: oh no, no more short hand `createFlow` :) ?

I don't mind, I would change the spec too.

d10r: Well, if the forwarder contract is for all agreements, `creatFlow` is probably not specific enough. 
I'm also considering abstracting the difference btw. create/update/delete into just `constantFlow()`.

hellwolf (dismissed):
Great start:

- I am not afraid and okay with introducing a new term `setFlow` in general, but let's get aligned with others who need to deal with its implications too.

To complete this end to end, these need to be sorted out:

- [ ] "Distribution", how can this be deployed and discovered by the downstream consumers?
- [ ] The main downstream consumer will be the SDK, and it would the place to hide these details from the user. In exchange, users get a bit more legible block scanner transaction data.

d10r:
> * The main downstream consumer will be the SDK, and it would the place to hide these details from the user.

It's actually a bit of a suprise to me that the SDK shall be the main consumer, because in my mind somebody using SF through the SDK already got a convenient API anyway and may not care much about what happens under the hood.  
On the other hand, I've seen the logic of setFlow() (figure out which of create/update/delete is needed) replicated in several contracts already, it causes quite some code bloat for a pretty common use case. Having it in the SDK wouldn't help with avoiding that overhead.  
I was assuming that the agreement forwarder shall provide a more accessible Solidity API too - complementing the agreement libs - with the advantage that you don't need to first deploy it yourself in order to use it. Or maybe it should even replace the agreement libs in order not to offer confusingly many options in the docs and examples.

Before defining "distribution", I think we need to better understand/define who we actually want to target as consumers.  
The one use case we clearly defined so far is: somebody wants to trigger an SF action via an Explorer - I would add executing SF actions via Gnosis Transaction Builder to that.  
I propose to first get some feedback from the devx team in that regard.

As for "setFlow", I see 2 questions:
* Are we ok with not qualifying it to be a "constant" flow? That relates to all methods in that contract: it doesn't in any way namespace them to be CFA or IDA specific (except on the code level itself via grouping and comment headers)
* Is "set" a good prefix for the aggregation of create/update/delete?

My reasoning here: it's a non-upgradable contract outside of the protocol (although with privileged access), not managing state.
If we need/want to better qualify/namespace methods, e.g. in order to add DFA functionality, we will create a new version of the contract and deploy it to a new address, not creating any backwards compatibility issues. If "upgrading" logic via new deployment to a different address vs via proxy, name changes are much less of a concern imo.

jtriley2p:
# Trusted Forwarder

The original design discussed was for the trusted forwarder to be a minimal agreement wrapper contract for end users,
not to provide novel functionality.

Ideally this contract will expose every agreement directly as a more intuitive way to call the core contracts
both from etherscan and from other wallets without the explicit need for the `sdk-core`.

## Requested Changes

Remove the `setFlow` and batching functions.

Add only functions that expose the agreements that may be called.

## Interface

```solidity
interface IAgreementForwarder {
  // CFA
  function createFlow(address token, address receiver, int96 flowRate) external;
  function updateFlow(address token, address receiver, int96 flowRate) external;
  function deleteFlow(address token, address sender, address receiver) external;
  function createFlowByOperator(address token, address sender, address receiver, int96 flowRate) external;
  function updateFlowByOperator(address token, address sender, address receiver, int96 flowRate) external;
  function deleteFlowByOperator(address token, address sender, address receiver) external;
  function authorizeFlowOperatorWithFullControl(address token, address flowOperator) external;
  function revokeFlowOperatorWithFullControl(address token, address flowOperator) external;
  function updateFlowOperatorPermissions(
    address token,
    address flowOperator,
    uint8 permissions,
    int96 flowRateAllowance
   ) external;

  // IDA
  function createIndex(address token, uint32 indexId) external;
  function updateIndex(address token, uint32 indexId, uint128 indexValue) external;
  function distribute(address token, uint32 indexId, uint256 amount) external;
  function approveSubscription(address token, address publisher, uint32 indexId) external;
  function revokeSubscription(address token, address publisher, uint32 indexId) external;
  function updateSubscription(address token, uint32 indexId, address subscriber, uint128 units) external;
  function deleteSubscription(address token, address publisher, uint32 indexId, address subscriber) external;
  function claim(address token, address publisher, uint23 indexId, address subscriber) external;
}
```

hellwolf:
Reduce the scope to CFA Forwarder

packages/ethereum-contracts/contracts/agreements/CFAv1Forwarder.sol
```diff
@@ -0,0 +1,493 @@
...
+    }
+
+    /**
+     * @notice Sets the given flowrate between msg.sender and a given receiver.
+     * If there's no pre-existing flow and `flowrate` non-zero, a new flow is created.
+     * If there's an existing flow and `flowrate` non-zero, the flowrate of that flow is updated.
+     * If there's an existing flow and `flowrate` zero, the flow is deleted.
+     * If the existing and given flowrate are equal, no action is taken.
+     * On creation of a flow, a "buffer" amount is automatically detracted from the sender account's available balance.
+     * This buffer is redeemed on deletion of that flow, if the sender account is then solvent (non-zero balance).
```
0xdavinchee: a nitpick, but I think maybe you meant to say critical here? also, the account is not critical/solvent when their balance is at 0

packages/ethereum-contracts/contracts/agreements/CFAv1Forwarder.sol
```diff
@@ -0,0 +1,493 @@
...
+
+    /**
+     * @notice Get all available information about a flow (if exists).
+     * If only the flowrate is needed, consider using `getFlowrate` instead.
+     * @param token Super token address
+     * @param sender The sender of the flow
+     * @param receiver The receiver of the flow
+     * @return lastUpdated Timestamp of last update (flowrate change) or zero if no flow exists
+     * @return flowrate Current flowrate of the flow or zero if no flow exists
+     * @return deposit Deposit amount locked. Returned to the the flow is deleted while solvent. TODO: rename to buffer?
```
0xdavinchee: just a reminder there is a TODO here
I would say that maybe it is a good idea to switch to buffer here, BUT it may be confusing because we use deposit everywhere else in the codebase. If we want to begin the transition to buffer at the code level, I think it would be helpful to put a comment in here (e.g. "buffer and deposit are the same thing") for those who are not familiar with it.

packages/ethereum-contracts/contracts/agreements/CFAv1Forwarder.sol
```diff
@@ -0,0 +1,493 @@
...
+    }
+
+    /**
+     * @notice Get aggregated flow information (if any exist) of an account.
+     * If only the net flowrate is needed, consider using `getAccountFlowrate` instead.
+     * @param token Super token address
+     * @param account Account to query
+     * @return lastUpdated Timestamp of last update of a flow to or from the account (flowrate change)
+     * @return flowrate Current net aggregate flowrate
+     * @return deposit Deposit amount locked. Returned to the the flow is deleted while solvent. TODO: rename to buffer?
```
0xdavinchee: reminder there's a TODO comment here

packages/ethereum-contracts/contracts/agreements/CFAv1Forwarder.sol
```diff
@@ -0,0 +1,493 @@
...
+                            token,
+                            sender,
+                            receiver,
+                            flowrate,
+                            new bytes(0) // placeholder
+                        )
+                    );
+            } else if (prevFlowRate != flowrate) {
+                // update flow
+                cfaCallData = cfaCallData = sender == msg.sender ?
```
0xdavinchee: looks like there is an extra assignment here

packages/ethereum-contracts/contracts/agreements/CFAv1Forwarder.sol
```diff
@@ -0,0 +1,493 @@
...
+                            receiver,
+                            new bytes(0) // placeholder
+                        )
+                    );
+            } else {
+                // do nothing
+                return;
+            }
+        } else {
+            revert("invalid flowrate");
```
0xdavinchee: custom error pls? 👀

packages/ethereum-contracts/contracts/agreements/CFAv1Forwarder.sol
```diff
@@ -0,0 +1,493 @@
...
+        }
+
+        _forwardBatchCall(address(_cfa), cfaCallData);
+    }
+
+    function _createFlow(ISuperToken token, address sender, address receiver, int96 flowrate, bytes memory userData)
```
0xdavinchee, line 281: I think we can and should reuse these functions below: `_createFlow`, `_updateFlow`, `_deleteFlow` above in the `_setFlowrateFrom` function - there is some duplication. The only difference I am noticing is that it takes a userData which will just be `new bytes(0)` in this case.

packages/ethereum-contracts/contracts/agreements/CFAv1Forwarder.sol
```diff
@@ -0,0 +1,493 @@
...
+                (
+                    token,
+                    flowOperator,
+                    permissions,
+                    flowrateAllowance,
+                    new bytes(0) // placeholder
+                )
+            );
+
+        // TODO: does non-zero userData make sense here?
```
0xdavinchee: no - probably not

packages/ethereum-contracts/scripts/deploy-deterministically.js
```diff
@@ -61,21 +93,12 @@ module.exports = eval(`(${S.toString()})()`)(async function (
...
-    const LoaderContract = new web3.eth.Contract(SuperfluidLoader.abi);
-    const deployTx = LoaderContract.deploy({
-        data: SuperfluidLoader.bytecode,
-        arguments: [resolverAddr],
+    const Contract = new web3.eth.Contract(ContractArtifact.abi);
+    const deployTx = Contract.deploy({
+        data: ContractArtifact.bytecode,
+        arguments: deployArgs,
     });
+    console.log("got Contract");
```
0xdavinchee: just want to double check if this is the console we want?

0xdavinchee (commented):
Code is real clean and nice overall, but just a few comments :)

packages/ethereum-contracts/contracts/agreements/CFAv1Forwarder.sol
d10r: hmm, I think the statement is correct: buffer redeemed (to the flow sender) if sender account solvent (non-zero balance here implying it's positive). But will see if I can rephrase it to be clearer.

packages/ethereum-contracts/contracts/agreements/CFAv1Forwarder.sol
d10r: right, I decided to stick to the current names and introduce "buffer" only in the param description

packages/ethereum-contracts/contracts/agreements/CFAv1Forwarder.sol
d10r: oh damn, missed that possible simplification. Love you!

packages/ethereum-contracts/contracts/agreements/CFAv1Forwarder.sol
0xdavinchee: ah okay gotcha

packages/ethereum-contracts/contracts/agreements/CFAv1Forwarder.sol
```diff
@@ -0,0 +1,413 @@
...
+import { CallUtils } from "../libs/CallUtils.sol";
+
+/**
+ * The CFAv1Forwarder contract provides an easy to use interface to
+ * ConstantFlowAgreementV1 specific functionality of Super Tokens.
+ * Instances of this contract can operate on the protocol only if configured as "trusted forwarder"
+ * by protocol governance.
+ */
+contract CFAv1Forwarder {
+    error CFAFwd_InvalidFlowrate();
```
0xdavinchee: I think we should maintain our `SNAKE_CASE` standard for custom errors

hellwolf: yes, let's do that. we can check what we can do in .solhint configuration, "custom error" probably is  still too early for it to support, but it does have some support of snake case enforcement for immutable/constant variables.

0xdavinchee: took the liberty to address this quickly

0xdavinchee (approved):
looks good, lgtm! 💪🏼

hellwolf (approved):
LFG

PR #983, opened 2022-07-20
[SDK-core] Use `serialize-error`

kasparkallas:
Why?
`JSON.stringify` is not sufficient for serializing errors:
image

`serialize-error` is a popular small library that handles error serialization: https://www.npmjs.com/package/serialize-error


0xdavinchee (approved):
Looks good - LGTM

PR #984, opened 2022-07-20
create operation from framework

### LINKED ISSUE #980: [SDK-CORE BUG]  Unable to access Operation from Framework
**Describe the bug:**
When creating a new Operation directly with the SDK core , you're unable to get the Operation class from either the Framework object or from the default imports.

Additionally, when importing it like so: `const Operation = require("@superfluid-finance/sdk-core/dist/main/Operation");`
Users need to create new operations by calling `new Operation.default()` rather than just `new Operation()`

**Environment (please complete the following information):**
  - Device: [e.g. Macbook Pro - Apple Silicon]
  - OSS: OSX
  - Network/Connection: WiFi
  - Project Type: module
  - Package version: [e.g. ^0.4.4]
  - TypeScript/Javascript: Javascript
  - Blockchain interfacing library: ethers
  - Provider: Alchemy
  - Chain: Goerli, other testnets
  - If running in node:
    - Node version: v14.17.0
    - Framework: Hardhat

**Steps To Reproduce:**
1. Initialize framework with Framework.create()
2. Try creating a new `Operation` (i.e. for usage within a batch call) as depicted [here](https://docs.superfluid.finance/superfluid/developers/sdk-initialization/sdk-core/batch-calls)
3. Get `TypeError: Operation is not a constructor`

Current Workaround:
Import the Operation class directly from the SDK like so: 
`const Operation = require("@superfluid-finance/sdk-core/dist/main/Operation");` 
the operation class is accessible by calling `new Operation.default(...)`

**Expected behavior:**
Ideally, the Operation class is available directly from the Framework object, or as something that can be imported along with the Framework from the SDK Core. Ideally, users could access the Operation class in one of the following two ways:

- `const { Framework, Operation } = require("@superfluid-finance/sdk-core)`
- [after the Framework has been initialized] `new sf.Operation(...)`

**Additional context:**
I needed to use the Operation class when using `callAppAction` within a batch call

0xdavinchee:
Related to #980

### DIFF
--- a/packages/sdk-core/CHANGELOG.md
+++ b/packages/sdk-core/CHANGELOG.md
@@ -9,6 +9,7 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 
 ## Added
 - Support for SetEvent and Subgraph v1.4.4
+- `Framework.operation` method for easily creating `Operation` instances
 
 ## Fixes
 - Compile AJV validations to prevent unsafe-eval and did not allow SDK-Core usage inside Google Chrome extension due to unsafe-eval CSP
@@ -16,6 +17,7 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 ### Changed
 - `SFError` refactor to be more conventional. It inherits `Error` and uses `cause` to wrap internal errors.
 - Use `serialize-error` for serializing error object inside the message.
+- Export Operation & OperationType
 
 ### Breaking
 - `SFError.errorObject` renamed to `SFError.cause`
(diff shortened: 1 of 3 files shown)


hellwolf:
@kasparkallas ?

I don't have context to review this one. But it looks small and fine.

kasparkallas:
`Operation` type is still not exported from `@superfluid-finance/sdk-core` for anyone that wants to reference the _type_, not the instance (created by `sf.operation(...`).

Should probably be exported same as `BatchCall`:
image

Also changelog text missing.

kasparkallas (approved):

PR #995, opened 2022-07-25
[SDK-Core] Call App Action Support Added/Fixed CallAppAction BatchCall

### LINKED ISSUE #973: [SDK-CORE BUG] - CallAppAction batchCall doesn't work
## Bug Description

SDK-Core `batchCall` does not properly handle a `CALL_APP_ACTION` Operation. The batchable operation succeeds when called directly on the Superfluid host contract and when used with the JS-SDK, so the issue appears to be specific to SDK-Core.

## Relevant Environment Information

  - Project Type: node
  - Package version: 0.4.4
  - TypeScript/Javascript: Typescript
  - Blockchain interfacing library: ethers js
  - Node version: v16.6.1
  - Framework: Hardhat

## Reproduce Steps

1. Initialize framework in any environment
2. Create ethers signer
3. Create Operation with operation type "CALL_APP_ACTION"
4. Exec in a batch call
5. Transaction fails

## Expected Behavior

"CALL_APP_ACTION" with a function whose only argument is `bytes memory ctx` and that returns `bytes memory newCtx` should, when called by the SDK-Core batch call should succeed.

## Actual Behavior

The call always reverts with "Call Utils: target reverted".

## Context

It appears this may be due to the fact that the function selector is removed when the operation type is not "SUPERFLUID_CALL_AGREEMENT". This may be fixable by not removing the function selector when the operation type is "CALL_APP_ACTION".

Source: https://github.com/superfluid-finance/protocol-monorepo/blob/763ca6966046ad0f959aab86d11c5b50c9e00e70/packages/sdk-core/src/BatchCall.ts#L103

0xdavinchee:
* Call App Action Operation creator added in Host class
* Call App Action support in batchCall fixed
* SuperAppTester basic contract added for testing call app action
* Tests added for:
    * call app action executed from operation
    * call app action executed from batch call
* **[BREAKING]** `BatchCall.getCallAgreementFunctionArgs` deprecated and removed
* `BatchCall.getCallDataFunctionArgs` is replacing it
* CHANGELOG updated and formatting cleanup
* yarn.lock updated


packages/sdk-core/tasks/setupTestEnvironment.sh
```diff
@@ -16,4 +16,6 @@ yarn deploy-contracts-local
 cd ../sdk-core
 yarn generate:graphql-types
 yarn generate:abi-files
-yarn generate:web3-types
\ No newline at end of file
+yarn generate:web3-types
+# compile contracts before starting tests to generate typechain folder
+npx hardhat compile
```
kasparkallas, line 21: Tests slower now?

packages/sdk-core/src/Host.ts
```diff
@@ -40,4 +40,24 @@ export default class Host {
...
     };
+
+    /**
+     * Creates an Operation of the `callAppAction` function on the host contract.
+     * @param app the address of the Super App you are calling
+     * @param callData the encoded callData for the function
+     * @param overrides ethers overrides object for more control over the transaction sent.
+     * @returns {Operation} an `Operation` class
+     */
+    populateCallAppActionTxnAndReturnOperation = (
```
kasparkallas: The function name is the whole implementation. That's a bit weird.

packages/sdk-core/.gitignore
```diff
@@ -4,3 +4,5 @@ coverage
 /src/abi
 /src/typechain
 *.generated.ts
+/typechain-types
```
kasparkallas, line 7: We were including these before?

packages/sdk-core/tasks/setupTestEnvironment.sh
0xdavinchee: yup, it'll take extra time to compile the contract

packages/sdk-core/src/Host.ts
0xdavinchee: touché, will rename as: `callAppAction`

packages/sdk-core/tasks/setupTestEnvironment.sh
kasparkallas: Approximately how much slower?

packages/sdk-core/.gitignore
0xdavinchee: no `typechain-types` is generated from the `.sol` file when you run `hardhat compile`, the other `/src/typechain` folder is generated from abi files which we pull from `ethereum-contracts` as part of the build process
`artifacts` are also generated from contract compilation

packages/sdk-core/tasks/setupTestEnvironment.sh
hellwolf: 1 pre-throw routing of the disk

kasparkallas: ~20 seconds

kasparkallas (approved):

packages/sdk-core/tasks/setupTestEnvironment.sh
hellwolf: @0xdavinchee need harder hat

0xdavinchee: we actually timed it, it takes an extra 5s the first time locally, probably ~20s in build and in subsequent local test runs, almost instant because it won't recompile

PR #1001, opened 2022-07-26
SDK-Core 0.5.2 Patch Fix and Version Bump

0xdavinchee:


hellwolf (approved):
["the little patch that could"](https://en.wikipedia.org/wiki/The_Little_Engine_That_Could)

PR #1024, opened 2022-08-08
Fix release branch

0xdavinchee:


0xdavinchee:
should merge to dev, then will cherry pick into release branch if all is well

.github/workflows/handler.publish-release-packages.yml
```diff
@@ -43,22 +43,23 @@ jobs:
...
+          tasks/npm-publish.sh packages/ethereum-contracts/ latest
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
           NPMJS_TOKEN: ${{ secrets.NPMJS_TOKEN }}
 
       - name: Publish js-sdk@latest package
         if: env.PUBLISH_JS_SDK == 1
-        run: tasks/npm-publish.sh packages/js-sdk/ latest
+        run: |
+          yarn --cwd packages/js-sdk build
```
hellwolf, line 62: unfortunately js-sdk build requires ethereum-contracts build.

.github/workflows/handler.publish-release-packages.yml
```diff
@@ -107,6 +110,7 @@ jobs:
         id: publish-sdk-redux
         if: env.PUBLISH_SDK_REDUX == 1
         run: |
+          yarn --cwd packages/sdk-redux build
```
hellwolf, line 115: the build dependencies start to become harder to reason now with these code, not sure if there is a good solution other than write build dependencies explicitly here

0xdavinchee: yeah, I agree
SDK-Core and JS-SDK both need contracts built, looking at sdk-redux and it looks like it doesn't have any build dependencies

.github/workflows/handler.publish-release-packages.yml
0xdavinchee: resolved

hellwolf (approved):

PR #1025, opened 2022-08-09
add tokenDecimals option to deploy-test-token.js

### LINKED ISSUE #994: [ETHEREUM-CONTRACTS] Add decimals as an option to `deploy-test-token.js`
## Summary
Currently our `deploy-test-token.js` script hardcodes the number of decimals to `18` when deploying the `TestToken` contract. 
However, @kasparkallas brought up a good point which is that there are many ERC20 tokens in the wild which are part of our resolver and prevalent in crypto (USDC) which have a different number of decimals. 
Allowing the deployment of test token with a custom number of decimals is going to be useful for testing logic along the entire stack (contracts to client) and we should consider redeploying the fUSDC tokens in the testnets which won't be deprecated with the correct number of decimals (6) so we can properly test this logic internally as well as allowing external devs building on top of Superfluid to have a more realistic testing experience.

## TODO
- [x] Allow another property to be included when running `deploy-test-token.js`: `TOKEN_DECIMALS`, if it is left empty, default to the standard 18
- [ ] (OPTIONAL) Redeploy fUSDC with 6 decimals and remove the current fUSDC from the resolver and add this one to the resolver

0xdavinchee:
https://github.com/superfluid-finance/protocol-monorepo/issues/994

### DIFF
--- a/packages/ethereum-contracts/scripts/deploy-test-token.js
+++ b/packages/ethereum-contracts/scripts/deploy-test-token.js
@@ -16,7 +16,7 @@ const {
  * @param {Address} options.from Address to deploy contracts from
  * @param {boolean} options.resetToken Reset the token deployment
  *
- * Usage: npx truffle exec scripts/deploy-test-token.js : {TOKEN_SYMBOL}
+ * Usage: npx truffle exec scripts/deploy-test-token.js : {TOKEN_DECIMALS} {TOKEN_SYMBOL}
  */
 module.exports = eval(`(${S.toString()})()`)(async function (
     args,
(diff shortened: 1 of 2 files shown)


packages/ethereum-contracts/scripts/deploy-test-token.js
```diff
@@ -16,7 +16,7 @@ const {
  * @param {Address} options.from Address to deploy contracts from
  * @param {boolean} options.resetToken Reset the token deployment
  *
- * Usage: npx truffle exec scripts/deploy-test-token.js : {TOKEN_SYMBOL}
+ * Usage: npx truffle exec scripts/deploy-test-token.js : {TOKEN_DECIMALS} {TOKEN_SYMBOL}
```
hellwolf, line 19: okay...

hellwolf (approved):

packages/ethereum-contracts/scripts/deploy-test-token.js
d10r: appending the arg and making it optional (keep default value) would have avoided breaking the interface

PR #1032, opened 2022-08-11
Deprecate networks

### LINKED ISSUE #918: [SDK-CORE/SUBGRAPH] - Use metadata repo in SDK-Core/Subgraph
## Problem

* There is duplicate information for the different networks we are deployed to in multiple packages within protocol-monorepo

## Todos

* The data included [here](https://github.com/superfluid-finance/metadata) should be reused in as many locations as possible for: sdk-core, subgraph, etc.

0xdavinchee:


packages/sdk-core/package.json
```diff
@@ -56,6 +56,7 @@
         "url": "https://github.com/superfluid-finance/protocol-monorepo/issues"
```
hellwolf, line 56: need to bump the version again

hellwolf (approved):

PR #1033, opened 2022-08-11
Use Hardhat Compile

### LINKED ISSUE #976: [TOOLING - ETHEREUM-CONTRACTS] Use Hardhat compile as build instead of truffle compile
We are currently using truffle compile when building contracts, although hardhat is faster.
The reason we haven't switched over yet is because of the `build:contracts-size` and `build-abi-json.sh` commands which relies on the structure of the build output to be what truffle outputs. 
We can either use a (pre-built package)[https://www.npmjs.com/package/hardhat-contract-sizer] or modifying the `print-contract-sizes.sh` and `build-abi-json.sh` scripts for hardhat (we could keep the one for truffle and create a new one for hardhat based off of it).

npx hardhat compile  25.14s user 1.13s system 94% cpu 27.886 total
yarn build:contracts  63.44s user 0.88s system 130% cpu 49.447 total

0xdavinchee:
* use hardhat compile
* shell scripts for flattening folder and cleaning up the build artifacts


0xdavinchee:
this breaks js-sdk as it loads contracts using truffle artifacts and this doesn't play well with linking
this breaks the deploy-framework script which is used by: js-sdk, sdk-core and subgraph so these packages also get broken by this
solution: write a deploy-framework script which works for truffle + hardhat, should only care about ABIs, can use ethers under the hood

PR #1043, opened 2022-08-22
[ETHEREUM-CONTRACTS] Add Custom Errors

### LINKED ISSUE #686: [ETHEREUM-CONTRACTS] Use custom error revert feature from solidity 0.8.0
After migration to Solidity 0.8.0, one feature we should start to use is the custom error revert.

## Checklist

- [x] Use new revert patter `if (!errorCheckSucceeded) revert CustomError(...)`
- [ ] Resolve the test expectRevert issue with openzeppelin-testhelpers: https://github.com/OpenZeppelin/openzeppelin-test-helpers/issues/180
- [x] Make callUtils to be aware of the custom error: https://github.com/superfluid-finance/protocol-monorepo/issues/625

0xdavinchee:
## Tips for Reviewing PR

There are 74 files changed, but do not fret fren, it ain't that bad.
The most important changes are in the contracts where we move from require w/ error string to if w/ revert custom error.
To ensure this is done correctly, it's easy to just check that the check has been flipped, e.g. `require(a != b && c > f)` becomes `if (a == b || c <= f)`.
Most of the other changes are just as trivial.

## Overall Changes

* Bump interface files to 0.8.4
* Added Custom Error support in **most** contracts - left some out because it broke tests and it was simpler to not change it in certain places
* Refactor tests to properly expect custom errors
* Added AgreementHelper class


packages/ethereum-contracts/contracts/agreements/ConstantFlowAgreementV1.sol
```diff
@@ -140,9 +140,11 @@ contract ConstantFlowAgreementV1 is
          internal pure
          returns (uint256 deposit)
      {
-         require(flowRate > 0, "CFA: not for non-positive flow rate");
-         require(uint256(int256(flowRate)) * liquidationPeriod <= uint256(int256(type(int96).max)),
-             "CFA: flow rate too big");
+        // @note not sure why this was > 0 before, should be allowed to be 0?
```
0xdavinchee: follow up: this is safe to do, it is being used by `getDepositRequiredForFlowRate` which is an external function that has a check: `if (flowRate == 0) return 0;`, so the revert should never be hit if it is 0

packages/ethereum-contracts/test/scripts/deployment.test.js
```diff
@@ -210,14 +211,19 @@ contract("Embeded deployment scripts", (accounts) => {
...
+                    await deployFramework(errorHandler, {
                         ...deploymentOptions,
                         nonUpgradable: true,
                         useMocks: true, // force an update attempt
-                    }),
-                    "SF: non upgradable"
-                );
+                    });
+                } catch (err) {
+                    if (process.env.IS_TRUFFLE) {
```
0xdavinchee, line 220: follow up: we cannot use the expectCustomError logic because the return value/error of the deployFramework code doesn't interface with the expectCustomError code by hardhat.

packages/ethereum-contracts/test/contracts/agreements/ConstantFlowAgreementV1.behavior.js
```diff
@@ -219,9 +224,15 @@ async function _shouldChangeFlow({
             break;
         case "createFlowByOperator":
         case "updateFlowByOperator":
-            tx = await testenv.contracts.superfluid.callAgreement(
+            cfa = await testenv.sf.contracts.IConstantFlowAgreementV1.at(
```
0xdavinchee, line 227: follow up: we get the truffle contracts here again (instead of the ethers ones initialized in testenvironment) in order to be able to expect the events via the tx

packages/ethereum-contracts/contracts/agreements/AgreementBase.sol
```diff
@@ -15,6 +14,8 @@ abstract contract AgreementBase is
 {
     address immutable internal _host;
 
+    error AgreementBase_OnlyHost();
```
hellwolf: hmmm, we probably should just have one `OnlyHost`  error across, for example, move to a library "SuperfluidErrors" or even better if it could be defined as global instance.

hellwolf (approved):
- looks good in general
- need to be extra careful the `require(b, estr)` vs if `(b) rervert err` conversion.
- piggyback the find-dup-devDependencies.sh check

PR #1046, opened 2022-08-24
[SDK-core] Don't wrap transaction errors with `SFError` & don't serialize a massive internal error into `SFError` message

kasparkallas:
Steadily moving towards a more optimal error-handling solution based on feedback and usage.

Ideas:
* There is value to `SFError`, especially on `Framework` initialization.
* There is value to serializing the cause into `SFError` but it should be in much more human-friendly minimal format.
* `SFError` sometimes wraps transaction execution errors multiple times which only obfuscates the actual cause of the error. Even for common situations like "user rejected transaction".

Changes:
* Let original error bubble up when executing `Operation` & `BatchCall`.
* Serialize _a much smaller_ version of the internal error (by using the small `miniSerializeError` snippet from RTK).
* Remove `serialize-error` dependency.
* Remove escaping of the quotes from the serialized internal error.

Example of `SFError` with a cause:
```
SFError: Framework Initialization Error: There was an error initializing the framework
Caused by: {
  "name": "Error",
  "code": "CALL_EXCEPTION",
  "message": "call revert exception [ See: https://links.ethers.org/v5-errors-CALL_EXCEPTION ] (method="get(string)", data="0x", errorArgs=null, errorName=null, errorSignature=null, reason=null, code=CALL_EXCEPTION, version=abi/5.6.4)"
}
```

### DIFF
--- a/packages/sdk-core/CHANGELOG.md
+++ b/packages/sdk-core/CHANGELOG.md
@@ -5,6 +5,10 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 
 ## [Unreleased]
 
+### Breaking
+- Don't throw `SFError` when executing `Operation` or `BatchCall`; let the original error bubble up
+- Serialize a much smaller version of the cause in `SFError` (only `name`, `message`, `code`)
+
 ## [0.5.5] - 2022-08-31
 ### Added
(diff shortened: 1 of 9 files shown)


0xdavinchee (approved):
v nice

Issue #1039, opened 2022-08-18
[ETHEREUM-CONTRACTS] ]add gov wrapper function for creating registration keys


hellwolf:
To be clarified.

0xdavinchee:
pretty sure https://github.com/superfluid-finance/protocol-monorepo/pull/1049 addresses this issue

PR #1049, opened 2022-08-25
Gov: Convenience method for app registration keys and some cleanup

### LINKED ISSUE #915: [GOV-PROC] Using setConfig for deployer address shows very obscure info
image

Since `setConfig` is the very low level operation, it doesn't show nice high level information in the governance dashboard e.g. `whitelistDeployer(deployerAddress, expiration timestamp)`. Without nice high level information, it slows down governance operator from doing proper due diligence. 

The solution is to do a governance logic update to add newer high level functions to the "facade".

d10r:
We wanted to (re-)introduce (context: https://github.com/superfluid-finance/protocol-monorepo/pull/769) convenience methods for governing app registration keys, see https://github.com/superfluid-finance/protocol-monorepo/issues/1039.

After an incident with key handout (by mistake communicated "app keys" instead of the needed "registration keys", I simplified the code to not expose the "app key" aka "config key" (that's what it basically was: the config key for the gov module KV store) to the public gov interface.

Piggy-backed:
* emit event when changing app factory authorizations
* removed `clearTrustedForwarder` and changed `disableTrustedForwarder` to use `_clearConfig` internally. (assuming there's no good reason to do it otherwise, or is there?)
* more consistent syntax and code in other convenience methods (no functional change), e.g. don't use `return` for `_setConfig` and `_clearConfig` which don't return anything.

### DIFF
--- a/packages/ethereum-contracts/contracts/gov/SuperfluidGovernanceBase.sol
+++ b/packages/ethereum-contracts/contracts/gov/SuperfluidGovernanceBase.sol
@@ -365,7 +365,7 @@ abstract contract SuperfluidGovernanceBase is ISuperfluidGovernance
     )
         public
     {
-        return _clearConfig(host, superToken, SuperfluidGovernanceConfigs.SUPERTOKEN_MINIMUM_DEPOSIT_KEY);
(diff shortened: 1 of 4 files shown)


0xdavinchee (approved):
looks good, seems much simpler and intuitive to set up registration keys now ✅

hellwolf (approved):

PR #1057, opened 2022-08-29
remove hardhat deployer

jtriley2p:
The hardhat deployer has seen diminishing returns given the latest updates to the ethereum contracts, particularly the SuperfluidFrameworkDeployer.

It would be a better use of time to direct developers to a resource indicating how the framework deployer contract should be used rather than maintaining an npm package.


hellwolf:
okay, I will see what I can do with the published package https://www.npmjs.com/package/@superfluid-finance/hardhat-deployer, maybe can mark it as deprecated: https://docs.npmjs.com/unpublishing-packages-from-the-registry

hellwolf:
deleted

hellwolf (approved):

PR #1062, opened 2022-08-31
yAcademy Gas Savings Applied

0xdavinchee:
This PR applies the following gas savings across the codebase:
- ~use != 0 instead of > 0 for uint's~
- ~redundant 0 initializations removed (uint/int's don't need to be initialized to 0, that is the default value)~
- prefix ++ in for loops
- use constant internal over immutable where applicable
- use calldata instead of memory where applicable
- use external over public 
- remove fixme and other small optimizations (unnecessary casting, computations, using cached variables)


hellwolf:
so, no visible improvements so far per gas measuring right? Just to have some sanity check before we close it off, could it be any chance due to the way we run the tests (using mock contracts wrapper, etc.) ?

0xdavinchee:
> so, no visible improvements so far per gas measuring right? Just to have some sanity check before we close it off, could it be any chance due to the way we run the tests (using mock contracts wrapper, etc.) ?

no real extreme improvements in terms of gas optimizations, just low hanging fruit which might shave off double digits, the most impactful is likely the switch between memory and calldata as a param for batchCall and this one is more noticeable

hellwolf (approved):
underwhelming meh

PR #1066, opened 2022-09-01
Formatted Contracts Demo

0xdavinchee:
Decided to do a super quick experiment with prettier-plugin-solidity to show what our contracts would look like formatted by prettier w/ our prettier settings.

Curious what ya'll think


hellwolf:
Nice exercise, unfortunately the negative effects outweighs any benefits:

The main negative effect is it destroys git history and making tracing regression introduced by code change very hard for anyone.

--

But as a useful exercise, we could do a 45 minutes history lesson of various brackets related coding styles in other language, and get some informed opinion for a coding style standard we can feel objective "benefits" rather than subjective opinions.

:) But historically, it maybe be a bit futile thing, since this is the tab vs space type of things.

PR #1074, opened 2022-09-07
[SDK-CORE] 0.5.6 Patch Fix

0xdavinchee:
* Fix broken `subgraphAPIEndpoint`

### DIFF
--- a/packages/sdk-core/CHANGELOG.md
+++ b/packages/sdk-core/CHANGELOG.md
@@ -5,6 +5,10 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 
 ## [Unreleased]
 
+## [0.5.6] - 2022-09-07
+### Fixes
+- Correct `subgraphAPIEndpoint` in `getResolverData`
+
 ## [0.5.5] - 2022-08-31
 ### Added
 - Support for: `optimism-goerli` and `arbitrum-goerli` added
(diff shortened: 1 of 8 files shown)


packages/sdk-core/src/constants.ts
```diff
@@ -44,7 +44,9 @@ const getResolverData = (chainId: number): IResolverData => {
             nativeTokenSymbol: "",
         };
     return {
-        subgraphAPIEndpoint: `${baseUrl}-${networkData?.shortName}`,
+        subgraphAPIEndpoint: envReleaseTag
+            ? `${baseUrl}-${networkData?.name}`
```
0xdavinchee: using `name` instead of `shortName`

packages/sdk-core/src/constants.ts
```diff
@@ -44,7 +44,9 @@ const getResolverData = (chainId: number): IResolverData => {
...
+            : networkData.subgraphV1.hostedEndpoint,
```
0xdavinchee: we use `.hostedEndpoint` outside of tests w/ a `SUBGRAPH_RELEASE_TAG` env variable set

packages/sdk-core/test/5_subgraph.test.ts
```diff
@@ -36,5 +37,16 @@ describe("Subgraph Tests", () => {
         it("Should be able to use the listener", async () => {
             await testListenerInitialization(query);
         });
+
+        it("Should have the correct subgraph endpoints", async () => {
```
0xdavinchee, line 41: Just a sanity check to ensure that the endpoints returned from the resolverData map is correct

packages/sdk-core/CHANGELOG.md
```diff
@@ -220,7 +224,8 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
   - New `SuperToken` class with `SuperToken` CRUD functionality and an underlying `Token` class with basic `ERC20` functionality
   - New `BatchCall` class for creating and executing batch calls with supported `Operation's`
 
-[Unreleased]: https://github.com/superfluid-finance/protocol-monorepo/compare/sdk-core%40v0.5.5...HEAD
+[Unreleased]: https://github.com/superfluid-finance/protocol-monorepo/compare/sdk-core%40v0.5.6...HEAD
+[0.5.6]: https://github.com/superfluid-finance/protocol-monorepo/compare/sdk-core%40v0.5.4...sdk-core%40v0.5.6
```
kasparkallas: Should be compared to v0.5.5 instead?

packages/sdk-core/test/5_subgraph.test.ts
```diff
@@ -36,5 +37,35 @@ describe("Subgraph Tests", () => {
...
+                            const query = new Query({
+                                customSubgraphQueriesEndpoint:
+                                    x.subgraphAPIEndpoint,
+                            });
+                            const event = await query.listEvents(
+                                {},
+                                { take: 1 }
+                            );
+                            expect(event.data.length).to.be.greaterThan(0);
+                        }
```
kasparkallas, line 67: The deep nesting makes the test a bit complicated. I can imagine it's easily possible that the test will always pass because it never gets through all the conditionals. Maybe it can be simplified a bit?

packages/sdk-core/CHANGELOG.md
0xdavinchee: good catch

kasparkallas (approved):

PR #1075, opened 2022-09-07
[SDK-core] Don't wrap SubgraphClient's errors with SFError

kasparkallas:
Why?
Pretty useless wrapping which again just creates obfuscation.


0xdavinchee (approved):
gm

PR #1080, opened 2022-09-19
[ETHEREUM-CONTRACTS] Patrician Period Patch

0xdavinchee:
This fixes an issue where if `cfa._isPatricianPeriod` is run with `signedTotalCFADeposit = 0`, this view function will revert with: `Panic(uint256)` because of division by 0.

This happens because of the following formulas:
```
int256 totalCFAOutFlowrate = signedTotalCFADeposit / int256(liquidationPeriod);
// totalCFAOutFlowrate = 0 / something arbitrary non-zero number = 0

return totalRewardLeft / totalCFAOutFlowrate > int256(liquidationPeriod - patricianPeriod);
// return: some arbitrary number / 0
```

### DIFF
--- a/packages/ethereum-contracts/contracts/agreements/ConstantFlowAgreementV1.sol
+++ b/packages/ethereum-contracts/contracts/agreements/ConstantFlowAgreementV1.sol
@@ -1432,6 +1432,10 @@ contract ConstantFlowAgreementV1 is
         internal pure
         returns (bool)
     {
+        if (signedTotalCFADeposit == 0) {
+            return false;
+        }
+
         int256 totalRewardLeft = availableBalance + signedTotalCFADeposit;
         int256 totalCFAOutFlowrate = signedTotalCFADeposit / int256(liquidationPeriod);
         // divisor cannot be zero with existing outflow
--- a/packages/ethereum-contracts/nodemon.json
+++ b/packages/ethereum-contracts/nodemon.json
@@ -1,0 +1,3 @@
+{
+    "ignore": ["typechain-types", "artifacts", "cache", "build"]
+}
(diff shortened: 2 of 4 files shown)


d10r (approved):

hellwolf (changes requested):
locally this code makes no sense, how is signedTotalCFADeposit make it fail? Fix it in the right place instead.

In another words, this function must be pure and has no implicit assumptions baked in anywhere.

0xdavinchee:
> locally this code makes no sense, how is signedTotalCFADeposit make it fail? Fix it in the right place instead.
> 
> In another words, this function must be pure and has no implicit assumptions baked in anywhere.

I don't agree that this code makes no sense, please take another look...

If `signedTotalCFADeposit = 0`, then `0 / signedTotalCFADeposit = 0` AND `totalCFAOutFlowrate = 0`.
if `totalCFAOutFlowrate = 0`, then `totalRewardLeft / totalCFAOutFlowrate` will lead to a revert because of division by 0.

The function is already an internal pure function with no implicit assumptions baked in anywhere - everything is passed in and it has no knowledge on anything except for what the function is trying to check with the inputs.

What am I missing here?

hellwolf:
> > locally this code makes no sense, how is signedTotalCFADeposit make it fail? Fix it in the right place instead.
> > In another words, this function must be pure and has no implicit assumptions baked in anywhere.
> 
> I don't agree that this code makes no sense, please take another look...
> 
> If `signedTotalCFADeposit = 0`, then `0 / signedTotalCFADeposit = 0` AND `totalCFAOutFlowrate = 0`. if `totalCFAOutFlowrate = 0`, then `totalRewardLeft / totalCFAOutFlowrate` will lead to a revert because of division by 0.
> 
> The function is already an internal pure function with no implicit assumptions baked in anywhere - everything is passed in and it has no knowledge on anything except for what the function is trying to check with the inputs.
> 
> What am I missing here?

okay, I read too quickly.

hellwolf (approved):

PR #1082, opened 2022-09-19
[SDK-CORE] Fix Batch Call Expect

0xdavinchee:
This test keeps failing in Canary and is not always caught in `ci.feature` as sdk-core tests aren't always run.

* this test just needs to know something is undefined
* in the workflow it lets us know which property is undefined, but locally it gives the original message: "Cannot read properties of undefined"


hellwolf (approved):
mkay, I though it should be run in feature branch if sdk-core code changes.

PR #1094, opened 2022-09-28
[ETHEREUM-CONTRACTS] CFAv1Forwarder Fixes

0xdavinchee:
* make _host and _cfa state variables immutable - these should not change
* fix `_deleteFlow` logic
* fix up natspec a bit
* return true from _forwardBatchCall
* add comment about EIP 2771
* add CFAv1Forwarder tests
* add CFAv1Forwarder to deploy-framework + resolver


packages/ethereum-contracts/scripts/deploy-framework.js
```diff
@@ -436,6 +438,18 @@ module.exports = eval(`(${S.toString()})({skipArgv: true})`)(async function (
         }
     }
 
+    // deploy CFAv1Forwarder
+    await deployAndRegisterContractIf(
+        CFAv1Forwarder,
+        `CFAv1Forwarder.${protocolReleaseVersion}`,
```
hellwolf: similar to loader:

```
    await deployAndRegisterContractIf(
        SuperfluidLoader,
        "SuperfluidLoader-v1",
```

The forwarder should be for all protocol release versions.

hellwolf (approved):

Issue #1096, opened 2022-09-28
[ETHEREUM-CONTRACTS] Configurable global CFA hooks.


hellwolf:
## Notes

- the hook address (ICFAGlobalHook) is an immutable field in the CFA logic contract.
- if the hook is set, it is invoked for each create/update/delete flows.
- the hook actions are try/catched, so no danger of blocking CFA operations globally.

0xdavinchee:
gonna close this cause this feature is merged

PR #1099, opened 2022-09-30
[ETHEREUM-CONTRACTS] CFA Hooks

0xdavinchee:
* IConstantFlowAgreementHook interface API implementation
* IConstantFlowAgreementHook used in ConstantFlowAgreeementV1 per spec in https://github.com/superfluid-finance/protocol-monorepo/issues/1096 
* Hook mock contracts added for testing
* Tests

> Note: It is important that we add the actual hook contracts in the places where I've placed `@note`


packages/ethereum-contracts/contracts/interfaces/agreements/IConstantFlowAgreementHook.sol
```diff
@@ -0,0 +1,47 @@
...
+    /// @notice A hook which executes on stream creation if the hook contract is set in the CFA
+    /// @dev This should be implemented with an onlyCFA modifier, so that only the CFA can call the function
+    /// @param newFlowData new flow data, see ConstantFlowAgreementV1 to see properties on FlowParams struct
+    /// @param token the streamed super token
+    /// @param oldFlowRate previous flowrate
+    /// @return bool
+    function onCreate(
+        ConstantFlowAgreementV1.FlowParams memory newFlowData,
+        ISuperfluidToken token,
+        int96 oldFlowRate
```
ngmachado: On create, there is no `oldFlowRate`.  We should remove this.

ngmachado (commented):
I'm worry about the eventual additional cost of this hook. Users don't have any opt-out if they don't need/want this hook.

- Hook is immutable. This can be a problem if we need to update `constantFlowAgreementHook` 
   - 1 - redeploy a new CFA every time we need to upgrade `constantFlowAgreementHook`
   - 2 - Be upgradable by governance action.
   - 3 - Make explicit that `constantFlowAgreementHook` should implement a upgrade mechanism like proxy.
   
   _More incline to solution 2 or 3._

packages/ethereum-contracts/contracts/interfaces/agreements/IConstantFlowAgreementHook.sol
hellwolf: (1) is the way for now, we push any upgradability problem to  the edge

Specifically, talking about that "hooked" contract, it's more relevant to think if that one should be upgradable or not. Because it might be a NFT contract :roll_eyes: ??

ngmachado: Yes, you are correct, that should be another discussion not related with this implementation itself.

ngmachado (approved):

hellwolf:
I will give it a few more reviews.

@d10r

d10r:
fyi: @ngmachado wants to keep this open for a bit more to make sure the hook includes all needed data

packages/ethereum-contracts/contracts/interfaces/agreements/IConstantFlowAgreementHook.sol
```diff
@@ -0,0 +1,50 @@
...
+    ) external returns (bool);
+
+    /// @notice A hook which executes on stream deletion if the hook contract is set in the CFA
+    /// @dev This should be implemented with an onlyCFA modifier, so that only the CFA can call the function
+    /// @param newFlowData the new flow data taken by the hook
+    /// @param token the streamed super token
+    /// @param oldFlowRate previous flowrate
+    /// @return bool
+    function onDelete(
```
d10r: Methods of the IConstantFlowAgreementV1 interface have the token argument as first argument. Should imo be the same here.

d10r (commented):
lgtm, 1 change proposal

packages/ethereum-contracts/contracts/interfaces/agreements/IConstantFlowAgreementHook.sol
hellwolf: I concur. An easy step to lower the entropy with relatively little instability tenancy.

0xdavinchee: done

packages/ethereum-contracts/contracts/agreements/ConstantFlowAgreementV1.sol
```diff
@@ -447,6 +452,23 @@ contract ConstantFlowAgreementV1 is
         }
 
         _requireAvailableBalance(flowVars.token, flowVars.sender, currentContext);
+
+        // @note It is possible this silently fails due to out of gas reasons, and users should
+        // still be able to recreate the hook behavior. This logic should exist in the NFT contract though.
```
hellwolf, line 457: oh no :) now you used the N word! It's supposed to be a surprise!

0xdavinchee: 🐰🥚

hellwolf (approved):

PR #1105, opened 2022-10-03
Yellowpaper 1 - Denotational Semantics of General Payment Primitives, and Its Payment System

hellwolf:
# Abstract

Payment systems in information age are still largely modeled after their analog predecessors. While
electronic money payment systems do utilize computing technology and Internet, this paper presents a
case that a true modernization can be reached by (a) making payments happening continuously over
time, (b) involving more than two parties in a payment if necessary, (c) having compositional
financial contracts.

This paper first explores the foundation of modern payment system, which consists of a money
distribution model, payment primitives, payment execution systems of financial contracts and
different forms of money mediums. Then the paper uses \textit{denotational semantics} to formally
define payment primitives for modern payment system. Lastly, this paper also includes an overview
of \textit{Superfluid protocol}, a reference implementation of the payment primitives and its
payment system.

This paper includes literate Haskell code that compiles as the constructive artifact for the formal
specifications.

This paper is the first one of the series yellow papers about modern payment system dubbed
``semantic money''.


hellwolf:
will do one more proof-reading before sharing PDF.

0xdavinchee (approved):

PR #1108, opened 2022-10-06
Update of the nix flake tooling

hellwolf:
- Merge all dev shells in to root flake.nix
- Foundry nix still flakey, since it only has nightly builds.


hellwolf:
@d10r looking good to you?

d10r (approved):

PR #1109, opened 2022-10-06
Fix subgraph testing: copy upstream docker compose file

0xdavinchee:


hellwolf:
works :) ??!!

0xdavinchee:
Let's see

hellwolf:
nothing happened

hellwolf:
## Background

All our builds are failing, because thegraph is blocking our builds, since their latest build is broken.

Just noticed two things:

(1) Per https://github.com/actions/runner-images/blob/main/images/linux/Ubuntu2004-Readme.md

This  https://github.com/superfluid-finance/protocol-monorepo/blob/dev/.github/workflows/call.setup-deploy-and-test-local-subgraph.yml is not necessary:
      - name: "Run setup because linux and docker-compose"
        run: |
          sudo curl -L "https://github.com/docker/compose/releases/download/1.29.2/docker-compose-$(uname -s)-$(uname -m)" -o /usr/local/bin/docker-compose
          sudo chmod +x /usr/local/bin/docker-compose
          chmod +x setup.sh
          ./setup.sh
        working-directory: ./graph-node/docker

(2) The fix is simple, since it looks like what we need is just that docker-compose.yml file not the entire repo, we could have a copy as `
protocol-monorepo/packages/subgraph/test/graph-node-docker-compose.yml` (which is based on upstream, but pin the tag to a good build e.g. v0.28.0), and update the workflows:

$ grep 'docker-compose up' .github/ -R
.github/workflows/call.setup-deploy-and-test-local-subgraph.yml:        run: docker-compose up &
.github/workflows/call.test-subgraph-on-previous-sdk-core-versions.yml:        run: docker-compose up &


to docker-compose up -f path/to/packages/subgraph/test/graph-node-docker-compose.yml

0xdavinchee:
LFG!!!!!! 🥲

hellwolf (approved):
great stuff!

but let's leave a tech debt note in new issue that newer version of graph-node would break the test setup.

Issue #1112, opened 2022-10-09
[SDK-CORE] Use metadata for framework initialization


0xdavinchee:
## Overview

The `create` method in `Framework.ts` can use the metadata package to retrieve all the necessary contract addresses for initialization. 
We can also consider lazy loading of the `SuperToken` class. 

### Non-Breaking
- [ ] We keep `Famework.create` as returning a promise, but when a resolver address is passed as an option or something other than v1 is passed as the `protocolReleaseVersion` then we use the existing code path, else we use the metadata package.

### Breaking
- ~[ ] `Framework.create` no longer returns a promise and we initialize the contracts with the metadata addresses. If it is in a local environment, the metadata addresses will not exist for this and we will need to add a new function `initialize` which runs the existing framework creation code which loads the resolver.~  
  - **`Framework.create` must return a promise to check that the chainId from the provider is the same as the inputted chainId.**

PR #1113, opened 2022-10-10
[ETHEREUM-CONTRACTS/SDK-CORE/SUBGRAPH] Use more Typechain capabilities

0xdavinchee:
* Provide end-consumer access to typechain files via `ethereum-contracts` as the origin source and also exported via `sdk-core`
* no more abi file imports, using typechain generated factories all around
* fix up lint scripts
* add typechain build pipeline in ethereum-contracts


packages/ethereum-contracts/package.json
```diff
@@ -20,7 +20,8 @@
         "/build/contracts-sizes.txt",
```
surewht, line 20: Will we increase the version in another pr?

0xdavinchee: yeah, I think once we want to do a release it will make sense to - once we complete the cfav1 forwarder changes + maybe: https://github.com/superfluid-finance/protocol-monorepo/issues/1112

surewht (approved):
Awesome^^

0xdavinchee:
blocked by doc:html consuming the new exported types from ethereum-contracts and polluting the reference docs

packages/ethereum-contracts/CHANGELOG.md
```diff
@@ -5,6 +5,17 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 
 ## Unreleased
 
+### Added
+
+- CFA Hooks added (#1099)
```
kasparkallas, line 10: What's a "Hook"?

packages/sdk-core/package.json
```diff
@@ -57,6 +55,7 @@
         "url": "https://github.com/superfluid-finance/protocol-monorepo/issues"
     },
     "dependencies": {
+        "@superfluid-finance/ethereum-contracts": "1.4.1",
```
kasparkallas, line 58: I would think the published dev package will be broken if this is not incremented.

packages/sdk-core/src/ERC20Token.ts
```diff
@@ -1,22 +1,28 @@
...
 
-    constructor(address: string) {
+    constructor(
+        address: string,
+        providerOrSigner: ethers.providers.Provider | ethers.Signer
+    ) {
         this.address = address;
-        this.contract = new ethers.Contract(
+
+        this.contract = ERC20WithTokenInfo__factory.connect(
```
kasparkallas: Why is it good to connect the contract here? Aren't now all the contracts possibly connected to a signer? And when write operations are invoked then that signer is not used... unless there's an error in our code and we miss passing in the new signer in which case this signer is used.

kasparkallas (commented):
1. Please quickly host the TypeDoc somewhere so we could preview it.
2. I wish there was a simple example available for consumers to generate TypeChain themselves from `@superfluid-finance/ethereum-contracts` without the bash script ABI extraction overhead. I could imagine someone wanting to generate TypeChain for hardhat or web3 or something else that is not pre-generated by us.

packages/ethereum-contracts/CHANGELOG.md
0xdavinchee: check out the #1099 and also https://github.com/superfluid-finance/protocol-monorepo/issues/1096

packages/sdk-core/package.json
0xdavinchee: > 1. Please quickly host the TypeDoc somewhere so we could preview it.

will do this w/ netlify and post link when done

> 2. I wish there was a simple example available for consumers to generate TypeDoc themselves from `@superfluid-finance/ethereum-contracts` without the bash script ABI extraction overhead. I could imagine someone wanting to generate TypeDoc for hardhat or web3 or something else that is not pre-generated by us.

Are you referring to SDK-CORE here? If yes, it is fairly simple to do this now and the bash script ABI extraction has been deleted from the SDK-Core. I am pretty sure if you have the project locally, you just have to run `yarn build` in ethereum-contracts and then `yarn doc:html` in sdk-core. Then you just have to run `npx serve dist/docs` in sdk-core to have this locally

packages/sdk-core/src/ERC20Token.ts
0xdavinchee: > Why is it good to connect the contract here? Aren't now all the contracts possibly connected to a signer? 

yes, the contracts may be connected to a signer, but the proper use here is to connect a provider, so we can simply change the allowed type here if you think this is an issue.

> And when write operations are invoked then that signer is not used...

However, let's say we don't do that and we allow the signer and we throw away the signer later, there is no issue with this as you have to do this anyways if you are switching between signers in the same session-one signer at start vs. different signer later at execution is pretty similar to no signer at start then signer on first write operation and different signer on subsequent or each write operation.

> unless there's an error in our code and we miss passing in the new signer in which case this signer is used.

If the user of the product decides to execute transactions via the contract directly, this may be an issue and may warrant only allowing provider type to be passed on initialization. 
However, if they are using our class and creating an `Operation` and doing `.exec()`, this will absolutely fail even if you cast it as any because you would be trying to execute `.sendTransaction` on an undefined object.

packages/sdk-core/package.json
0xdavinchee: https://visionary-marzipan-1b34ea.netlify.app/

0xdavinchee: okay, we can increment the version in this PR

kasparkallas: Oops, for 2 I accidentally wrote "TypeDoc" but meant "TypeChain".

0xdavinchee: > Oops, for 2 I accidentally wrote "TypeDoc" but meant "TypeChain".

this never required a bash script ABI extraction to do so within `@superfluid-finance/ethereum-contracts`, you could simply run `npx hardhat compile` or `npx hardhat typechain` to get the typechain types for all contracts

packages/sdk-core/src/ERC20Token.ts
kasparkallas: What's the benefit of connecting here though? It looks like the connection is always overriden anyways.

0xdavinchee: Sure that's a fair point.

Firstly: Some extra typecasting is avoided with `.connect`.
Code with .connect:
```
this.contract = ERC20WithTokenInfo__factory.connect(
    this.address,
    providerOrSigner
);
```

Code without `.connect`:
```
this.contract = new ethers.Contract(
    this.address,
    ERC20WithTokenInfo__factory.abi
) as ERC20WithTokenInfo;
```

Secondly, for write operations, I think this will always be overridden and it is probably safe to do this. However, by providing a provider/signer in the initialization of the contract, it opens up the possibility of not requiring the end user to pass a provider on reads. This does become more problematic if the consumer is on multiple chains but this can be pretty easily solved with just passing in a new provider.

Lastly, why not, this operation is not blocking and will not slow anything down - I see no good reason to not use this (imo) cleaner way of initializing a contract.

kasparkallas (approved):

.github/workflows/call.check-query-schema-against-subgraph.yml
```diff
@@ -32,6 +32,7 @@ jobs:
         run: |
           yarn build:contracts-truffle
           yarn build:contracts-hardhat
+          yarn build:typechain
```
hellwolf: not a blocking comment, but having the same build:* sequence across several places tickles DRY principle. something to tackle later.

hellwolf (approved):
will unblock this since many people already reviewed.

hellwolf:
> The base branch requires all commits to be signed. [Learn more about signing commits.](https://docs.github.com/articles/about-gpg/)


One could do a signed force commit that squashed everything in one to unblock this.

.github/workflows/call.check-query-schema-against-subgraph.yml
0xdavinchee: will resolve this in the same PR along w/ squashed commit - just does yarn build instead

PR #1115, opened 2022-10-11
ISSUE #1089 - Create2 Hybrid SuperTokenFactory

### LINKED ISSUE #1089: [ETHEREUM-CONTRACTS] Deterministic (network-wise) Wrapped-ERC20 Super Token Address for Simpler Discoverability
Currently the Super Token factory can help anyone to issue new Wrapped ERC20 Super Tokens any time. It hinders the discoverability, for which Superfluid Finance's resolver mechanism is a sub-optimal patchwork for this limitation.

It needs not to be so, host contract could deem canonical wrapped ERC20 (in a sense that it uses super token logic  the host provides) something uniquely available per each network, deterministically generated, so that the discoverability problem goes away automatically.

## Technical notes

- A pre-built binary + CREATE2 could provide a decent technical solution.
  - Need to backward support resolver listed super tokens, and a lookup-table to cement "determined" super token address would be more practical way forward than CREAET2 (which in itself will be less useful) everything.
- Old mechanism should still be available, but new mechanism should be favored for listing criteria in superfluid finance's resolver.

0xdavinchee:
**DO NOT MERGE YET**

We must first decide what the hardcoded address will be in `SuperTokenFactory.sol` (preferably the same addresses for all networks to reduce deployment overhead).

- Create canonical SuperToken function
- Initialize canonical SuperTokens list function
- Tests which test both

See https://github.com/superfluid-finance/protocol-monorepo/issues/1089#issuecomment-1273503402


hellwolf:
@d10r how do we manage the ops side of the thing of it?

d10r:
looks good, but there's a few details I dislike:
1. `createCanonicalERC20Wrapper` should imo do only that: create wrappers. I think it should revert if already exists. Afaik create2 itself reverts if the contract already exists, would thus also simplify the code.
2. Add `getCanonicalERC20Wrapper` as this is probably the most important use case. Return zero (or revert?) of not yet deployed.
3. Rename `computeWrapperSuperTokenAddress` to `computeCanonicalERC20WrapperAddress` (better consistency of the names) and simplify the 2nd argument name to `isDeployed`.

0xdavinchee:
> * `createCanonicalERC20Wrapper` should imo do only that: create wrappers. I think it should revert if already exists. Afaik create2 itself reverts if the contract already exists, would thus also simplify the code.

From what I understand of this, for the first part: we don't want to return the SuperToken if it exists, but revert instead.
For the second part, I believe you are referring to the check we have there. We cannot just rely on CREATE2 to revert here if that's what you're referring to.

packages/ethereum-contracts/contracts/mocks/SuperTokenFactoryMock.sol
```diff
@@ -68,6 +68,27 @@ contract SuperTokenFactoryMock is SuperTokenFactoryBase
         return _helper.create(host, 0);
     }
 
+    function initializeCanonicalWrapperSuperTokens(
```
hellwolf: update validateStorageLayout!!!

packages/ethereum-contracts/contracts/superfluid/SuperTokenFactory.sol
```diff
@@ -156,6 +210,73 @@ abstract contract SuperTokenFactoryBase is
...
+    {
+        superTokenAddress = _canonicalWrapperSuperTokens[
+            _underlyingTokenAddress
+        ];
+    }
+
+    /// @dev This is virtual so that SuperTokenFactoryMock can test it
+    /// @dev Note that this should also be kind of a throwaway function which will be executed only once.
+    /// @inheritdoc ISuperTokenFactory
+    function initializeCanonicalWrapperSuperTokens(
```
hellwolf, line 282: open permission? why?

hellwolf: perhaps at least onlyGovernance so we are giving us a bit room for adjustment until we through this away.

it is not difficult to send custom transactions from our multisig gov

hellwolf (changes requested):
commented

packages/ethereum-contracts/contracts/superfluid/SuperTokenFactory.sol
```diff
@@ -87,6 +91,54 @@ abstract contract SuperTokenFactoryBase is
 
     function createSuperTokenLogic(ISuperfluid host) external virtual returns (address logic);
 
+    /// @inheritdoc ISuperTokenFactory
+    function createCanonicalERC20Wrapper(ERC20WithTokenInfo _underlyingToken)
+        external
+        returns (ISuperToken)
+    {
```
d10r, line 114: pls add a check if the list of canonical wrappers is initialized, if not revert here. Avoids a race condition with the initialization tx. Unlikely to happen, but still...

d10r:
> @d10r how do we manage the ops side of the thing of it?

preps and ops for initial deployment:
* manually compile a list of pre-existing wrappers we consider canonical (name and symbol matching, semi-upgradable) per network
* set the address of the gh agent (`0xd15D5d0f5b1b56A4daEF75CfE108Cb825E97d015`) as [initializer](https://github.com/superfluid-finance/protocol-monorepo/pull/1115/files#diff-8d32b78f9249c83c353fe27aa7e0ec26a75a097cbf530d1c7665e7a08cd6158cR268)
* create ad-hoc script for initializing after the upgrade.

update docs:
* update [the docs](https://docs.superfluid.finance/superfluid/developers/super-tokens/deployment#using-a-block-explorer-to-deploy-your-wrapper) to use the canonical wrapper deploy method
* update [the standalone dapp](https://github.com/superfluid-finance/protocol-monorepo/blob/dev/packages/ethereum-contracts/utils/supertoken-deployer.html) to create canonical wrappers

0xdavinchee:
@hellwolf @d10r will we be using governance or GH agent for this?

packages/ethereum-contracts/contracts/superfluid/SuperTokenFactory.sol
```diff
@@ -31,13 +27,19 @@ abstract contract SuperTokenFactoryBase is
 
     ISuperToken internal _superTokenLogic;
 
+    /// @notice A mapping from underlying token addresses to canonical wrapper super token addresses
+    /// @dev Reasoning: (1) provide backwards compatibility for existing listed wrapper super tokens
+    /// @dev (2) prevent address retrieval issues if we ever choose to modify the bytecode of the UUPSProxy contract
+    /// @dev NOTE: address(0) key points to the NativeAssetSuperToken on the network.
+    mapping(address => address) internal _canonicalWrapperSuperTokens;
```
hellwolf, line 44: Please add a comment nearby here so that each time the code impacts code storage layout, one does not forget to update:

https://github.com/superfluid-finance/protocol-monorepo/blob/697044bbe58a5b2b4fe1992d66913f8e7e51df22/packages/ethereum-contracts/contracts/mocks/SuperTokenFactoryMock.sol#L21

This is a MUST practice for doing upgradable proxy code.

hellwolf:
@d10r since it looks like we are not introducing breaking changes to  `Upgradability.SEMI_UPGRADABLE` behavior, what's our super token creation guideline going forward?

d10r:
> @d10r since it looks like we are not introducing breaking changes to `Upgradability.SEMI_UPGRADABLE` behavior, what's our super token creation guideline going forward?

imo for simple wrappers we should switch the recommendation to use `createCanonicalERC20Wrapper`, see https://github.com/superfluid-finance/protocol-monorepo/pull/1115#issuecomment-1289311527

packages/ethereum-contracts/contracts/interfaces/superfluid/ISuperTokenFactory.sol
```diff
@@ -76,6 +82,52 @@ interface ISuperTokenFactory {
...
+        returns (ISuperToken);
+
+    /**
+     * @notice Computes/Retrieves wrapper super token address given the underlying token address
+     * @dev We return from our canonical list if it already exists, otherwise we compute it
+     * @dev note that this function only computes addresses for SEMI_UPGRADABLE SuperTokens
+     * @param _underlyingToken Underlying ERC20 token address
+     * @return superTokenAddress Super token address
+     * @return isDeployed whether the super token is deployed AND set in our canonical mapping
+     */
```
d10r: imo the word _our_ doesn't fit here

packages/ethereum-contracts/contracts/interfaces/superfluid/ISuperTokenFactory.sol
```diff
@@ -76,6 +82,52 @@ interface ISuperTokenFactory {
...
+        returns (address superTokenAddress);
+
+    /**
+     * @notice Initializes list of canonical wrapper super tokens.
+     * @dev Can only be set once by a hardcoded address
+     * @param _data an array of canonical wrappper super tokens to be set
+     */
+    function initializeCanonicalWrapperSuperTokens(
+        InitializeData[] calldata _data
+    ) external;
```
d10r: I feel like we shouldn't add this to the interface (same for the `InitializeData` struct).
@hellwolf do you agree?

packages/ethereum-contracts/contracts/superfluid/SuperTokenFactory.sol
```diff
@@ -18,26 +19,44 @@ import { SuperToken } from "../superfluid/SuperToken.sol";
...
         _host = host;
     }
 
-    /// @dev ISuperTokenFactory.getHost implementation
+    modifier onlyGovernanceOwner() {
+        SuperfluidGovernanceII gov = SuperfluidGovernanceII(address(_host.getGovernance()));
+        address owner = gov.owner();
+        if (msg.sender != owner) revert SuperfluidGovernanceII.SF_GOV_II_ONLY_OWNER();
+        _;
+    }
```
d10r: We shouldn't make this dependent on `SuperfluidGovernanceII`, is not used e.g. on testnets and may change in the future.
We have a problem here: neither `ISuperfluidGovernance` nor `SuperfluidGovernanceBase` include `Ownable`. But all implementations we have in use do. @hellwolf Shall we establish that going forward all gov implementations have to implement `Ownable` and use `Ownable` here?

packages/ethereum-contracts/contracts/interfaces/superfluid/ISuperTokenFactory.sol
0xdavinchee: replaced with **the**

packages/ethereum-contracts/contracts/interfaces/superfluid/ISuperTokenFactory.sol
0xdavinchee: I guess since it's a throwaway only for us to call by creating our own transaction, this is probably not necessary here, going to remove it

packages/ethereum-contracts/contracts/superfluid/SuperTokenFactory.sol
hellwolf: agreed

hellwolf (approved):

packages/ethereum-contracts/contracts/interfaces/superfluid/ISuperTokenFactory.sol
```diff
@@ -14,6 +14,11 @@ import { SuperfluidErrors } from "./Definitions.sol";
  * @author Superfluid
  */
 interface ISuperTokenFactory {
+    struct InitializeData {
+        address underlyingToken;
+        address superToken;
+    }
+
```
d10r: remove that from the interface too

0xdavinchee: done, waiting on build

PR #1104, opened 2022-10-03
[ETHEREUM-CONTRACTS] SuperfluidFrameworkDeployer Additions

0xdavinchee:
## SuperfluidFrameworkDeployer.sol changes
* add deploy native asset super token
* add deploy pure super token
* add tests for token deployments
* modify .prettierrc and tsconfig.json

## SDK-Core Changes
* refactored the test harnessing (use TestEnvironment)

## Subgraph Changes
* **IMPORTANT** For the liquidation test, I delete the existing flow then recreate and run the liquidation test. For some reason, updating then liquidating leads to negative `balanceUntilUpdatedAt` which should never occur.
* new hardcoded addresses (sorry future me or dev who has to work on this)
* fixed up some expect values based on SuperfluidFrameworkDeployer logic (related to wrapper token naming)

> Note: Attempted to replace existing deployment scripts with a deployment script based off SuperfluidFrameworkDeployer, but this lead to far more headaches due to issues w/ library linking. This works out of the box with foundry, but not in JS environments.

PR #1117, opened 2022-10-12
CFA forwarder deployment related changes

d10r:
* added script for deploying CFA forwarder (incl. creating gov action and resolver setting)
* deploy and register CFA forwarder for dev/test deployments
* removed `solidity-coverage` from truffe config (blocked verification, seems unused / migrated to hardhat)
* updated truffle-plugin-verify, now supports optimism-goerli and arbitrum-goerli


packages/ethereum-contracts/truffle-config.js
```diff
@@ -129,7 +129,6 @@ function createNetworkDefaultConfiguration(
 const E = (module.exports = {
     plugins: [
         //"truffle-security",
-        "solidity-coverage",
```
d10r, line 132: I removed this because the plugin makes verification fail:
```
$ npx truffle run --network eth-goerli verify ConstantFlowAgreementV1@0x7F3Ba5Ded3C017c5bA0469085AAf0Ba5F7cA42b4
ExtendableError: 
Error: truffle-plugin.json not found in the solidity-coverage plugin package!
```

This error disappears if not enabling the plugin.

As far as I understand, coverage checking is now handled via hardhat and thus can be removed here.
@hellwolf is that correct?
If so, the "coverage" network should likely be removed too from this file.

0xdavinchee: can confirm that coverage is handled by HH:
`"test-coverage": "yarn run-hardhat coverage --testfiles testsuites/all-contracts.js --solcoverjs ./.solcover.js"`

packages/ethereum-contracts/scripts/deploy-framework.js
```diff
@@ -451,17 +451,24 @@ module.exports = eval(`(${S.toString()})({skipArgv: true})`)(async function (
...
-        CFAv1Forwarder,
-        "CFAv1Forwarder",
-        async (contractAddress) => contractAddress === ZERO_ADDRESS,
-        async () => {
-            const forwarder = await CFAv1Forwarder.new(superfluid.address);
-            output += `CFA_V1_FORWARDER=${forwarder.address}\n`;
-            return forwarder;
-        }
-    );
+    // deploy CFAv1Forwarder for test deployments
```
hellwolf, line 454: hmm, I am a bit confused by namings then, and I am a bit of out dated now on what assorted deployment scripts we have and what are they for.

what's your thinking behind it currently? @d10r

CODEOWNERS
```diff
@@ -12,6 +12,8 @@
 #
 /packages/spec-haskell/ @hellwolf
 /packages/ethereum-contracts/ @hellwolf @0xdavinchee
+/packages/ethereum-contracts/contracts @hellwolf @0xdavinchee
```
hellwolf: piggybacking CODEOWNERS updates here.

It occurs to me @d10r should gatekeeping the changes to scripts instead.

packages/ethereum-contracts/tasks/deploy-cfa-forwarder.sh
```diff
@@ -0,0 +1,41 @@
...
+# For optimism the gas estimation doesn't work, requires setting EST_TX_COST.
+# For polygon-mainnet, GAS_PRICE usually needs to be set.
+#
+# On some networks you may need to use override ENV vars for the deploymentto succeed
+
+network=$1
+cfaAddr=$2
+
+# deploy
+#npx truffle exec --network $network scripts/deploy-deterministically.js : CFAv1Forwarder
```
0xdavinchee: this is commented out, just want to double check that this is expected? if it is, then this is good to go 👍🏼

packages/ethereum-contracts/scripts/deploy-framework.js
hellwolf: * deploy-framework.js
* deploy-test-environment.js 
* deploy-test-framework.js

hellwolf: still the same as before actually.. but maybe it's time for us to actually clarify it more (also taking into account backward compatibility if needed, or abandon it if unhelpful) since I have forgotten many things since then.

0xdavinchee: Looks like `deploy-framework` is called in `deploy-test-environment`. 

From doing a quick search, we use `deploy-test-environment.js` for deployments to testnets and `deploy-framework.js` for mainnet as well as our contracts test suite (TestEnvironment.ts).

The `deploy-test-framework.ts` is used for deploying SF in SDK-Core/Subgraph using the `SuperfluidFrameworkDeployer`: https://github.com/superfluid-finance/protocol-monorepo/pull/1104

packages/ethereum-contracts/tasks/deploy-cfa-forwarder.sh
d10r: this was indeed not intended, thx for spotting

CODEOWNERS
d10r: I'd prefer gatekeeping the contracts tbh (contracts dir only, excluding tests), I feel like we're being too liberal with changes there.
Regarding scripts, there's those for ops (gov & other) which I feel qualified to gatekeep, and those mainly used in tests (inside and outside monorepo). I don't think I'm best qualified to gatekeep the latter.
Probably we should try to better categorize the scripts (into different dirs?) along that line.
Maybe related: I've in the past wondered what should go into `tasks` and what should go into `scripts`.

packages/ethereum-contracts/scripts/deploy-framework.js
d10r: This has left me a bit head-scratching too - the forwarder is kind of a special case, bcs not really part of the protocol, but plugged in via forwarder mechanism.

My understanding is:
* `deploy-framework.js` wants to be the protocol deploy script for all: mainnets (initial & upgrade), testnets (initial & upgrade), devnets.  
* `deploy-test-environment.js` is for testnets (initial & upgrade= and devnets, it additionally (to the protocol) deploys (or upgrades to latest logic) test tokens.
* `deploy-test-framework.ts` uses - as @0xdavinchee already mentioned - `SuperfluidFrameworkDeployer`, which isn't as self-sufficient as I believed it is. It seems to have no truffle dependency (?), thus may be a useful basis for hardhat projects. I just had that topic with Sam, discussing what the canonical way of deploying for tests in our example projects should be.

0xdavinchee: Will mark this as resolved for now and create a new issue for classifying/cleaning up these scripts.

0xdavinchee (approved):
LGTM

CODEOWNERS
hellwolf: Everyone should still review the code diligently, it's up to the codeowner to facilitate that.

hellwolf (approved):

PR #1092, opened 2022-09-26
Draft: integrated CFAv1Forwarder into the sdk

surewht:

PR #1118, opened 2022-10-12
[ETHEREUM-CONTRACTS/SDK-CORE] CFAv1 Forwarder Integration

### LINKED ISSUE #1090: [SDK-CORE] Integrate Agreement Forwarder into SDK-Core
## Summary
We recently created the `CFAv1Forwarder.sol` contract which provides an easy to use interface for `ConstantFlowAgreementV1.sol`. In addition to a nicer API for devs to interact with at the contract level when writing code, this also provides a better experience when debugging via Etherscan as you will be able to see the inputs passed instead of just `callAgreement` being called with a bunch of bytecode.

## TODOs
- [x] Decide whether we want to provide the option for users to still use the old `callAgreement` semantic.
- ~[ ] Add `CFAv1Forwarder` to `contracts.json` in SDK-Core~ contracts.json is deprecated
- [x] Integrate the use of the Agreement Forwarder in SDK-Core for the `ConstantFlowAgreementV1` class
- [x] Provide the Agreement Forwarder contract address in `Framework.settings.config` and provide the contract in `Framework.contracts`
- [x] Tests that assert that we are actually using the new `CFAv1Forwarder` contract 
- [x] Update CHANGELOG.md
- [x] Notify DevX team for possible updates to documentation

## Some Possible Gotchas
- [x] You will need to deploy the `CFAv1Forwarder` contract, which means you'll need to modify some of the scripts in `packages/ethereum-contracts`
- ~[ ] Ideally the address should be deterministic so it can be hardcoded, but it may still be safer to add it to the `Resolver` contract during deployment (`packages/ethereum-contracts/scripts/deploy-deterministically.js`): the resolver contract just stores a set of key/value which allows us to look up addresses based on different keys. For example, listed (whitelisted on the resolver) supertokens have the key: `supertokens.${protocolReleaseVersion}.${tokenSymbol}x`~

0xdavinchee:
@tokdaniel did some of this in https://github.com/superfluid-finance/protocol-monorepo/pull/1092 - I just copied it over
* Breaking: `sender` is **required** in create/update/deleteFlow
* Split SDK-Core SuperToken test files into separate files: CFA, CFA_Operator, IDA
* Add CFAv1Forwarder to SuperfluidFrameworkDeployer + set in resolver and enable as trusted forwarder
* Integrate CFAv1Forwarder in SDK-Core
* tests to ensure that forwarder is being used and works as expected
* validateOperationShouldUseCallAgreement checks that we are using agreement forwarder
* Finally did a proper fix for the hardcoded subgraph addresses, there may be further improvements that can be made.
* NOTE: To ensure BatchCall still works out of the box with no changes - Operation class takes `forwarderPopulatedTxn`, the other solution would've been to write logic (which runs during batch call creation/execution) for each of the call agreement functions which take the args out of the forwarder txn and convert it into a call agreement txn
  * We went ahead with this method of adding an optional property on the Operation class given the design/architecture of the SDK-Core and the Operation class: easier maintenance and less likely to mess up the logic at the expense of a possibly heavier Operation class object

This PR is blocking https://github.com/superfluid-finance/protocol-monorepo/pull/1119


surewht (approved):
lgtm^^

hellwolf (approved):

packages/sdk-core/src/Operation.ts
```diff
@@ -18,12 +18,18 @@ export default class Operation {
     readonly populateTransactionPromise: Promise<ethers.PopulatedTransaction>;
     readonly type: OperationType;
 
+    // @note This property is used to ensure BatchCall operations still function
+    // when using the agreement forwarder
+    readonly forwarderPopulatedPromise?: Promise<ethers.PopulatedTransaction>;
```
kasparkallas, line 23: As we discussed, ideally a factory function (`() => ...`) would be slightly better for `forwarderPopulatedPromise` and `populateTransactionPromise` to avoid always evaluating both of the promises.

0xdavinchee: agreed, we can iterate on this in the future, just wanted to push some of this stuff out ASAP so @tokdaniel could demo his stuff today

PR #1127, opened 2022-10-17
[SDK-CORE] Load w/ metadata

### LINKED ISSUE #1112: [SDK-CORE] Use metadata for framework initialization

0xdavinchee:
* Initialize Framework class w/ metadata for supported networks
* Continue using resolver/superfluidLoader for non-supported networks


kasparkallas (commented):
Changelog?

packages/sdk-core/README.md
```diff
@@ -50,6 +50,13 @@ SDK-Core is in early active development and can have breaking releases without w
 
 # Prerequisites
 
+> NOTE: You need to have graphql installed as a dependency in order to use SDK-Core:
```
kasparkallas: Also `ethers`, I guess...

0xdavinchee: ah true

packages/sdk-core/src/constants.ts
```diff
@@ -32,29 +31,41 @@ const subgraphReleaseTag =
...
+const getAddressesData = (chainId: number): NetworkData => {
     const networkData = metadata.networks.find((x) => x.chainId === chainId);
-    if (!networkData)
-        return {
-            subgraphAPIEndpoint: "",
-            networkName: "",
-            resolverAddress: "",
-            nativeTokenSymbol: "",
-        };
+    if (!networkData) return EMPTY_NETWORK_DATA;
```
kasparkallas: Why not just return `null` or `undefined`? It's better to be explicit than have empty strings and weird empty objects be passed around...

packages/sdk-core/README.md
```diff
@@ -50,6 +50,13 @@ SDK-Core is in early active development and can have breaking releases without w
 
 # Prerequisites
 
+> NOTE: You need to have graphql and ethers installed as a dependency in order to use SDK-Core:
+
+```bash
+yarn install --D graphql ethers
```
kasparkallas: I don't think `--D` is necessary here...

kasparkallas (changes requested):

packages/sdk-core/README.md
0xdavinchee: si

packages/sdk-core/src/constants.ts
0xdavinchee: hmmm touché

0xdavinchee:
@kasparkallas pls re-review, requested changes have been applied

packages/sdk-core/src/constants.ts
```diff
@@ -32,29 +31,41 @@ const subgraphReleaseTag =
     (globalThis.process && globalThis.process.env.SUBGRAPH_RELEASE_TAG) ||
     DefaultSubgraphReleaseTag.value;
 
-const getResolverData = (chainId: number): IResolverData => {
+export const EMPTY_NETWORK_DATA = {
```
kasparkallas: Not necessary anymore?

0xdavinchee: still used in `Framework.ts`, can change the logic in there slightly to disappear this object.

kasparkallas: Please do :wink:

0xdavinchee: donezo

kasparkallas (approved):

PR #1128, opened 2022-10-18
[SDK-CORE] Gas Multiplier

0xdavinchee:
* Provide the capability for users to multiply the estimated gas limit (done by the passed provider to SDK-Core) by an arbitrary amount
* Default multiplier is 1.2x (number is based off Truffle)


packages/sdk-core/src/Operation.ts
```diff
@@ -36,13 +36,16 @@ export default class Operation {
      * Executes the operation via the provided signer.
      * @description Populates all fields of the transaction, signs it and sends it to the network.
      * @param signer The signer of the transaction
+     * @param gasLimitMultiplier A multiplier to provide gasLimit buffer on top of the estimated gas limit (1.2x is the default)
      * @returns {ethers.providers.TransactionResponse} A TransactionResponse object which can be awaited
      */
     exec = async (
-        signer: ethers.Signer
+        signer: ethers.Signer,
+        gasLimitMultiplier = 1.2
```
kasparkallas, line 44: From looking at about 10 transactions with no multiplier, the consumed gas on Etherscan never went over 90%. So I guess now it will start to hover around 70%-75%.

packages/sdk-core/src/Operation.ts
```diff
@@ -54,12 +57,27 @@ export default class Operation {
...
             ? await this.forwarderPopulatedPromise
             : await this.populateTransactionPromise;
-        return await signer.populateTransaction(txnToPopulate);
+        const signerPopulatedTransaction = await signer.populateTransaction(
+            txnToPopulate
+        );
+
+        return {
+            ...signerPopulatedTransaction,
+            gasLimit:
```
kasparkallas: So if I explicitly passed in a gas limit with `Overrides` then this thing will multiply it?

0xdavinchee: yeah, prolly shouldn't be doing this given the name `Overrides`.

packages/sdk-core/src/Operation.ts
0xdavinchee: This 1.2 is a buffer to account for unpredictability. In probably 99.9% of cases, this should result in gas being refunded at the end of the txn. Also pls re-review changes here.

packages/sdk-core/test/2_operation.test.ts
```diff
@@ -116,10 +135,50 @@ makeSuite("Operation Tests", (testEnv: TestEnvironment) => {
...
+            );
+            await operation.exec(testEnv.alice);
+            const updateOpTxn1 = await updateOp1.exec(testEnv.alice, 1);
+            const updateOpTxn2 = await updateOp2.exec(testEnv.alice, 2);
+            expect(updateOpTxn1.gasLimit.mul(toBN(2))).to.equal(
+                updateOpTxn2.gasLimit
+            );
+        });
+
+        it("Should not apply multiplier to Overrides gas limit", async () => {
```
kasparkallas, line 170: :+1:

kasparkallas (approved):

kasparkallas:
Changelog?

PR #1133, opened 2022-10-19
[CI] Changelog reminder handler

0xdavinchee:
It might be helpful to use some automation here to remind us to update the changelog when making PRs (idea from @kasparkallas originally) to packages which have a `CHANGELOG.md`. 

I personally forgot to do this from time to time and this is something which is not super important to practice memory retrieval for so this seems like a nice to have feature.

* changlog reminder handler
* cleanup of check-changeset.sh


hellwolf (approved):
mkay bob

Issue #1134, opened 2022-10-24
[ETHEREUM-CONTRACTS] Scripts/Tasks


0xdavinchee:
## Summary
We would like to consolidate our deployment scripts and have a better understanding of what the purpose of each script is (categorizing it possibly) and what tools we would like to use/support for the different scripts. 

Some suggestions (from others and myself):
- Separate scripts for internal use (testing, ops, gov) and external use (testing/test deployment) in terms of folder structure and what we build and release with the contracts
- Providing extensive documentation in our developer docs and maybe even some tutorials on how to use the tool in different environments
- Investigating existing tools which we can leverage to make this process nicer
- Keeping a pulse on new tools which we can leverage (e.g. [Hardhat Ignition](https://medium.com/nomic-foundation-blog/hardhat-ignition-5a34a4e3d2de))

TODOS
- [ ] Remove/consolidate any scripts
- [ ] Add any new scripts which we might not currently have

0xdavinchee:
the separation of JS scripts into `dev-scripts` and `ops-scripts` has substantially simplified the external usage of deployment scripts and largely has solved this issue. We even have "migration" scripts in `scripts`, instructing users how to migrate.
in addition, we are moving towards supporting foundry more and should start urging developers to use our `SuperfluidFrameworkDeployer` and `FoundrySuperfluidTester` contracts for testing as Foundry is gaining more adoption

PR #1137, opened 2022-10-24
Create new script for deploying the SuperfluidFrameworkDeployer

saflamini:
Intended for usage within test suites

1) Users will import script and deploy framework deployer 

2) From here, they'll be able to call getFramework() and deploy tokens

cc @0xdavinchee

### DIFF
--- a/packages/ethereum-contracts/scripts/deploy-superfluid-deployer.js
+++ b/packages/ethereum-contracts/scripts/deploy-superfluid-deployer.js
@@ -1,0 +1,72 @@
+const {ethers } = require("hardhat");
+
+const SlotsBitmapLibraryArtifact = require("@superfluid-finance/ethereum-contracts/build/contracts/SlotsBitmapLibrary.json");
+const SuperfluidFrameworkDeployerArtifact = require("@superfluid-finance/ethereum-contracts/build/contracts/SuperfluidFrameworkDeployer.json");
+
+const ERC1820Registry = require("@superfluid-finance/ethereum-contracts/scripts/artifacts/ERC1820Registry.json");
+
+const ERC1820_ADDRESS = "0x1820a4b7618bde71dce8cdc73aab6c95905fad24";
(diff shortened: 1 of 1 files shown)


0xdavinchee (changes requested):
gm - looks good overall, but a few thoughts:
1. Can you create a branch off superfluid-finance? It's breaking one of the CI handlers and I have no way of fixing it cause it's a forked repo - the main reason to create stuff from here is so that we can make small changes without having to wait on each other.
2. I still feel a bit iffy about this because it is duplication which means maintenance down the line, although the maintenance overhead is likely to be fairly trivial here. The ideal solution would be to build/modify the existing `deploy-test-environment.ts` file or scrap that one and replace it with this one.
3. Lastly, I am not 100% sure if this solves the issue for TypeScript first projects where it will still complain because a .d.ts file doesn't exist. Not a major blocker, but ideally it works in all environments.

I think we should have another call, maybe also with @d10r to discuss our strategy for consolidating, purging and packaging the [different scripts](https://github.com/superfluid-finance/protocol-monorepo/issues/1134) to the end user.
We will likely need some extra build steps and we should be explicit in what we plan on exposing to end users.

saflamini:
> Can you create a branch off superfluid-finance? It's breaking one of the CI handlers and I have no way of fixing it cause it's a forked repo - the main reason to create stuff from here is so that we can make small changes without having to wait on each other.

Yes, will do.

> 3. Lastly, I am not 100% sure if this solves the issue for TypeScript first projects where it will still complain because a .d.ts file doesn't exist. Not a major blocker, but ideally it works in all environments.

Hmm - I wonder if we can just fix this by adding a type declaration file to the scripts folder and ship it with the NPM package?

> I think we should have another call, maybe also with @d10r to discuss our strategy for consolidating, purging and packaging the [different scripts](https://github.com/superfluid-finance/protocol-monorepo/issues/1134) to the end user. We will likely need some extra build steps and we should be explicit in what we plan on exposing to end users.

@0xdavinchee agree on the call with @d10r

PR #1142, opened 2022-10-26
[ETHEREUM-CONTRACTS] Localize Custom Errors

0xdavinchee:
* Remove SuperfluidErrors Library from Definitions.sol
* Localize custom errors to individual contracts
* Modify tests accordingly
* Retain the name so that it is easy to debug which contract is throwing the error
* No more use of error codes as this is redundant given the above
* Comments with the hashed custom error name for quick debugging


d10r (approved):
lgtm

hellwolf (approved):

PR #1145, opened 2022-10-28
[SDK-Core] Map `deposit` from Subgraph to `Stream`

kasparkallas:
Why?
The data could be used in Dashboard's stream detail view.


0xdavinchee (approved):
Thought I approved this last week, my bad!

PR #1161, opened 2022-11-15
[SDK-redux] Don't `skipLibCheck`

kasparkallas:
When building a SDK, probably best to check all types are correct across all the libs and peerDependencies and not let any type problems sneak in, even if they're issues in the dependencies.


0xdavinchee (approved):
LGTM

PR #1165, opened 2022-11-15
[SDK-redux] Add `@types/react` devDependency

kasparkallas:
Still battling CI errors that I can't reproduce locally.


d10r:
I get the same error on my local machine.
May it be about case sensitive filesystem?

kasparkallas:
> I get the same error on my local machine. May it be about case sensitive filesystem?

That's great! Let's talk on Slack.

kasparkallas:
Solved it in: https://github.com/superfluid-finance/protocol-monorepo/pull/1168

Issue #1151, opened 2022-11-03
[ETHEREUM-CONTRACTS] Brittle build process in non-virgin environment - need better "clean" script?


d10r:
My starting point: run the TOGA test, because I want to modify the contract.  
This is taking place in my main copy of the monorepo, not a fresh clone.

Attempt 1:
```
yarn dev
$ yarn dev
yarn run v1.22.19
$ tasks/dev.sh
$ nodemon -e sol,js,ts -x 'yarn test:contracts:hardhat 2>&1'
[nodemon] 2.0.19
[nodemon] to restart at any time, enter `rs`
[nodemon] watching path(s): *.*
[nodemon] watching extensions: sol,js,ts
[nodemon] starting `yarn test:contracts:hardhat 2>&1`
$ yarn run-hardhat test testsuites/all-contracts.ts
$ IS_HARDHAT=true hardhat test testsuites/all-contracts.ts
An unexpected error occurred:

Error: Cannot find module '../typechain-types'
```

Ok, I know. I manually deleted `typechain-types` because it made a `yarn build` fail before.  
I expected it to be rebuilt by `yarn build`, but that seems to not be the case:

```
Compiling your contracts...
===========================
> Everything is up to date, there is nothing to compile.
$ yarn run-hardhat compile
$ IS_HARDHAT=true hardhat compile
Nothing to compile
No need to generate any newer typings.
```

So, Hardhat seems to think there's nothing to be done, likely because artifacts are up to date.  
For truffle stuff, `yarn build` does preliminary cleanup, but not for hardhat stuff. See npm scripts:
```
        "build:contracts": "rm -f build/contracts/*;yarn build:contracts-truffle",
        "build:contracts-truffle": "yarn run-truffle compile",
        "build:contracts-hardhat": "yarn run-hardhat compile",
```
This is confusing. "build:contracts" is neither a generalized script (building for both truffle and hardhat) nor just an alias for truffle (one has preliminary cleanup, the other doesn't). Maybe we want to keep "build:contracts" as alias (legacy reasons?), but cleanup policy should be consistent.

What's also not being cleaned up: *d.ts files in `scripts` directory. Can also make a re-run of `yarn build` fail.  
If the expectation is for `yarn build` to work on non-virgin checkouts too, we probably need to update the root `clean` script to include everything needed, especially everything generated by the typescript build steps (typechain-types, *.d.ts files, ?)

[...]

PR #1167, opened 2022-11-15
eth-mainnet config & npm script changes

d10r:
@0xdavinchee pls check:
in order to address https://github.com/superfluid-finance/protocol-monorepo/issues/1151, I slightly modified npm scripts.  
* [this change](https://github.com/superfluid-finance/protocol-monorepo/compare/eth-mainnet?expand=1#diff-57747701108d67c6d11910275a322f548ecd6fb9322170b7c45ad10c4e141587L37) avoids duplicate execution of truffle compilation for the wildcard script target `build:*`. I didn't add `rm -rf build/contracts/*` to `build:contracts-truffle` because not seeing a reason why we want that for truffle, but not for hardhat.
* [this change](https://github.com/superfluid-finance/protocol-monorepo/compare/eth-mainnet?expand=1#diff-57747701108d67c6d11910275a322f548ecd6fb9322170b7c45ad10c4e141587R41) avoids failure on repeated execution of `yarn build:scripts` for me. I'm however not sure if this is a good solution. There's still *d.ts and *.d.ts.map files left in `scripts/libs`, but those seem to not cause errors.
I noticed that `tsconfig.typechain` extends `tsconfig` while `tsconfig.scripts` doesn't.
The npm script `build:typechain` doesn't fail on repeated invocation despite no files being deleted. Should we maybe have `tsconfig.scripts` extend `tsconfig` too for similar behaviour?


packages/ethereum-contracts/package.json
```diff
@@ -29,16 +29,16 @@
...
         "run-truffle": "IS_TRUFFLE=true truffle",
         "run-forge": "IS_FOUNDRY=true forge",
         "run-nodemon": "nodemon -e sol,js,ts -x",
         "build": "run-s build:*",
-        "build:contracts": "rm -f build/contracts/*;yarn build:contracts-truffle",
         "build:contracts-truffle": "yarn run-truffle compile",
         "build:contracts-hardhat": "yarn run-hardhat compile",
         "build:typechain": "tsc -p tsconfig.typechain.json",
-        "build:scripts": "tsc -p tsconfig.scripts.json",
+        "build:scripts": "rm -rf scripts/*.d.ts scripts/*.d.ts.map; tsc -p tsconfig.scripts.json",
```
0xdavinchee, line 41: my only concern with this is that it will break when trying to remove the .d.ts file and .d.ts.map files if they don't exist, I have a solution I worked on locally, but removed, I will add it tomorrow, but it involves writing a bash script which checks existence of the .d.ts in scripts and does the remove based on that - I think CI is broken rn cause of this

kasparkallas (approved):

0xdavinchee (approved):
LGTM

hellwolf (approved):

PR #1168, opened 2022-11-16
[SDK-redux] Add missing devDependencies vol. 2

kasparkallas:
:crossed_fingers:

### DIFF
--- a/packages/sdk-redux/package.json
+++ b/packages/sdk-redux/package.json
@@ -42,10 +42,12 @@
         "promise-retry": "^2.0.1"
     },
     "devDependencies": {
-        "react": "^18",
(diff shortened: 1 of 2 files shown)


kasparkallas:
Okay, it was about `skipLibCheck=false` and missing devDependencies after all.

The problem got a bit obfuscated because: 
1. I couldn't reproduce locally even with a clean repo
2. I couldn't do `yarn add -D *package*`  because I was getting another unrelated error

Regarding 2, deleting yarn.lock and installing again fixed it, but I don't want to do it here. Talked briefly about this with @0xdavinchee 

The error I was getting looked like this:
image

0xdavinchee (approved):
LGTM

PR #1170, opened 2022-11-16
[DIDI's MUSE] CFAv1Forwarder v2 (WIP)

d10r:
Iterating on the API based on https://www.notion.so/superfluidhq/CFAv1Forwarder-Official-Draft-6886ffd0989747d5af78770c4e9c42fb

First attempt:
Deployed to Avalanche Fuji (faster and less crowded than Görli): https://testnet.snowtrace.io/address/0x480Fe426d2a7b2BC3cCee8EaF14E3Caa79281dD1#readContract
not yet tested, but should work.

This version uses overloading for `setFlow` and `setFlowFrom`.
Other methods for setting flowrate (create, update, delete, applyDelta) have only one method each with all potentially needed params. The tradeoff here is to make the less often needed (?) methods a bit more verbose in order to keep the overall API leaner.
For delta there's also just 1 method instead of 2 (applyDelta instead of increase/decrease).
We may also consider always requiring the `sender`, as e.g. the transfer methods of EIP-721 do.
Not sure if `setFlow` or `setFlowFrom` would be better in that case, I tend towards the former.


hellwolf:
I guess we will continue this with the change from https://github.com/superfluid-finance/protocol-monorepo/pull/1219?

Issue #1106, opened 2022-10-05
[SUBGRAPH] Rewrite Subgraph Tests as Unit Tests w/ Matchstick OR Huge Refactor of Existing Tests


0xdavinchee:
> Note: This is not a priority now, but I just wanted to note this as this has been something that has been bothering me for a bit and triggered by a recent experience with jumping back into the tests.

## Summary
The current Subgraph tests what happens when we actually trigger contract events in a local blockchain with a local subgraph and sees how the Subgraph handles the indexing over time with the same state.

We essentially do this:
- Initialize empty state
- Modify state with known parameters
- Locally modify state in a way we expect the subgraph to modify the state => this is done by essentially duplicating the mapping logic, but in the tests (pretty much no way of getting around duplicating the subgraph mapping logic here)
- Wait for the Subgraph to index the state modification transaction 
- Query the subgraph for the newly indexed 

I am proposing this because I think that this test suite is overkill and the maintenance costs are far too high to justify not trying to do this in a different way which may provide just as much value in a cleaner way. I don't think we have to or should throw these tests away completely, but it might be a worthy investment to try out [matchstick](https://github.com/LimeChain/matchstick) and incrementally add unit tests there to see if makes sense to make a full transition.

As the creator of these tests I have had to go back to work on these tests a few times and it is not easy to follow due to a few reasons:
- how I architected the tests
- the amount of state stored and logic replicated from the subgraph code
- Unrelated, but other people who have worked on it (Marten and Mehul) have also said it's pretty hard to follow

I think we should strongly consider either refactoring this greatly to make it more maintainable by either reducing the scope OR incrementally rewriting some of the tests with a different framework.

0xdavinchee:
TODOS

- [x] Figure out how to matchstick
- [x] Come up with a plan of how what the unit tests will consist of: test the different mapping functions/creation of entities
  - [x] This should be as simple as understanding what events create what entities and ensuring that this is done as expected
- [ ] Come up with a plan of how what the integration tests will consist of: test that the subgraph is properly indexing and creating an event
  - [ ] Delete the state persistence stuff and just check that all the entities are created when a specific event is emitted and retrievable
- [ ] Figure out how to organize the tests in a way so that it is most maintainable 
- [ ] Write the unit tests
- [ ] Purge a bunch of the integration test code
- [ ] Handle the CI jazz
- [ ] Profit for future me/other devs so they don't hate me as much I hate myself when working on tests here 🥲

[...]

PR #1172, opened 2022-11-17
[SUBGRAPH] Subgraph Matchstick Unit Tests

0xdavinchee:
tl;dr
- Mapping logic changes are in `src/mappings/*.ts` and `src/*.ts` files
- This PR is a checkpoint so it's more reviewable
- Using Matchstick to run unit tests for mapping logic
- A good chunk of event entities implemented + validated
- Some HOL and aggregate entities

## Overview

This is just an initial PR for the Subgraph Matchstick Unit Tests, you can check [here](https://github.com/superfluid-finance/protocol-monorepo/issues/1106) to see what has been covered so far. I wanted to submit this PR as a checkpoint.

## In this PR
- Structure of tests is set
- Testing a good % of event entities 
- Some HOL and aggregate entities tested
- Lots of helper functions
- Tests added to test subgraph call handler yml
- README updated
- `schema.graphql` subgraph documentation updated
- Refactor of `let` to `const`


packages/subgraph/src/mappingHelpers.ts
```diff
@@ -101,6 +103,8 @@ export function getOrInitSuperToken(
         token.createdAtTimestamp = currentTimestamp;
         token.createdAtBlockNumber = block.number;
         token.isSuperToken = true;
+        token.name = "";
```
0xdavinchee, line 106: logic change alert: set token name and symbol to empty string here

packages/subgraph/src/mappingHelpers.ts
```diff
@@ -116,21 +120,26 @@ export function getOrInitSuperToken(
...
         let tokenStatistic = getOrInitTokenStatistic(tokenAddress, block);
-        tokenStatistic = updateTotalSupplyForNativeSuperToken(
-            token,
+        tokenStatistic = getInitialTotalSupplyForSuperToken(
             tokenStatistic,
             tokenAddress
         );
         tokenStatistic.save();
 
+        // Per our TokenStatistic Invariant: whenever we create TokenStatistic, we must create TokenStatisticLog
```
0xdavinchee, line 129: logic change alert: create token statistic entity

packages/subgraph/src/mappingHelpers.ts
```diff
@@ -80,17 +80,19 @@ export function getOrInitAccount(
  * We also create token stats in here if it doesn't exist yet.
  */
 export function getOrInitSuperToken(
+    event: ethereum.Event,
```
0xdavinchee, line 83: logic change alert: passing event and triggeredByEventName - both used in token statistic log creation

packages/subgraph/src/mappingHelpers.ts
```diff
@@ -152,20 +161,25 @@ export function getOrInitSuperToken(
...
     token.createdAtTimestamp = block.timestamp;
+    token.decimals = 0;
+    token.name = "";
+    token.symbol = "";
     token.createdAtBlockNumber = block.number;
     token.isSuperToken = false;
+    token.isNativeAssetSuperToken = false;
     token.isListed = false;
-    token = getTokenInfoAndReturn(token as Token);
+    token = handleTokenRPCCalls(token, resolverAddress);
```
0xdavinchee, line 182: logic change alert: handleTokenRPCCalls to also check if underlying token is listed

packages/subgraph/src/mappingHelpers.ts
```diff
@@ -152,20 +161,25 @@ export function getOrInitSuperToken(
...
-    let tokenId = tokenAddress.toHex();
+    const tokenId = tokenAddress.toHex();
     let token = new Token(tokenId);
 
     if (tokenAddress.equals(ZERO_ADDRESS)) {
         return;
     }
 
     token.createdAtTimestamp = block.timestamp;
+    token.decimals = 0;
```
0xdavinchee, line 175: logic change alert: default decimals, name, symbol

packages/subgraph/src/mappingHelpers.ts
```diff
@@ -152,20 +161,25 @@ export function getOrInitSuperToken(
...
+    token.name = "";
+    token.symbol = "";
     token.createdAtBlockNumber = block.number;
     token.isSuperToken = false;
+    token.isNativeAssetSuperToken = false;
```
0xdavinchee, line 180: logic change alert: isNativeAssetSuperToken set to false for underlying token

packages/subgraph/src/mappings/resolver.ts
```diff
@@ -67,6 +67,8 @@ export function handleSet(event: Set): void {
             token.save();
         }
     }
+    resolverEntry.targetAddress = event.params.target;
```
0xdavinchee, line 79: logic change alert: always set target address for resolverEntry when handling set (must do this after the logic in the if clause above it)

packages/subgraph/src/utils.ts
```diff
@@ -78,13 +78,14 @@ export function handleTokenRPCCalls(
...
 ): Token {
-    token = getIsListedToken(token, resolverAddress);
-
     // we must handle the case when the native token hasn't been initialized
     // there is no name/symbol, but this may occur later
     if (token.name.length == 0 || token.symbol.length == 0) {
         token = getTokenInfoAndReturn(token);
     }
+    
+    // we do getIsListedToken after getTokenInfoAndReturn because it requires the token symbol
```
0xdavinchee, line 87: logic change alert: change order of rpc calls

packages/subgraph/src/utils.ts
```diff
@@ -109,34 +110,34 @@ export function getIsListedToken(
...
 
-export function updateTotalSupplyForNativeSuperToken(
-    token: Token,
+/**
+ * Gets and sets the total supply for TokenStatistic of a SuperToken upon initial creation
+ * @param tokenStatistic 
+ * @param tokenAddress 
+ * @returns TokenStatistic
+ */
+export function getInitialTotalSupplyForSuperToken(
```
0xdavinchee, line 131: logic change alert: change function to be called for all super tokens (not just custom)

0xdavinchee (commented):
Some comments to track the mapping logic changes at a high level

hellwolf (approved):

packages/subgraph/src/mappingHelpers.ts
```diff
@@ -80,17 +80,19 @@ export function getOrInitAccount(
  * We also create token stats in here if it doesn't exist yet.
  */
 export function getOrInitSuperToken(
+    event: ethereum.Event,
     tokenAddress: Address,
-    block: ethereum.Block
+    triggeredByEventName: string
```
kasparkallas, line 85: What's the difference between this and `event.parameters.name`?

packages/subgraph/src/mappingHelpers.ts
kasparkallas: Otherwise they're null or something?

packages/subgraph/src/mappingHelpers.ts
kasparkallas: :spaghetti:

packages/subgraph/tests/assertionHelpers.ts
```diff
@@ -0,0 +1,296 @@
...
+    const id = createEventID(eventName, event);
+    const order = getOrder(event.block.number, event.logIndex);
+
+    assert.fieldEquals(entityType, id, "id", id);
+    assert.fieldEquals(entityType, id, "blockNumber", event.block.number.toString());
+    assert.fieldEquals(entityType, id, "logIndex", event.logIndex.toString());
+    assert.fieldEquals(entityType, id, "order", order.toString());
+    assert.fieldEquals(entityType, id, "timestamp", event.block.timestamp.toString());
+    assert.fieldEquals(entityType, id, "transactionHash", event.transaction.hash.toHex());
+    assert.fieldEquals(entityType, id, "gasPrice", event.transaction.gasPrice.toString());
```
kasparkallas, line 28: I don't see `"name"` asserted.

packages/subgraph/src/mappingHelpers.ts
kasparkallas: When I check existing Subgraphs then it is `false` there too. This explicitness is forced by Matchstick and it shouldn't have effect on production?

packages/subgraph/src/mappings/resolver.ts
kasparkallas: I don't understand why this has to be set here. Doesn't `getOrInitResolverEntry` already handle it?

packages/subgraph/src/utils.ts
kasparkallas: I wonder why this changed?

packages/subgraph/src/mappingHelpers.ts
0xdavinchee: yup - the matchstick unit tests run into an issue where it tries to retrieve the values which are casted as not null (as they are defined as not null in our `schema.graphql`)

packages/subgraph/src/mappingHelpers.ts
0xdavinchee: yup, same thing as above where it expects not null, but is null. it has no effect on production as all underlying are false currently and if not, should be.

packages/subgraph/src/mappingHelpers.ts
0xdavinchee: Working on the refractor now, using nameof(event) where applicable.

packages/subgraph/src/mappingHelpers.ts
0xdavinchee: wat spaghetti

packages/subgraph/src/utils.ts
0xdavinchee: it didn't make sense to only execute this function for native asset/custom super tokens

packages/subgraph/src/mappings/resolver.ts
0xdavinchee: no - we don't set it in `getOrInitResolverEntry` because we have logic slightly above in the if block which relies on `targetAddress` not being set yet (for example removing a token from the resolver), which is why we do it at the end of the call.

packages/subgraph/tests/assertionHelpers.ts
0xdavinchee: true, this should be there

surewht (approved):

kasparkallas (approved):

PR #1184, opened 2022-11-22
[SUBGRAPH] Fix resolver breaking logic

0xdavinchee:
Issue: After a deployment to all endpoint today, we started receiving a lot of errors. This was due to none of the tokens being listed.

The initial thought was that some logic changes in the recent PR with matchstick broke it, but in fact it was an earlier [PR](https://github.com/superfluid-finance/protocol-monorepo/commit/810e42362259ba6bd242a09857c154d280617469) which broke this.

`let version = resolverAddress.equals(RESOLVER_ADDRESS) ? "test" : "v1";` this will always equal to "test" which led to all tokens being considered not listed: it would return 0 address when trying to do a `.get` on "supertokens.test.USDCx" for example.

### DIFF
--- a/packages/subgraph/src/addresses.template.ts
+++ b/packages/subgraph/src/addresses.template.ts
@@ -16,6 +16,6 @@ export function getNativeAssetSuperTokenAddress(): Address {
     return Address.fromString("{{nativeAssetSuperTokenAddress}}");
 }
 
-export function getIsTestNetwork(): boolean {
+export function getIsLocalIntegrationTesting(): boolean {
     return "{{testNetwork}}".length > 0;
 }
--- a/packages/subgraph/src/addresses.template.ts
+++ b/packages/subgraph/src/addresses.template.ts
@@ -15,3 +15,7 @@ export function getResolverAddress(): Address {
 export function getNativeAssetSuperTokenAddress(): Address {
     return Address.fromString("{{nativeAssetSuperTokenAddress}}");
 }
+
+export function getIsTestNetwork(): boolean {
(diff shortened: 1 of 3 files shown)


packages/subgraph/src/addresses.template.ts
```diff
@@ -15,3 +15,7 @@ export function getResolverAddress(): Address {
 export function getNativeAssetSuperTokenAddress(): Address {
     return Address.fromString("{{nativeAssetSuperTokenAddress}}");
 }
+
+export function getIsTestNetwork(): boolean {
```
kasparkallas: This might be a bit misleading because it doesn't mean a test network like "goerli", right?

0xdavinchee: yeah, I can rename

packages/subgraph/tests/mockedFunctions.ts
```diff
@@ -52,7 +52,7 @@ export function mockedHandleSuperTokenInitRPCCalls(
     // getIsListedToken(token, resolver) => resolver.try_get(key)
     mockedResolverGet(
         resolverAddress,
-        "supertokens.test." + tokenSymbol,
+        "supertokens.v1." + tokenSymbol,
```
kasparkallas, line 55: For local tests, it shouldn't be "test"?

0xdavinchee: for now, no, it's a little bit confusing I guess, but we're currently using matic addresses for the unit tests, addresses need to be deterministic

kasparkallas (approved):

PR #1189, opened 2022-11-24
[SUBGRAPH] Add `deposit` field to `AgreementLiquidatedByEvent`/`AgreementLiquidatedV2Event`

0xdavinchee:
This PR adds the full deposit amount of the stream being liquidated to the liquidation event.

## Changes summary
- `deposit` property added to liquidation events
- `stream` property added to `StreamRevision` entity
- It includes unit tests with mocked variables
- `id` field on `StreamRevision` entity changing `from sender-receiver-token` to `keccak256(abi.encode(sender,receiver))`
    > NOTE: this is **UNLIKELY** to be a breaking change as it is an entity that should be solely used for internal logic, but I will confirm with @kasparkallas before moving on with the deployment

### DIFF
--- a/packages/subgraph/schema.graphql
+++ b/packages/subgraph/schema.graphql
@@ -1518,8 +1518,9 @@ type StreamRevision @entity {
 
     """
     The "most recently alive" stream between a sender and receiver.
+    Note: The `revisionIndex` property may not be the same as the `revisionIndex` of `mostRecentStream`. Which means `mostRecentStream` has been closed and no new stream has been opened.
     """
-    stream: Stream!
+    mostRecentStream: Stream!
 }
 
 """
--- a/packages/subgraph/schema.graphql
+++ b/packages/subgraph/schema.graphql
@@ -969,6 +969,11 @@ type AgreementLiquidatedByEvent implements Event @entity {
     bondAccount: Bytes!
     rewardAmount: BigInt!
     bailoutAmount: BigInt!
+
+    """
+    The full deposit amount of the stream that was liquidated.
+    """
+    deposit: BigInt!
 }
 
 type AgreementLiquidatedV2Event implements Event @entity {
@@ -1000,6 +1005,11 @@ type AgreementLiquidatedV2Event implements Event @entity {
     targetAccountBalanceDelta: BigInt!
     version: BigInt!
     liquidationType: Int!
+
+    """
+    The full deposit amount of the stream that was liquidated.
+    """
+    deposit: BigInt!
 
     """
     TO BE DEPRECATED in v2 endpoint - use rewardAmountReceiver instead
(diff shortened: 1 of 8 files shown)


packages/subgraph/src/mappings/cfav1.ts
```diff
@@ -95,13 +95,15 @@ export function handleFlowUpdated(event: FlowUpdated): void {
...
     const isDelete = flowRate.equals(BIG_INT_ZERO);
+
     if (isDelete) {
         const streamRevision = getOrInitStreamRevision(
             senderAddress,
             receiverAddress,
             tokenAddress
         );
         streamRevision.revisionIndex = streamRevision.revisionIndex + 1;
+
```
kasparkallas, line 106: Is it still pointing to the previous stream here?

packages/subgraph/schema.graphql
```diff
@@ -1499,9 +1509,17 @@ type StreamPeriod @entity {
...
+    ID composed of: keccak256(abi.encode(sender,receiver))-tokenAddress
+    """
     id: ID!
     revisionIndex: Int!
     periodRevisionIndex: Int!
+
+    """
+    The "most recently alive" stream between a sender and receiver.
+    """
+    stream: Stream!
```
kasparkallas: So the revision becomes a cursor... :)

packages/subgraph/src/mappings/superToken.ts
```diff
@@ -317,6 +326,11 @@ function _createAgreementLiquidatedV2EventEntity(
         event.params.rewardAmountReceiver,
     ]);
 
+    const streamRevisionId =
+        event.params.id.toHex() + "-" + event.address.toHexString();
```
kasparkallas, line 330: The ID of the events is guaranteed to be: `keccak256(abi.encode(sender,receiver))`?

packages/subgraph/src/mappings/cfav1.ts
0xdavinchee: yeah, `StreamRevision` will always point to the "most recent alive" stream

packages/subgraph/schema.graphql
0xdavinchee: yup lmao

packages/subgraph/src/mappings/superToken.ts
0xdavinchee: yeah, the `id` field on the liquidation events is always going to be the `agreementId` and liquidations only exist for flows **currently** and will be so

packages/subgraph/src/mappings/cfav1.ts
kasparkallas: It's a bit weird. There will be a revision index, for example "5", but the stream being pointed at is "4", as long as the "5" is not created yet. Maybe rename to `recentStream` on `StreamRevision` if this behaviour is wanted.

0xdavinchee: fair, that is more descriptive

kasparkallas (approved):

PR #1196, opened 2022-12-05
[SDK-CORE] fix hardhat ethers dependency bug

### LINKED ISSUE #1174: [SDK-CORE] `@nomiclabs/hardhat-ethers/types` is in published package but not specified as a dependency
If any consumer uses tsconfig's `skipLibCheck=false` and they don't have `@nomiclabs/hardhat-ethers/types` installed themselves then it starts failing:
image

I think the root of the issue is the monorepo & yarn & CI set-up. As far as I can tell, when something is building in the pipeline then hard to say whether it got the dependencies from the root of the repo (which might come from another package), whether it used a local symlink, whether a devDependency is exposed in the final output, etc.

### LINKED ISSUE #1191: Framework.ts  depended @nomiclabs   is it necessary??
I  build for application  my  app error

`Error: node_modules/@superfluid-finance/sdk-core/dist/module/Framework.d.ts:1:38 - error TS2307: Cannot find module '@nomiclabs/hardhat-ethers/types' or its corresponding type declarations.

1 import { HardhatEthersHelpers } from "@nomiclabs/hardhat-ethers/types";`

It has a dependency with nomiclabs for the javascript sdk. The dependency for the test environment is reasonable, but is this dependency needed in production? thanks.

0xdavinchee:
addresses https://github.com/superfluid-finance/protocol-monorepo/issues/1174 and https://github.com/superfluid-finance/protocol-monorepo/issues/1191


hellwolf (approved):
lgtm

PR #1198, opened 2022-12-05
[CI/SUBGRAPH] Auto deploy subgraph

0xdavinchee:
- add subgraph deployment to canary CI
  - This is triggered **EVERY** time, but actually deployment only occurs if changes are made to the mapping logic.
- create reusable workflow for deploying subgraph
- use reusable workflow in ci.canary and handler.deploy-subgraph


hellwolf (approved):
cool stuff

PR #1199, opened 2022-12-05
[SUBGRAPH] Add `gasUsed` property to all events

### LINKED ISSUE #1103: [SUBGRAPH] add gasUsed property
# Problem

The subgraph contains information about `gasPrice`, but not about `gasUsed`, which makes it harder to compute the price of the transaction. 

#  Possible Solution
This [wasn't possible](https://github.com/graphprotocol/graph-node/issues/2619) until recently, but is now available on `event.receipt.gasUsed` [docs](https://thegraph.com/docs/en/developing/creating-a-subgraph/#transaction-receipts-in-event-handlers).

0xdavinchee:
* bump subgraph package versions
* add `gasUsed` property to all event entities
* add tests for `gasUsed` property


packages/subgraph/src/utils.ts
```diff
@@ -80,6 +80,15 @@ export function initializeEventEntity(
     entity.set("timestamp", Value.fromBigInt(event.block.timestamp));
     entity.set("transactionHash", Value.fromBytes(event.transaction.hash));
     entity.set("gasPrice", Value.fromBigInt(event.transaction.gasPrice));
+    if (event.receipt) {
+        // @note if we don't cast event.receipt as ethereum.TransactionReceipt,
+        // we get a compile error because it thinks event.receipt is:
+        // ethereum.TransactionReceipt | null
```
kasparkallas: Normally TS should be able to infer the correct type with this logic:
```ts
    const receipt = event.receipt;
    if (receipt) {
        entity.set("gasUsed", Value.fromBigInt(receipt.gasUsed));
    } else {
        entity.set("gasUsed", Value.fromBigInt(BIG_INT_ZERO));
    }
```

packages/subgraph/src/utils.ts
```diff
@@ -80,6 +80,15 @@ export function initializeEventEntity(
...
+        const receipt = event.receipt as ethereum.TransactionReceipt;
+        entity.set("gasUsed", Value.fromBigInt(receipt.gasUsed));
+    } else {
+        entity.set("gasUsed", Value.fromBigInt(BIG_INT_ZERO));
```
kasparkallas: If the receipt should always be present (i.e. specified in the `.yaml`) then I'd prefer throwing in general.

packages/subgraph/src/utils.ts
0xdavinchee: yeah I think someone from their side suggested this too, will make this change 👍🏼

packages/subgraph/src/utils.ts
0xdavinchee: true

packages/subgraph/src/utils.ts
kasparkallas: Remove the comment too.

packages/subgraph/src/utils.ts
kasparkallas: It's not going to throw ATM as the else-block is just removed? :stuck_out_tongue:

packages/subgraph/src/utils.ts
0xdavinchee: ok

packages/subgraph/src/utils.ts
0xdavinchee: it will throw because when we go and try to save the entity, this field is non-nullable

kasparkallas (approved):

PR #1233, opened 2022-12-13
[Github Actions] Fixes

0xdavinchee:
- remove duplicate check-changeset in ci.feature
- add check for subgraph changes in canary for deployment
  - I was incorrect about the subgraph not deploying on each merge to dev-this deploys because the commit hash is changed and this gets placed in a file which is part of the mapping
  - Solution: use check-changeset to only deploy if the subgraph files/folders in check-changeset have changed


hellwolf (approved):
lgtm

PR #1238, opened 2022-12-14
[ETHEREUM-CONTRACTS] | #1038 | Allow regular contracts to be the target of `callAppAction`

### LINKED ISSUE #1038: [ETHEREUM-CONTRACTS] Relax app registration requirement for callAppAction feature
To be clarified.

# Notes

- There is no strong security reason why only super app can use callAppAction. So long as a contract trust superfluid host contract, it could also use the feature.
- Alternative is to push people to use eip-2771 more and let batch call to support eip-2771.

0xdavinchee:
## Description

**Motivation:** To relax the requirement that `callAppAction` can only be used on super app contracts. 

**Context:** This change allows all contracts to be the target of `callAppAction` as long as the `function` on the contract satisfies two requirements:
1. The `function` you want to call via `callAppAction` must take `bytes` (the passed context from the host) as the last parameter.
2. You must return the "correct" context bytes from the contract in the `function`, that is, untouched.
> NOTE: this means that EOAs cannot be targets of `callAppAction` as any call against an EOA returns empty bytes.

This was achieved simply by deleting the `isAppActive` modifier from `Superfluid.sol`, `ISuperfluid.sol` and any associated custom errors only used by this modifer.

Issue: #1038

## Type of change

- [x] New feature (non-breaking change which adds functionality)

## How Has This Been Tested?

I have added tests to ensure that non-super app contracts can be the target of `callAppAction` and `callAppActionWithContext`.

- [x] 8.1-8.3 tests `callAppAction` reverting when passing invalid calldata
- [x] 8.4 tests `callAppAction` reverting when an EOA is the target, ctx is malformatted because bytes returned is "0x"
- [x] 8.5.0 tests a revert given requirement 1 failing
- [x] 8.5.1 tests a revert given requirement 1 passing, but requirement 2 failing (returning empty bytes)
- [x] 8.5.2 tests happy path `callAppAction`
- [x] 8.5.3 tests happy path `callAppActionWithContext`
- [x] 8.5.4 tests a revert `callAppActionWithContext` given that we ignore the context passed by the host via the initial `callAppAction` and pass in our own context


hellwolf:
Contract A

** foo (app action)
  [a] Mostly likely foo should check if msg.sender == host.
  [b] currently needs A to be whitelisted, in order to get correct msgSender.
  [c*] foo calls callAgreement should fail due to ctx not clean.
  [d*] foo calls callAgreementWithCtx should also fail due to app level restriction.

packages/ethereum-contracts/contracts/superfluid/Superfluid.sol
```diff
@@ -646,7 +646,6 @@ contract Superfluid is
     )
```
hellwolf, line 646: let's also ban calling `this` (the host contract itself) as target.

hellwolf: also remove the ISuperApp type, and call app e.g. "target"

0xdavinchee:
closing this PR for now

PR #1242, opened 2022-12-19
[ETHEREUM-CONTRACTS] | #1038 | Allow regular contracts to be the target of callAppAction

0xdavinchee:
## Description

**Motivation:** To relax the requirement that `callAppAction` can only be used on super app contracts. 

**Context:** This change allows all contracts to be the target of `callAppAction` as long as the `function` on the contract satisfies two requirements:
1. The `function` you want to call via `callAppAction` must take `bytes` (the passed context from the host) as the last parameter.
2. You must return the "correct" context bytes from the contract in the `function`, that is, untouched.
> NOTE: this means that EOAs cannot be targets of `callAppAction` as any call against an EOA returns empty bytes.

This was achieved simply by deleting the `isAppActive` modifier from `Superfluid.sol`, `ISuperfluid.sol` and any associated custom errors only used by this modifer.

Issue: #1038

## Type of change

- [x] New feature (non-breaking change which adds functionality)

## How Has This Been Tested?

I have added tests to ensure that non-super app contracts can be the target of `callAppAction` and `callAppActionWithContext`.

- [x] 8.1-8.3 tests `callAppAction` reverting when passing invalid calldata
- [x] 8.4 tests `callAppAction` reverting when an EOA is the target, ctx is malformatted because bytes returned is "0x"
- [x] 8.5.0 tests a revert given requirement 1 failing
- [x] 8.5.1 tests a revert given requirement 1 passing, but requirement 2 failing (returning empty bytes)
- [x] 8.5.2 tests happy path `callAppAction`
- [x] 8.5.3 tests happy path `callAppActionWithContext`
- [x] 8.5.4 tests a revert `callAppActionWithContext` given that we ignore the context passed by the host via the initial `callAppAction` and pass in our own context


0xdavinchee:
closing this PR for now as this is stale and nobody is asking for it

PR #1255, opened 2023-01-06
[SDK-redux] Fix transaction tracking for reverted transactions

kasparkallas:
Ethers throws for some transaction errors when waiting for the transaction to be mined but not when transaction gets reverted. This does makes sense because the transaction _was_ mined but in case of other errors it wasn't. To check whether the transaction was reverted, there's a status property on the transaction receipt:
image

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -3468,6 +3468,19 @@
     ethereumjs-util "7.1.5"
     stack-trace "0.0.10"
 
+"@superfluid-finance/ethereum-contracts@1.4.3":
+  version "1.4.3"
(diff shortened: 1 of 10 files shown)


hellwolf (approved):
Looks alright to me

yarn.lock
```diff
@@ -3468,6 +3468,19 @@
     ethereumjs-util "7.1.5"
     stack-trace "0.0.10"
 
+"@superfluid-finance/ethereum-contracts@1.4.3":
```
kasparkallas, line 3471: Not quite sure why this showed up here but should be fine.

PR #1261, opened 2023-01-24
[DIDI's MUSE] concept for simplified, delegatable CFA receive hooks

d10r:
Showcase for how we could implement delegatable CFA hooks.
* uses a simplified version (just one method)
* assumes we remove permissioning
* suggests to get rid of before/after split, because the only use case for that seems to be getting the previous flowrate on update or delete
* expects the hook's msg.sender to be the token contract - this eliminates the need for the handler to be aware of the host (TCI!)

Limitations:
* lacks ctx passing and checking (should likely be added)
* in this PR it isn't actually invoked in the CFA, because attempts to do so resulted in stack too deep / contract too big (depending on the chosen optimizer). Thus also no tests included.
But includes mock contracts to demonstrate envisioned usage.

Related considerations:
* no CRUD: instead of a CRUD model, there could also be a model where every possible flow exists, defaulting to flowrate 0 - this is how the CFA already partly works, e.g. getFlow for a "non-existing" flow won't fail, but return 0 values.
* no App-Credit: such a simple interface works only without App-Credit. Possible ways to deal with this
  * Add what's needed for App-Credit handling
  * Don't allow hook implementations to manipulate streams
  * Provide a few predefined hook building blocks known to not allow streams to become insolvent. Probably requires a way to compose hook building blocks + make sure removing a hook delegation can't circumvent it. More research needed.

### DIFF
--- a/packages/ethereum-contracts/contracts/agreements/ConstantFlowAgreementV1.sol
+++ b/packages/ethereum-contracts/contracts/agreements/ConstantFlowAgreementV1.sol
@@ -20,6 +20,9 @@ import { AgreementBase } from "./AgreementBase.sol";
 import { SafeCast } from "@openzeppelin/contracts/utils/math/SafeCast.sol";
 import { AgreementLibrary } from "./AgreementLibrary.sol";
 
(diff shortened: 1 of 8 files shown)


d10r:
related to https://github.com/superfluid-finance/protocol-monorepo/pull/1261

PR #1264, opened 2023-01-26
[ETHEREUM-CONTRACTS] add forge build to pre commit for contracts

0xdavinchee:
## Purpose
To ensure that compilation passes before commit is allowed. Devs no longer have to remember to build before committing, it is automated.
> Caveat: this doesn't catch things which the solidity compiler allows (doing a return statement leading to unreachable code for example).

## Context
Worst case: **20 seconds** to compile (on fresh compile).
Workflows:
- working on contracts w/ TDD (should add insignificant amount of time to pre-commit)
- working on contracts w/o TDD (will catch any compile errors if any and may take up to 20 seconds if all contracts require compilation)
- working on peripheries (may add 20 seconds on initial commit/build)


hellwolf:
One can always use `git commit --no-verify` for skipping pre-commit hooks.

hellwolf (approved):

PR #1266, opened 2023-01-27
Fix the daily slack bot

elvijsTDL:
- Fixed message not running when no PRs are open
- Extra messages in cases there are only draft PRs or none at all
- Fixed pending workflow status


hellwolf (approved):
GGG

PR #1270, opened 2023-01-31
[SDK-core] Add StreamId to FlowUpdatedEvent when querying from Subgraph

kasparkallas:
Why?
Useful for UI-s.

### DIFF
--- a/packages/sdk-core/src/events.ts
+++ b/packages/sdk-core/src/events.ts
@@ -77,7 +77,7 @@ export interface FlowUpdatedEvent extends EventBase {
     flowRate: string;
     flowOperator: string;
     deposit: string;
-    streamId: string; // Subgraph ID.
+    streamId: string; // `Stream.id` from Subgraph, composed of: senderAddress-receiverAddress-tokenAddress-revisionIndex
 }
 
 export interface FlowOperatorUpdatedEvent extends EventBase {
--- a/packages/sdk-core/src/events.ts
+++ b/packages/sdk-core/src/events.ts
@@ -77,6 +77,7 @@ export interface FlowUpdatedEvent extends EventBase {
     flowRate: string;
     flowOperator: string;
     deposit: string;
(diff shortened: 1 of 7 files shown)


yarn.lock
```diff
@@ -3468,19 +3468,6 @@
     ethereumjs-util "7.1.5"
     stack-trace "0.0.10"
 
-"@superfluid-finance/ethereum-contracts@1.4.3":
```
0xdavinchee, line 3471: what happened here? @kasparkallas

packages/sdk-core/src/events.ts
```diff
@@ -77,6 +77,7 @@ export interface FlowUpdatedEvent extends EventBase {
     flowRate: string;
     flowOperator: string;
     deposit: string;
+    streamId: string; // Subgraph ID.
```
0xdavinchee: I would personally be a bit more explicit here and say something like `Subgraph Superfluid Stream ID`.
When I see Subgraph ID, I initially think of this (might just be me though): `QmVQcG7NusjRarbmL4717WoaaAx8mSynTrek73NmznYYp3` the IPFS hash of the deployed subgraph.

yarn.lock
kasparkallas: It weirdly appeared here with a `yarn install`: https://github.com/superfluid-finance/protocol-monorepo/pull/1255#discussion_r1087562403 and now it disappeared again. :joy:

packages/sdk-core/src/events.ts
kasparkallas: Changed: 
image

0xdavinchee: even better!

yarn.lock
0xdavinchee: packages r pain 🥲

0xdavinchee (approved):
comments addressed and tests passing - LGTM

PR #1278, opened 2023-02-07
SuperAppBaseFlow

joeljesu:


packages/ethereum-contracts/contracts/mocks/SuperAppBaseFlowTester.sol
```diff
@@ -0,0 +1,143 @@
...
+
+    address public afterReceiverHolder;
+    address public beforeReceiverHolder;
+
+    constructor(
+        ISuperfluid host
+    ) SuperAppBaseFlowMock(
+        host,
+        0
+    ) {}
```
hellwolf: use a comment to remove no-empty-blocks warning.

packages/ethereum-contracts/contracts/apps/SuperAppBaseFlow.sol
```diff
@@ -0,0 +1,265 @@
...
+        _host = host_;
+
+        host_.registerApp(SuperAppDefinitions.APP_LEVEL_FINAL | callBackDefinitions);
+    }
+
+    /**
+     * @dev Expect Super Token involved in callback to be an accepted one 
+     *      This function can be overridden with custom logic and to revert if desired
+     */
+    function _isAcceptedSuperToken(ISuperToken superToken) public view virtual returns (bool) {
```
d10r: if public, the fn name shouldn't be prefixed with _

Is there a reason why _isAcceptedSuperToken is public while _isAcceptedAgreement is internal?

packages/ethereum-contracts/contracts/apps/SuperAppBaseFlow.sol
```diff
@@ -0,0 +1,265 @@
...
+                superToken,
+                sender,
+                receiver,
+                cbdata,
+                ctx
+            );
+    }
+
+    // ---------------------------------------------------------------------------------------------
+    // BEFORE- CALLBACKS
```
d10r: maybe place those before the "after" callbacks?

d10r:
This is nice!
I'd consider going a step further and provide the following API
```
function onFlowCreated(sender, ctx);
function onFlowUpdated(sender, oldFlowRate, ctx);
function onFlowDeleted(sender, receiver, oldFlowRate, ctx);
```

Reasoning:
The before callbacks seem to be used mainly for providing the old flowrate in the after callback.
If we already provide that in this convenience wrapper, we could simplify its API further.

Main drawbacks I see:
* less flexibility: can't forward more than oldFlowrate (e.g. last updated as in the Tester contract). Minor issue imo as there's always the lower level API for more demanding use cases
* some gas waste: e.g. in order to provide the old flowrate by default, before callbacks need to be enabled by default for update ande delete, even if the information isn't needed. Not sure how much difference it makes, it's a convenience vs gas efficiency tradeoff. Could be mitigated by allowing to disable the before callbacks anyway, in that case having the old flowrate always set to 0 (as 0 isn't a possible value in a update or delete) or -1 (there's no such think as NaN in Solidity afaik).

I'd also consider adding a convenience wrapper for the callback selection - e.g. 3 bools for create/update/delete, using positive logic (not inverted as in the low level API).
And maybe add a function `getUserData(bytes calldata ctx)`?

packages/ethereum-contracts/contracts/apps/SuperAppBaseFlow.sol
joeljesu: Thanks for the _catch. 

There isn't a reason for _isAcceptedAgreement to be internal but there isn't really much of a reason for it not to be. I can make it public for good measure

packages/ethereum-contracts/contracts/apps/SuperAppBaseFlow.sol
joeljesu: I put the "before" callbacks after the "after" callbacks (lol) as the "after" callbacks are of more interest to the developer and are what's actually overriden and used. The other stuff is behind the scenes. But I can switch them around if you think we should be more conventional

joeljesu:
> This is nice! I'd consider going a step further and provide the following API
> 
> ```
> function onFlowCreated(sender, ctx);
> function onFlowUpdated(sender, oldFlowRate, ctx);
> function onFlowDeleted(sender, receiver, oldFlowRate, ctx);
> ```
> 
> Reasoning: The before callbacks seem to be used mainly for providing the old flowrate in the after callback. If we already provide that in this convenience wrapper, we could simplify its API further.
> 
> Main drawbacks I see:
> 
> * less flexibility: can't forward more than oldFlowrate (e.g. last updated as in the Tester contract). Minor issue imo as there's always the lower level API for more demanding use cases
> * some gas waste: e.g. in order to provide the old flowrate by default, before callbacks need to be enabled by default for update ande delete, even if the information isn't needed. Not sure how much difference it makes, it's a convenience vs gas efficiency tradeoff. Could be mitigated by allowing to disable the before callbacks anyway, in that case having the old flowrate always set to 0 (as 0 isn't a possible value in a update or delete) or -1 (there's no such think as NaN in Solidity afaik).
> 
> I'd also consider adding a convenience wrapper for the callback selection - e.g. 3 bools for create/update/delete, using positive logic (not inverted as in the low level API). And maybe add a function `getUserData(bytes calldata ctx)`?

I think I'll stick with the afterFlow- nominclature so it's clear that the actions written in there take place _after_ and they're working with the post-state.

The beforeAgreement- hook is also used for passing over the timestamp of the last flow update. 

The initial reason for excluding the oldFlowRate and oldUpdateTimestamp in the callback headers was that it would make the headers more crowded for something that's seldomly used. However, I agree the cognitive load of reasoning with the beforeAgreement- callback ought to be removed. This can be accomplish by encoding and providing the old flow rate and old update timestamp through `beforeData` by default in the beforeAgreement- callbacks. Wdyt?

I also like the boolean selection idea for callback selection, kind of like what we did in the forwarder I think. I'll see what I can do here.

Lastly a heads up that I am having a terribletime dealing with the protocol-monorepo locally. I've had to do so many little hacky things to get it working to reach this PR and they've just been wiped when I accidentally overrode my local changes with the remote. It may take a good while for me to get things up and running again to actually enact these suggestions

packages/ethereum-contracts/contracts/mocks/SuperAppBaseFlowMock.sol
```diff
@@ -0,0 +1,282 @@
+// SPDX-License-Identifier: AGPLv3
+pragma solidity >= 0.8.0;
+
+import {ISuperfluid, ISuperToken, ISuperApp, SuperAppDefinitions} from "../interfaces/superfluid/ISuperfluid.sol";
+import {SuperTokenV1Library} from "../apps/SuperTokenV1Library.sol";
+
+abstract contract SuperAppBaseFlowMock is ISuperApp {
```
d10r: this contract should inherit from SuperAppBaseFlow instead of duplicating its code, otherwise the tests aren't testing the actual code, but a copy of it (wihch may get out of sync).

packages/ethereum-contracts/contracts/mocks/SuperAppBaseFlowMock.sol
```diff
@@ -0,0 +1,282 @@
...
+    using SuperTokenV1Library for ISuperToken;
+    
+    ISuperfluid public _host;
+
+    /// @dev Can be used to track accepted status of tokens for Super App
+    mapping(ISuperToken => bool) internal _acceptedSuperTokens;
+
+    bytes public beforeCtxHolder;
+    bytes public afterCtxHolder;
+    string public callbackHolder;
```
d10r: to be checked: what exactly are you testing here? Is this really about the SuperAppBaseFlow or redundantly checking the underlying mechanisms?

packages/ethereum-contracts/test/foundry/apps/SuperAppBaseFlow.t.sol
```diff
@@ -0,0 +1,51 @@
...
+
+import "forge-std/console.sol";
+import "../FoundrySuperfluidTester.sol";
+import { SuperAppBaseFlow } from "@superfluid-finance/ethereum-contracts/contracts/apps/SuperAppBaseFlow.sol";
+import { SuperAppBaseFlowTester } from "@superfluid-finance/ethereum-contracts/contracts/mocks/SuperAppBaseFlowTester.sol";
+import { ISuperToken } from "@superfluid-finance/ethereum-contracts/contracts/interfaces/superfluid/ISuperfluid.sol";
+import { IConstantFlowAgreementV1 } from "@superfluid-finance/ethereum-contracts/contracts/interfaces/agreements/IConstantFlowAgreementV1.sol";
+import { SuperTokenV1Library } from "@superfluid-finance/ethereum-contracts/contracts/apps/SuperTokenV1Library.sol";
+
+contract SuperAppBaseFlowTest is FoundrySuperfluidTester {
```
d10r: first, start a nix shell in order to have the required tooling:
```
nix develop
```
(see CONTRIBUTING.md for how to set up Nix)

Then you can start a TDD session with 
```
nodemon -w test -w contracts/ -e sol --exec forge test --match-contract SuperAppBaseFlowTest -vvv
```

packages/ethereum-contracts/test/foundry/apps/SuperAppBaseFlow.t.sol
```diff
@@ -0,0 +1,51 @@
...
+
+    function testFlowToSuperApp() public {
+        vm.startPrank(alice);
+        superToken.createFlow(address(superApp), int96(69));
+        vm.stopPrank();
+    }
+
+    function testFlowOfNotAcceptedSuperTokenToSuperApp() public {
+        vm.startPrank(alice);
+        vm.expectRevert();
```
d10r: This test is expected to fail when sending a not-accepted token to the SuperApp. However the SuperAppBaseFlow doesn't revert when sending a not accepted SuperApp. It probably should (?)

This is currently not working as expected, for reasons we haven't yet figured out foundry doesn't recognize this exception even if triggered.

d10r (changes requested):

packages/ethereum-contracts/contracts/mocks/SuperAppBaseFlowMock.sol
joeljesu: It may be redundant checking, but definitely zooming into underlying behavior that happens during calls and doesn't reflect in post-state

d10r:
there's currently an issue with the foundry test - it fails to recognize reverts in the SuperApp.
If making `SuperAppBaseFlow.beforeAgreementCreated()` revert for unaccepted SuperTokens (which I think it should), the test `testFlowOfNotAcceptedSuperTokenToSuperApp` should recognize that.

This is the output of 2 runs with the revert in place (`if (!isAcceptedSuperToken(superToken)) revert("sad");`), the first run with the test not expecting a revert, the second one expecting a revert (with `vm.expectRevert();`):

```
$ nodemon -w test -w contracts/ -e sol --exec forge test --match-contract SuperAppBaseFlowTest   

# 1st run

[⠔] Compiling...
No files changed, compilation skipped

Running 3 tests for packages/ethereum-contracts/test/foundry/apps/SuperAppBaseFlow.t.sol:SuperAppBaseFlowTest
[FAIL. Reason: sad] testFlowOfNotAcceptedSuperTokenToSuperApp() (gas: 172255)
[PASS] testFlowToSuperApp() (gas: 592917)
[PASS] testSetAcceptedToken() (gas: 37899)
Test result: FAILED. 2 passed; 1 failed; finished in 8.75ms

Failing tests:
Encountered 1 failing test in packages/ethereum-contracts/test/foundry/apps/SuperAppBaseFlow.t.sol:SuperAppBaseFlowTest
[FAIL. Reason: sad] testFlowOfNotAcceptedSuperTokenToSuperApp() (gas: 172255)

Encountered a total of 1 failing tests, 2 tests succeeded
[nodemon] app crashed - waiting for file changes before starting...

# 2nd run:

[⠰] Compiling...
[⠘] Compiling 1 files with 0.8.18
[⠊] Solc 0.8.18 finished in 5.28s
Compiler run successful (with warnings)
warning[2018]: Warning: Function state mutability can be restricted to view
   --> packages/ethereum-contracts/contracts/superfluid/SuperTokenFactory.sol:341:5:
    |
341 |     function createSuperTokenLogic(
    |     ^ (Relevant source part starts here and spans across multiple lines).


Running 3 tests for packages/ethereum-contracts/test/foundry/apps/SuperAppBaseFlow.t.sol:SuperAppBaseFlowTest
[FAIL. Reason: Call did not revert as expected] testFlowOfNotAcceptedSuperTokenToSuperApp() (gas: 23441)
[PASS] testFlowToSuperApp() (gas: 592917)
[PASS] testSetAcceptedToken() (gas: 37899)
Test result: FAILED. 2 passed; 1 failed; finished in 20.09ms

Failing tests:
Encountered 1 failing test in packages/ethereum-contracts/test/foundry/apps/SuperAppBaseFlow.t.sol:SuperAppBaseFlowTest
[FAIL. Reason: Call did not revert as expected] testFlowOfNotAcceptedSuperTokenToSuperApp() (gas: 23441)

Encountered a total of 1 failing tests, 2 tests succeeded
```

Updating to latest forge (`forge 0.2.0 (9477764 2023-03-02T00:05:27.278234973Z)`) didn't help.

d10r:
changes:
* default impl of `isAcceptedSuperToken` always returns true. In order to apply a filter, the SuperApp has to override it. This way it's easier to allow this default behaviour + not put any restrictions on the possible logic, e.g. one could filter for SuperTokens listed in the Superfluid resolver.
* The methods to be overridden were renamed from afterX to onX - makes the names more distinct from the low level callbacks and is closer to the convention of how callbacks/hooks are usually named.
* the default implementations don't revert - remove hazard of getting jailed because activating a callback which isn't overridden.
* beforeAgreementCreated callback is always disabled because we don't need it (no data to be forwarded)

@ngmachado is completing the tests

ngmachado:
Made some changes:

constructor arguments to `bool activateOnCreated, bool activateOnUpdated, bool activateOnDeleted`

d10r (approved):

hellwolf (approved):
Excellent @sunnyjaycer @d10r

PR #1296, opened 2023-03-01
[SUBGRAPH/SDK-CORE] | Issues #1273 - ATS properties, 1291 - configs on Token, 1292 - userData on Stream

### LINKED ISSUE #1273: [SUBGRAPH] New aggregate properties
## Problem
@kasparkallas recently brought up some confusion surrounding the naming of one of the properties: `totalNumberOfActiveStreams`.

## Proposal
He proposed to add some new properties
- `activeOutgoingStreamCount`: this is equivalent to the current `totalNumberOfActiveStreams` property
- `activeIncomingStreamCount`:
- `inactiveOutgoingStreamCount`
- `inactiveIncomingStreamCount`: this is equivalent to the current `totalNumberOfClosedStreams` property

I also had a thought that we could add some aggregate properties/entities surrounding liquidations, PIC, TOGA.
For example, we can track the total amount of rewards an account has accrued from liquidations over its lifetime as well as the cost of these liquidations (this logic would be a bit harder to implement though). 
We could also track total # of liquidations per token, account-token, etc.

This would only be adding new properties, so it would be a **NON-BREAKING** change.

### LINKED ISSUE #1291: [SUBGRAPH] Governance Configs-per token level
We should have some governance configs (e.g. min deposit) on the `Token` entity

### LINKED ISSUE #1292: [SUBGRAPH] Add `userData` to `FlowUpdatedEvent`
Why?
`userData` is not stored on-chain. It's emitted with the event for indexing purposes. It's easy and cheap to index it with Subgraph.

`FlowUpdatedEvent` is an obvious choice where to put it. 
We could also put it on `StreamPeriod` but doesn't seem too necessary as it's easy to select it through the nested `startedAtEvent`. 
If we put it on `Stream` then it would only keep track of the last `userData`, so it'd be a changing field value. That might make sense too.

0xdavinchee:
## Summary
#1273 - New aggregate properties added for ATS and ATSLog entities:
- `activeOutgoingStreamCount`
- `activeIncomingStreamCount`
- `inactiveOutgoingStreamCount`
- `inactiveIncomingStreamCount`

#1291 - Governance Configs on `Token` entity
- `rewardAddress`, `liquidationPeriod`, `patricianPeriod`, and `minimumDeposit` added to `Token`
- New `DefaultGovernanceConfigs` HOL entity created for single entity containing our default governance parameters for all tokens if they don't have a config set specifically (when the `superToken` field is 0x0)
- Added `SuperTokenMinimumDepositChangedEvent` entity which was previously missing

cc @vmichalik 

#1292 - `userData` field added to `Stream` entity
- `userData` property added to `Stream` and represents the last updated `userData`

Misc
- @kasparkallas requested to throw an error on BatchCall in SDK-Core as well

SDK-Core subgraph mapping added as well.


packages/subgraph/src/mappingHelpers.ts
```diff
@@ -143,6 +163,30 @@ export function getOrInitSuperToken(
...
+
+    if (token.liquidationPeriod.equals(BIG_INT_ZERO) && defaultGovernanceConfigs != null) {
+        token.liquidationPeriod = defaultGovernanceConfigs.liquidationPeriod;
+    }
+
+    if (token.patricianPeriod.equals(BIG_INT_ZERO) && defaultGovernanceConfigs != null) {
+        token.patricianPeriod = defaultGovernanceConfigs.patricianPeriod;
+    }
+
+    if (token.minimumDeposit.equals(BIG_INT_ZERO) && defaultGovernanceConfigs != null) {
```
kasparkallas: Nest all the conditions in a single `if (defaultGovernanceConfigs != null)` to repeat less?

packages/subgraph/src/mappingHelpers.ts
```diff
@@ -154,6 +198,34 @@ export function getOrInitSuperToken(
...
 
+export function getOrInitDefaultGovernanceConfigs(
+    block: ethereum.Block
+): DefaultGovernanceConfigs {
+    let defaultGovernanceConfigs = DefaultGovernanceConfigs.load(
+        ZERO_ADDRESS.toHexString()
+    );
+
+    if (defaultGovernanceConfigs == null) {
+        defaultGovernanceConfigs = new DefaultGovernanceConfigs(
```
kasparkallas: Rename to singular? "DefaultGovernanceConfig"

packages/subgraph/src/mappingHelpers.ts
```diff
@@ -154,6 +198,34 @@ export function getOrInitSuperToken(
...
+            ZERO_ADDRESS.toHexString()
+        );
+        defaultGovernanceConfigs.createdAtTimestamp = block.timestamp;
+        defaultGovernanceConfigs.createdAtBlockNumber = block.number;
+        defaultGovernanceConfigs.rewardAddress = ZERO_ADDRESS;
+        defaultGovernanceConfigs.liquidationPeriod = BIG_INT_ZERO;
```
kasparkallas: What's the point of the default config if even `liquidationPeriod` is zero?

packages/subgraph/src/mappings/superfluidGovernance.ts
```diff
@@ -41,6 +46,29 @@ export function handleRewardAddressChanged(event: RewardAddressChanged): void {
     ev.rewardAddress = event.params.rewardAddress;
     ev.save();
 
+    // Update reward address for token or default governance configs
+    if (event.params.superToken.equals(ZERO_ADDRESS)) {
```
kasparkallas: Why would this ever be 0-address? Is this a pattern in the protocol?

packages/subgraph/src/mappings/superfluidGovernance.ts
```diff
@@ -41,6 +46,29 @@ export function handleRewardAddressChanged(event: RewardAddressChanged): void {
...
+        let defaultGovernanceConfigs = DefaultGovernanceConfigs.load(
+            ZERO_ADDRESS.toHexString()
+        );
+        if (defaultGovernanceConfigs) {
+            defaultGovernanceConfigs.rewardAddress = event.params.rewardAddress;
```
kasparkallas: I don't think `updated*` fields are getting updated on this code path.

packages/subgraph/schema.graphql
```diff
@@ -1656,6 +1688,43 @@ type StreamPeriod @entity {
     totalAmountStreamed: BigInt
 }
 
+"""
+DefaultGovernanceConfigs: A higher order entity that represents the default governance configs.
+When a config is set in governance with superToken field as address(0), it will be used as the default config.
+"""
+type DefaultGovernanceConfigs @entity {
```
kasparkallas: So there is this one default configuration entity? And when there's anything token specific then that is saved on the token?

packages/subgraph/src/mappingHelpers.ts
0xdavinchee: agree and resolved

packages/subgraph/src/mappingHelpers.ts
0xdavinchee: agree and resolved

packages/subgraph/src/mappingHelpers.ts
0xdavinchee: because it is the true "default", default in this case refers to what is set for a config (e.g. rewardAddress, liqPeriod, patPeriod, minDeposit) where the superToken field is `address(0)`.
each of these fields are only set at the protocol (and therefore the subgraph) when an event is emitted indicating that the default value has been set for a particular config.
it is better to initialize the entity with properties = 0 which means nothing has been set yet than to initialize at a value (which may be incorrect) we've used in the past.

packages/subgraph/src/mappings/superfluidGovernance.ts
0xdavinchee: yes, see the comment above or:

default configs refers to configs set for `event.params.superToken = address(0)`, that is, the super token parameter/argument for a governance config is address(0).

packages/subgraph/schema.graphql
0xdavinchee: yes

packages/subgraph/src/mappings/superfluidGovernance.ts
0xdavinchee: I am assuming you are referring to governance configs being updated here and that case is handled here.

packages/subgraph/src/mappings/superfluidGovernance.ts
```diff
@@ -61,6 +89,31 @@ export function handleCFAv1LiquidationPeriodChanged(
     ev.isKeySet = event.params.isKeySet;
     ev.liquidationPeriod = event.params.liquidationPeriod;
     ev.save();
+
+    // Update cfav1 liquidation period for token or default governance configs
+    if (event.params.superToken.equals(ZERO_ADDRESS)) {
+        let defaultGovernanceConfigs = DefaultGovernanceConfigs.load(
```
kasparkallas: Why not use `getOrInitDefaultGovernanceConfigs` here?

packages/subgraph/src/mappings/superfluidGovernance.ts
kasparkallas: I mean I don't notice these getting updated:
image

packages/subgraph/src/mappings/superfluidGovernance.ts
0xdavinchee: oh derp, good point, the pushed fix resolves this comment and the one above regarding the updatedAt fields...

packages/subgraph/src/mappingHelpers.ts
```diff
@@ -178,6 +207,13 @@ export function getOrInitToken(
     token.isSuperToken = false;
     token.isNativeAssetSuperToken = false;
     token.isListed = false;
+
+    // This is unset for the underlying ERC20.
```
kasparkallas: Can't leave them undefined?

Also, this is a bit smelly. Having properties on a type that for big portion of the instances don't make sense. Would have probably preferred an explicit TokenGovernance table.

packages/subgraph/src/mappingHelpers.ts
```diff
@@ -246,6 +282,7 @@ export function getOrInitStream(event: FlowUpdated): Stream {
         stream.streamedUntilUpdatedAt = BigInt.fromI32(0);
         stream.updatedAtTimestamp = currentTimestamp;
         stream.updatedAtBlockNumber = event.block.number;
+        stream.userData = Bytes.empty();
```
kasparkallas: Hmm, why not use the `userData` from the event?

packages/subgraph/src/mappings/superfluidGovernance.ts
```diff
@@ -61,6 +78,19 @@ export function handleCFAv1LiquidationPeriodChanged(
...
+        const defaultGovernanceConfig = getOrInitDefaultGovernanceConfig(event.block);
+        defaultGovernanceConfig.liquidationPeriod = event.params.liquidationPeriod;
+        defaultGovernanceConfig.save();
+    } else {
+        const token = Token.load(event.params.superToken.toHexString());
+        if (token) {
+            token.liquidationPeriod = event.params.liquidationPeriod;
+            token.save();
+        }
+    }
```
kasparkallas: (follow-up comment to "TokenGovernance") most of these code blocks would make more sense like this:
```ts
const governance = getOrInitGovernanceConfig(event.params.superToken.toHexString(), event.block);
governance.liquidationPeriod = event.params.liquidationPeriod;
governance.updatedAtTimestamp = event.block.timestamp;
governance.updatedAtBlock = event.block.number
governance.save();
```
Would not need conditionals in these situations. Also token-specific governance would have timestamps.

packages/subgraph/src/mappingHelpers.ts
0xdavinchee: true

kasparkallas (approved):
LGTM :+1:

.github/workflows/call.deploy-subgraph.yml
```diff
@@ -89,7 +89,10 @@ jobs:
         working-directory: ${{ env.subgraph-working-directory }}
 
       - name: "Prepare subgraph manifest"
-        run: ./tasks/prepare-manifest.sh ${{ inputs.network }}
+        # note that this step is only used for generating an arbitrary subgraph.yaml
+        # and addresses.ts file for the yarn codegen step below
+        # the actual addresses will be retrieved during the actual deployment of the subgraph
+        run: ./tasks/prepare-manifest.sh
```
hellwolf: hardcode "matic" here instead.

hellwolf (approved):
one minor comment, please address.

PR #1298, opened 2023-03-01
[ETHEREUM-CONTRACTS/SDK-CORE] Increase/DecreaseAllowance for ACL Flowrate

### LINKED ISSUE #1300: [ETHEREUM-CONTRACTS] Better support for smart contract wallets
### Discussed in https://github.com/superfluid-finance/protocol-monorepo/discussions/1299

<div type='discussions-op-text'>

<sup>Originally posted by **0xdavinchee** March  2, 2023</sup>
## Overview
Smart contract wallets will likely become more and more prevalent in the crypto ecosystem to improve wallet UX and general UX when using crypto. 

There are currently some known pain points when setting up salaries or vesting via a gnosis safe and we aim to address this at the protocol level.

## Current flow and rationale
From @vmichalik:
- when an admin is setting up salaries or vesting, they need to give flowrate allowance to open a stream
- if they are setting up multiple streams from a gnosis safe, they’ll go through the process to set up each scheduled stream, and those transactions will be pending on the multisig
- we have no way of knowing what the total flowrate allowance needs to be in the process of crafting each transaction by checking the current state since the other transactions are still pending
- if instead of updating the flow (without enough information), each transaction increases the flow, the total will always be correct
- this is way safer and better than unlimited approving flowrate

## Solution for known issue
Below are some of the features we can implement to improve some of the current flows:
- [x] Implement `increaseFlowRateAllowance` and `decreaseFlowRateAllowance` functions in `ConstantFlowAgreementV1`. This works the same as `increaseAllowance` and `decreaseAllowance` for `SuperToken`
- [x] Make `increaseAllowance` and `decreaseAllowance` a batch-able operation in `Superfluid.sol`
</div>

0xdavinchee:
## Context
Related issue: https://github.com/superfluid-finance/protocol-monorepo/issues/1300
Related discussion: https://github.com/superfluid-finance/protocol-monorepo/discussions/1299

### ETH-CONTRACTS
- increase/decrease flow operator flow rate allowance function added to cfaV1
- increase/decrease erc20 token allowance added as a BatchOperation type
  - `operationIncreaseAllowance`
  - `operationDecreaseAllowance`

### SDK-Core
- support for new functions added

## Other notes
I intentionally did not add the functions to `CFAv1Forwarder` and `SuperTokenV1Library`


packages/ethereum-contracts/contracts/interfaces/agreements/IConstantFlowAgreementV1.sol
```diff
@@ -408,6 +440,14 @@ abstract contract IConstantFlowAgreementV1 is ISuperAgreement {
         int96 flowRateAllowance
     );
 
+    event FlowOperatorAllowanceUpdated(
```
0xdavinchee: remove this and use the existing flowoperatorpermissions updated event

packages/ethereum-contracts/contracts/superfluid/SuperToken.sol
```diff
@@ -432,8 +432,7 @@ contract SuperToken is
 
     function decreaseAllowance(address spender, uint256 subtractedValue)
         public override returns (bool) {
-        _approve(msg.sender, spender, _allowances[msg.sender][spender].sub(subtractedValue,
-            "SuperToken: decreased allowance below zero"));
+        _approve(msg.sender, spender, _allowances[msg.sender][spender] - subtractedValue);
```
0xdavinchee: remove safemath lib from file

packages/ethereum-contracts/contracts/superfluid/SuperToken.sol
```diff
@@ -167,7 +165,7 @@ contract SuperToken is
             _approve(
                 holder,
                 spender,
-                _allowances[holder][spender].sub(amount, "SuperToken: transfer amount exceeds allowance"));
+                _allowances[holder][spender] - amount);
```
hellwolf: why?

packages/ethereum-contracts/test/foundry/agreements/ConstantFlowAgreementV1.ACL.t.sol
```diff
@@ -0,0 +1,294 @@
+// SPDX-License-Identifier: AGPLv3
```
hellwolf, line 1: Good.

hellwolf (commented):
one comment

packages/ethereum-contracts/contracts/superfluid/SuperToken.sol
hellwolf: It becomes harder for direct etherscan-diagnosis on what is the error.

0xdavinchee: reverted to use safemath

packages/ethereum-contracts/contracts/agreements/ConstantFlowAgreementV1.sol
```diff
@@ -813,6 +813,88 @@ contract ConstantFlowAgreementV1 is
         emit FlowOperatorUpdated(token, currentContext.msgSender, flowOperator, permissions, flowRateAllowance);
     }
 
+    /// @dev IConstantFlowAgreementV1.increaseFlowRateAllowance implementation
+    function increaseFlowRateAllowance(
+        ISuperfluidToken token,
+        address flowOperator,
+        int96 flowRateAllowanceDelta, // flowRateBudget
```
kasparkallas: We don't want to use an unsigned value (`uint48`)? Like in the [OpenZeppelin ERC20 implementation](https://docs.openzeppelin.com/contracts/2.x/api/token/erc20#ERC20-increaseAllowance-address-uint256-)

Maybe also name "addedFlowRateAllowance" and "subtractedFlowRateAllowance" respectively (or slightly shorter varians "added*"/"subtracted*" variants). Mentioning _delta_ with the signed type just adds to the confusion...

packages/ethereum-contracts/contracts/agreements/ConstantFlowAgreementV1.sol
```diff
@@ -813,6 +813,88 @@ contract ConstantFlowAgreementV1 is
...
+        bytes calldata ctx
+    ) public override returns (bytes memory newCtx) {
+        newCtx = ctx;
+
+        // [SECURITY] NOTE: we are holding the assumption here that ctx is correct and we validate it with
+        // authorizeTokenAccess:
+        ISuperfluid.Context memory currentContext = AgreementLibrary.authorizeTokenAccess(token, ctx);
+
+        if (currentContext.msgSender == flowOperator) revert CFA_ACL_NO_SENDER_FLOW_OPERATOR();
+        if (flowRateAllowanceDelta < 0) revert CFA_ACL_NO_NEGATIVE_ALLOWANCE();
```
kasparkallas: Having exactly 0 here is also a bit funny but I guess not necessary to revert.

packages/ethereum-contracts/contracts/agreements/ConstantFlowAgreementV1.sol
```diff
@@ -813,6 +813,88 @@ contract ConstantFlowAgreementV1 is
...
+            token,
+            flowOperatorId,
+            oldPermissions,
+            newFlowRateAllowance
+        );
+
+        emit FlowOperatorUpdated(
+            token,
+            currentContext.msgSender,
+            flowOperator,
```
kasparkallas, line 850: Just a sanity check for myself: `flowOperator` (the address) and `flowOperatorId` (assuming something hashed together) are different things, right?

packages/ethereum-contracts/contracts/agreements/ConstantFlowAgreementV1.sol
```diff
@@ -813,6 +813,88 @@ contract ConstantFlowAgreementV1 is
...
+            oldPermissions,
+            newFlowRateAllowance
+        );
+    }
+
+    /// @dev IConstantFlowAgreementV1.decreaseFlowRateAllowance implementation
+    function decreaseFlowRateAllowance(
+        ISuperfluidToken token,
+        address flowOperator,
+        int96 flowRateAllowanceDelta, // flowRateBudget
```
kasparkallas: Again a `uint` would make more sense.

packages/ethereum-contracts/contracts/superfluid/SuperToken.sol
```diff
@@ -676,6 +676,28 @@ contract SuperToken is
...
+
+    function operationDecreaseAllowance(
+        address account,
+        address spender,
+        uint256 subtractedValue
+    ) 
+        external override
+        onlyHost
+    {
+        _approve(account, spender, _allowances[account][spender] - subtractedValue);
```
kasparkallas: The allowance might go negative here? That's not allowed by ERC-20.



Please consult: 
* https://github.com/OpenZeppelin/openzeppelin-contracts/blob/f8e3c375d19bd12f54222109dd0801c0e0b60dd2/contracts/token/ERC20/ERC20.sol#L197
* https://github.com/superfluid-finance/protocol-monorepo/blob/e7fe12f0ca6b860b98b338372d9282a8cb599005/packages/ethereum-contracts/contracts/superfluid/SuperToken.sol#L433 (can't we call `decreaseAllowance` instead of `_approve` and effectively re-implementing?)

packages/ethereum-contracts/contracts/agreements/ConstantFlowAgreementV1.sol
0xdavinchee: yeah, if you are changing your allowance by 0, it doesn't do anything.

packages/ethereum-contracts/contracts/agreements/ConstantFlowAgreementV1.sol
hellwolf: 1) change to substracted/addedFlwoRateAllowance makes ense.
2) should stick to int96 and consider it is our "custom type for flowrate" befor custom type is a thing in solidity; but of course continue to range check here before using it.

packages/ethereum-contracts/contracts/agreements/ConstantFlowAgreementV1.sol
0xdavinchee: yes, the `flowOperator` is the grantee of the permissions.
`flowOperatorId` is `keccak256(abi.encode("flowOperator", flowSender, flowOperator))`

packages/ethereum-contracts/contracts/agreements/ConstantFlowAgreementV1.sol
0xdavinchee: using uint48 would not be sufficient:
uint48 max value is: `281474976710655`
int96 max value is: `39614081257132168796771975167`.
We can 
I can rename the `delta` params though, it will remove any possible confusion given the int param

packages/ethereum-contracts/contracts/superfluid/SuperToken.sol
0xdavinchee: allowance can go negative in OZ's ERC-20 implementation because they have the subtraction in an `unchecked` block. we are using SafeMath and the SuperToken contract does not inherit from OZ's ERC20 implementation

kasparkallas: I'm still slightly confused why we effectively have 2 implementations for the same thing in the same file:
* https://github.com/superfluid-finance/protocol-monorepo/blob/e7fe12f0ca6b860b98b338372d9282a8cb599005/packages/ethereum-contracts/contracts/superfluid/SuperToken.sol#L433
* https://github.com/superfluid-finance/protocol-monorepo/blob/e7fe12f0ca6b860b98b338372d9282a8cb599005/packages/ethereum-contracts/contracts/superfluid/SuperToken.sol#L690
No way to re-use the logic?

kasparkallas: So it's expected that we have the 2 slightly different implementations in the same file?
image
image

0xdavinchee: nah, initially I removed safecast, but didn't change the new implementation to use it as well

packages/sdk-core/src/interfaces.ts
```diff
@@ -510,3 +510,20 @@ export interface IWeb3GovernanceParams {
...
+export interface ERC20IncreaseAllowanceParams extends EthersParams {
+    readonly spender: string;
+    readonly amount: string;
+}
+
+export type ERC20DecreaseAllowanceParams = ERC20IncreaseAllowanceParams;
+
+export interface SuperTokenFlowRateAllowanceParams extends EthersParams {
+    readonly flowOperator: string;
+    readonly flowRateAllowanceDelta: string;
```
kasparkallas, line 523: `BigNumberish` is the better type which allows for both `string` and `BigNumber` but I guess we've always used `string` and doesn't make sense to do it differently here.

Also, the SDK has still the "delta" naming... Not a big deal though.

kasparkallas (approved):

packages/sdk-core/src/interfaces.ts
0xdavinchee: a sweeping change might be a good idea, it wouldn't be breaking either

hellwolf (approved):

PR #1303, opened 2023-03-06
[WORKFLOW] SDK-Core/Subgraph CI Cleanup Fixes

0xdavinchee:
## Summary
Currently, we have a bunch of workflow files/tests which were added as defense for **publishing an SDK-Core version which is incompatible with the deployed subgraph**.

### Only acceptable reason for breaking
The only time a breaking change should occur is when the query in SDK-Core contains properties which don't exist in the deployed subgraph. 
> That is, we never want to push a change to the subgraph which breaks previous SDK-Core versions via modifying entities, etc.

## Proposed Solution
untitled@2x

The workflow files added previously have unintended consequences which seem to impact unrelated PRs. Feature and dev builds where no subgraph/SDK-core changes have been made should not break because the deployed subgraph for feature/dev/v1 are out of sync for that particular branch.

As a result, what we truly care about is:

 **_whenever we want to publish a new SDK-Core release, it should not break._** 

See above for the only reason it should be breaking.

We should only run these tests prior to a publish of SDK-Core or a deployment of a subgraph.

### package.json
- remove sdk-core from root package.json tests

### ci.feature
- remove eager testing in ci.feature

### ci.canary
- remove eager testing in ci.canary
- call.setup-deploy-and-test-local-subgraph.yml only tests local subgraphs, even schema generation
- consolidate `runQueryTests.ts` and `queryTests.ts`

### call.test-local-subgraph
- remove sdk-core integration tests

### call.deploy-subgraph
- run tests prior to deploying subgraph to ensure compatibility

### call.test-sdk-core
- always have local subgraph to test against

### cd.sdk-core-stable.create-release-draft
- only create release draft if tests pass


hellwolf:
image
How many "previous versions" we are talking about here?

0xdavinchee:
> image
> How many "previous versions" we are talking about here?

https://github.com/superfluid-finance/protocol-monorepo/blob/dev/.github/workflows/call.test-subgraph-on-previous-sdk-core-versions.yml#L21

Looks like this file is out of date though, but since `0.3.2`.

.github/workflows/call.deploy-subgraph.yml
```diff
@@ -28,10 +28,39 @@ on:
...
+      - name: Show contexts
+        run: |
+          echo github.event_name: ${{ github.event_name }}
+          echo github.sha: ${{ github.sha }}
+          echo github.repository: ${{ github.repository }}
+          echo github.ref: ${{ github.ref }}
+          echo github.head_ref: ${{ github.head_ref }}
+          echo github.base_ref: ${{ github.base_ref }}
+
+  build-and-test-local-subgraph:
```
hellwolf, line 46: I appreciate the meticulousness, but repeating "key" and "name" with the same information is not necessarily adding anything but noise. I think it's worth to be conscious of not adding syntactic noise too casually.

hellwolf: "build-and-test-local-subgraph:" is almost conflict with "name: Build and Test Subgraph (Release Branch)" on the surface, so not sure.

packages/ethereum-contracts/dev-scripts/run-deploy-contracts-and-token.js
```diff
@@ -1,23 +1,10 @@
-const {Framework} = require("@superfluid-finance/sdk-core");
 const fs = require("fs");
-const {ethers} = require("hardhat");
 const {deployContractsAndToken} = require("./deploy-contracts-and-token");
 
 deployContractsAndToken()
-    .then(async (deployer) => {
+    .then(async ({deployer, tokenDeploymentOutput}) => {
```
hellwolf, line 5: okay.

hellwolf (approved):
Looks good cleanup. I haven't necessarily reviewed in great details, but do not want to block it neither.

PR #1305, opened 2023-03-07
[WORKFLOWS] Canary fix 

0xdavinchee:
- Canary build still had sdk-core coverage generating a schema against dev
  - Refactor call.test-sdk-core to coverage test sdk-core and upload coverage report given input
- undo `handler.publish-pr-packages` changes with `ls` cause it was breaking stuff


hellwolf (approved):
let's see

PR #1311, opened 2023-03-09
[ETHEREUM-CONTRACTS] Cleanup 1.5.1 Release Fork Tests

0xdavinchee:
- Use `SuperTokenFactory` instead of `SuperTokenFactoryUpdateLogicContractsTester` for the test
- `batchUpdateSuperTokenLogic` ETHx in fork test as it is the token we are testing flow operations with


hellwolf (approved):
suspicious...

PR #1313, opened 2023-03-13
[ETHEREUM-CONTRACTS] bump solc to 0.8.19

0xdavinchee:
Bumping solc version to 0.8.19 as this is required by the SemanticMoney library.


hellwolf (approved):
wen solidity 0.9:

PR #1316, opened 2023-03-14
[SDK-core] Add missing distribution index Subgraph ID mapping to the index events

kasparkallas:
It was already queried but not mapped. "index.id" and "indexId" confusion most likely...


0xdavinchee (approved):
looks straightforward enough

PR #1320, opened 2023-03-19
Ida lib comments

saflamini:
Change comments in IDAv1Library and SuperTokenV1Library to correct the record on how distribute() works. 

cc @kobuta23 / @0xdavinchee - anything else you'd add or modify here?

Also note: I considered putting this in the IDA interface contract as well, but decided against it for now.
The ideal place for this IMO is clarity within our official IDA documentation (will work on this)


0xdavinchee (approved):
LGTM

PR #1323, opened 2023-03-21
Update Operation.ts

0xdavinchee:


packages/sdk-core/src/Operation.ts
```diff
@@ -46,45 +46,44 @@ export default class Operation {
...
-        const populatedTransaction = await this.getPopulatedTransactionRequest(
+        const signedTxn = await this.getSignedTransaction(
             signer,
             gasLimitMultiplier
         );
-        return await signer.sendTransaction(populatedTransaction);
+
+        signer._checkProvider("sendTransaction");
+        // @note we are explicitly stating the provider exists because of the above check
+        return await signer.provider!.sendTransaction(signedTxn);
```
kasparkallas: I don't think this will work with Gnosis Safe because their SDK has custom handling on `Signer.sendTransaction` which is different from `Provider.sendTransaction`: 
* https://github.com/safe-global/safe-core-sdk/blob/b2d2414dd7b085fe3b396cf8b61e4e197c3504c0/packages/safe-ethers-adapters/src/signer.ts#L102
* https://github.com/safe-global/safe-apps-sdk/blob/main/packages/safe-apps-provider/src/provider.ts#L81

kasparkallas (approved):

PR #1328, opened 2023-03-23
[WORKFLOWS] Fix deploy framework

0xdavinchee:
## Summary
- This PR fixes several ops-scripts for the new FlowNFT addition, namely the `deploy-framework.js`, `etherscan-verify.sh`, `info-print-contract-addresses.js`, `gov-upgrade-super-token-logic.js` scripts
- It also involves some cleanup in some other script files (the documentation) as well as some updates to the contract interface files


packages/ethereum-contracts/contracts/interfaces/superfluid/ISuperToken.sol
```diff
@@ -45,6 +45,11 @@ interface ISuperToken is ISuperfluidToken, TokenInfo, IERC20, IERC777 {
         string calldata s
     ) external;
 
+    /**************************************************************************
+    * Constants
+    *************************************************************************/
+    function SUPERFLUID_NFT_DEPLOYER_LIBRARY_ADDRESS() external view returns (address);
```
hellwolf: This should not be there, this is a detail that should not be part of the interface.

hellwolf: The one uses this is using "SuperToken" contract, so it's not a problem for the use sites.

packages/ethereum-contracts/contracts/interfaces/superfluid/ISuperTokenFactory.sol
```diff
@@ -24,6 +24,16 @@ interface ISuperTokenFactory {
...
 
+    /**************************************************************************
+    * Immutable Variables
+    **************************************************************************/
+
+    /**
+     * @notice Get the canonical super token logic.
+     * @dev Is the same as `getSuperTokenLogic`
+     */
+    function _SUPER_TOKEN_LOGIC() external view returns(ISuperToken superTokenLogic);
```
hellwolf: external & _SUPER_TOKEN_LOGIC? (the underscore)

if it's the same as `getSuperTokenLogic`, why?

packages/ethereum-contracts/ops-scripts/gov-upgrade-super-token-logic.js
```diff
@@ -1,8 +1,9 @@
 const async = require("async");
+const fs = require("fs");
 const SuperfluidSDK = require("@superfluid-finance/js-sdk");
 const {ZERO_ADDRESS} = require("./libs/common");
 
-const MAX_REQUESTS = 100;
+const MAX_REQUESTS = 20;
```
hellwolf, line 6: why the change?

packages/ethereum-contracts/contracts/interfaces/superfluid/ISuperToken.sol
0xdavinchee: will remove

packages/ethereum-contracts/contracts/interfaces/superfluid/ISuperTokenFactory.sol
0xdavinchee: fair, will also remove this one from interface

packages/ethereum-contracts/ops-scripts/gov-upgrade-super-token-logic.js
d10r: because a high number increases the chance of failed RPC requests, downside with minimal upside

hellwolf: makes sense

packages/ethereum-contracts/contracts/interfaces/superfluid/ISuperTokenFactory.sol
```diff
@@ -33,7 +37,8 @@ interface ISuperTokenFactory {
     function initialize() external;
 
     /**
-     * @dev Get the current super token logic used by the factory
+     * @notice Get the canonical super token logic.
+     * @dev Is the same as `_SUPER_TOKEN_LOGIC`
```
hellwolf: remove

```
* @dev Is the same as `_SUPER_TOKEN_LOGIC`
```

hellwolf (changes requested):
"disabled-test:deployment:scripts-js-truffle": "yarn run-truffle test test/ops-scripts/deployment.test.js",

Should this still be disabled?

0xdavinchee:
> ```
>     "disabled-test:deployment:scripts-js-truffle": "yarn run-truffle test test/ops-scripts/deployment.test.js",
> ```
> 
> Should this still be disabled?

I don't think so, not sure why this is disabled since we use truffle to deploy not HH.

hellwolf:
> > ```
> >     "disabled-test:deployment:scripts-js-truffle": "yarn run-truffle test test/ops-scripts/deployment.test.js",
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Should this still be disabled?
> 
> I don't think so, not sure why this is disabled since we use truffle to deploy not HH.

It was disabled because it was broken. check the git blame. If it works again, let's uncomment it.

packages/ethereum-contracts/contracts/interfaces/superfluid/ISuperTokenFactory.sol
0xdavinchee: addressed

0xdavinchee: addressed

hellwolf (approved):

PR #256, opened 2021-02-08
Feature 249 trusted forwarder integration

hellwolf:

PR #1329, opened 2023-03-24
Issue #256 | Try/Catch FlowNFT Hooks

0xdavinchee:
See the issue here: https://github.com/superfluid-finance/superfluid-stories/issues/256

## Summary of Changes
- This adds the try/catch back to the flow nft hooks in the CFA.
- Create a new internal library to encapsulate some of the logic when reverting out of gas in the catch block
- Uncomment foundry coverage sections in ci.canary to prevent breaking changes
- Add test to check if `updateCode` reverts when called by non-host for FlowNFTBase
- Use `getTokenId` instead of repeating token id logic in test code: `helper_Get_NFT_ID` (also for coverage)

### DIFF
--- a/.github/workflows/ci.canary.yml
+++ b/.github/workflows/ci.canary.yml
@@ -135,15 +135,15 @@ jobs:
           yarn build
           yarn workspace @superfluid-finance/ethereum-contracts test-coverage
 
-      # - name: Install lcov
-      #   run: sudo apt-get -y install lcov
+      - name: Install lcov
+        run: sudo apt-get -y install lcov
 
-      # - name: Clean up and merge coverage artifacts
-      #   run: |
-      #     # extract coverage for NFT contracts from forge coverage
-      #     lcov -e lcov.info -o lcov.info 'packages/ethereum-contracts/contracts/superfluid/*NFT*.sol'
-      #     # merge hardhat and forge coverage files
-      #     lcov -a lcov.info -a packages/ethereum-contracts/coverage/lcov.info -o packages/ethereum-contracts/coverage/lcov.info
+      - name: Clean up and merge coverage artifacts
+        run: |
+          # extract coverage for NFT contracts from forge coverage
+          lcov -e lcov.info -o lcov.info 'packages/ethereum-contracts/contracts/superfluid/*NFT*.sol'
+          # merge hardhat and forge coverage files
+          lcov -a lcov.info -a packages/ethereum-contracts/coverage/lcov.info -o packages/ethereum-contracts/coverage/lcov.info
 
       - name: Create coverage artifact
         uses: actions/upload-artifact@v3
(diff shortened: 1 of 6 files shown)


packages/ethereum-contracts/contracts/agreements/ConstantFlowAgreementV1.sol
```diff
@@ -470,12 +470,21 @@ contract ConstantFlowAgreementV1 is
...
+        try
+            ISuperToken(address(flowVars.token)).constantOutflowNFT().onCreate{
+                gas: CFA_HOOK_GAS_LIMIT
+            }(flowVars.sender, flowVars.receiver)
+        // solhint-disable-next-line no-empty-blocks
+        {} catch {
+// If the CFA hook actually runs out of gas, not just hitting the safety gas limit, we revert the whole transaction.
+// This solves an issue where the gas estimaton didn't provide enough gas by default for the CFA hook to succeed.
+// See https://medium.com/@wighawag/ethereum-the-concept-of-gas-and-its-dangers-28d0eb809bb2
+            if (gasleft() <= gasLeftBefore / 63) {
```
hellwolf: We should finally create a library function for this, e.g.

```
function _revertWhenOutOfGas() internal {
           if (gasleft() <= gasLeftBefore / 63) {
                revert CFA_HOOK_OUT_OF_GAS();
            }
}
```

hellwolf (changes requested):

packages/ethereum-contracts/contracts/agreements/ConstantFlowAgreementV1.sol
0xdavinchee: addressed, added a `SafeGasLibrary` lib, open to any new name suggestions

.github/workflows/ci.canary.yml
```diff
@@ -135,15 +135,15 @@ jobs:
...
-      # - name: Install lcov
-      #   run: sudo apt-get -y install lcov
-
-      # - name: Clean up and merge coverage artifacts
-      #   run: |
-      #     # extract coverage for NFT contracts from forge coverage
-      #     lcov -e lcov.info -o lcov.info 'packages/ethereum-contracts/contracts/superfluid/*NFT*.sol'
-      #     # merge hardhat and forge coverage files
-      #     lcov -a lcov.info -a packages/ethereum-contracts/coverage/lcov.info -o packages/ethereum-contracts/coverage/lcov.info
+      - name: Install lcov
```
hellwolf, line 138: will it work?

hellwolf (approved):

.github/workflows/ci.canary.yml
0xdavinchee: did another test here: https://github.com/superfluid-finance/protocol-monorepo/actions/runs/4531480256/jobs/7981669257
previously spent a long time trying to get it to work, but just wanted to double check in ci.feature above

PR #1333, opened 2023-03-28
[WORKFLOWS] Clean Up Subgraph Deployment Workflow Files

0xdavinchee:
## Initial Context
The deployment to satsuma endpoints failed because it was trying to deploy matic to eth-mainnet, xdai and arbitrum goerli, it is pretty neat that Satsuma was able to catch this on their end!
See: https://github.com/superfluid-finance/protocol-monorepo/actions/runs/4541456798/jobs/8003929210

## Fix
Same as the previous fixes, just prepare the manifest so that it is properly creating the `subgraph.yaml` file for the desired network.

## Extended Context
Currently, deploying to our production subgraphs requires two steps: deploying to v1 endpoints and deploying to Satsuma. It is often bottlenecked by reliance on me to click the correct buttons and pass in the correct inputs for the handler workflow.
There is added verbosity, but things are far more explicit now, which is good considering that it is no longer free to deploy subgraphs on some of our production networks (Satsuma).

## Changes
- moved prepare manifest down into the base shell scripts: `deploy-to-satsuma-network.sh` and `deploy-to-hosted-service-network.sh`

### Created
- split `call.deploy-subgraph.yml` into `call.deploy-hosted-service-subgraph.yml` and `call.deploy-satsuma-subgraph.yml`
- `handler.deploy-production-subgraphs.yml`
- `handler.deploy-satsuma-subgraph.yml`
- `handler.deploy-hosted-service-subgraph.yml`

### Renamed
- `deploy-all-networks.sh` => `deploy-all-hosted-service-networks.sh`

### Removed
- `handler.deploy-subgraph.yml`
- `deploy-to-hosted-service.sh`: this was encapsulating some logic


.github/workflows/call.deploy-hosted-service-subgraph.yml
```diff
@@ -0,0 +1,102 @@
...
+
+      - name: "Get ABI"
+        run: node scripts/getAbi.js
+        working-directory: ${{ env.subgraph-working-directory }}
+
+      - name: "Prepare subgraph manifest"
+        # This step is required for generating an arbitrary subgraph.yaml
+        # and addresses.ts file for the yarn codegen step below
+        # the actual addresses will be retrieved during the actual deployment of the subgraph
+        run: ./tasks/prepare-manifest.sh matic
```
kasparkallas: If it's an arbitrary placeholder then better to use dummy network's details, not real one's.

Can't just add a dummy network to `networks.json` though as they're considered implicitly for hosted service.

packages/subgraph/tasks/deploy-to-satsuma-network.sh
```diff
@@ -15,6 +15,12 @@
 
 graph="../../node_modules/@graphprotocol/graph-cli"
 
+# prepare the manifest prior to deployment
+# this generates the subgraph.yaml and
+# inputs the correct addresses for the specified network ($2)
+/tasks/prepare-manifest.sh $2
```
kasparkallas: No leading "./" is OK?

packages/subgraph/tasks/deploy-to-hosted-service-network.sh
```diff
@@ -4,6 +4,12 @@
 
 graph="../../node_modules/@graphprotocol/graph-cli"
 
+# prepare the manifest prior to deployment
+# this generates the subgraph.yaml and
+# inputs the correct addresses for the specified network ($2)
+/tasks/prepare-manifest.sh $2
```
kasparkallas: No leading "./" is OK?

packages/subgraph/tasks/deploy-to-satsuma-network.sh
0xdavinchee: good catch

packages/subgraph/tasks/deploy-to-hosted-service-network.sh
0xdavinchee: good catch

.github/workflows/call.deploy-hosted-service-subgraph.yml
0xdavinchee: fair, addressed as well

kasparkallas (approved):

packages/subgraph/tasks/deploy-all-hosted-service-networks.sh
```diff
@@ -0,0 +1,14 @@
+#!/usr/bin/env bash
+
+# args
+# $1 = the configuration (v1, dev, feature)
+
+JQ="../../node_modules/node-jq/bin/jq"
```
hellwolf: why not `npx jq`

packages/subgraph/tasks/deploy-all-satsuma-networks.sh
```diff
@@ -0,0 +1,11 @@
+#!/usr/bin/env bash
+
+# args
+# $1 = the version label
+
+# TODO: use jq and get the networks from metadata.json networks file
+SATSUMA_NETWORKS=( "matic" "xdai" "eth-mainnet" )
+
+for i in "${SATSUMA_NETWORKS[@]}";do
+    ./tasks/deploy-to-satsuma-network.sh $1 $i
```
hellwolf: shellcheck fails: "$1", "$i"

packages/subgraph/tasks/deploy-to-hosted-service-network.sh
```diff
@@ -4,6 +4,12 @@
 
 graph="../../node_modules/@graphprotocol/graph-cli"
 
+# prepare the manifest prior to deployment
+# this generates the subgraph.yaml and
+# inputs the correct addresses for the specified network ($2)
+./tasks/prepare-manifest.sh $2
```
hellwolf: ditto, would add a shellcheck lint as tech debt.

hellwolf (commented):
some changes required.

packages/subgraph/tasks/deploy-all-hosted-service-networks.sh
0xdavinchee: we currently have `node-jq` installed in our root `package.json`, I timeboxed 20 minutes to try and install specifically in `subgraph` and `sdk-core` packages and use `npx jq` there, but it didn't work and I think it's best we just leave it as is for now and revisit as tech debt.

packages/subgraph/tasks/deploy-all-satsuma-networks.sh
0xdavinchee: addressed

packages/subgraph/tasks/deploy-to-hosted-service-network.sh
0xdavinchee: I added a new `shellcheck-all-shell-scripts.sh` which checks all our bash scripts and it would run in `ci.lint-workflow-files.yml`.
It is currently commented out as it would cause shellcheck is not happy with a bunch of our shell scripts. This is just a brute force solution and can be tossed. 
The other solution is adding: `shell: bash` in all our `run:` commands in the different CI files.

packages/subgraph/tasks/deploy-all-hosted-service-networks.sh
hellwolf: JQ=`npx --package=node-jq jq --`

hellwolf (approved):

packages/subgraph/tasks/deploy-to-hosted-service-network.sh
hellwolf: Actually, this is not the best:

you could add this to the lint step of the package.json

hellwolf: And `find tasks $(find packages -name tasks) -name '*.sh'` is a bit more lean.

hellwolf: Let me fix this

hellwolf: one more workflow file is not necessary.

hellwolf (changes requested):
couple of changes needed to keep things leaner

packages/subgraph/tasks/deploy-to-hosted-service-network.sh
0xdavinchee: I agree, I'd rather not add another shell script file and I prose that we instead do a separate PR where we just do `shell: bash` for all `run:` in our workflow files.

hellwolf (approved):
@0xdavinchee feel free to review the shellcheck fixes.

0xdavinchee:
> @0xdavinchee feel free to review the shellcheck fixes.

thank you for doing this! 🙏🏼

PR #1340, opened 2023-03-29
[ETHEREUM-CONTRACTS] Flow NFT Fix

0xdavinchee:
## Summary
After debugging an issue with try/catch surfaced by the [SuperSpreader](https://github.com/superfluid-finance/superfluid-stories/issues/258), we discovered there was a vulnerability in the FlowNFT integration which would have allowed malicious SuperToken contracts to arbitrarily mint NFTs that belong to another SuperToken. 
> NOTE: This was not something that would've impacted any part of the existing protocol, but wouldn't have allowed proper token gating using the NFT streams.

The initial fix was for some odd try/catch behavior in solidity, you can do multiple external calls, but only the first one is checked **if** the next call is attempting to be executed from `address(0)`.

## Changes
- The fix for the bug is passing in the `flowVars.token` to the hooks. The NFT contracts are already linked to their respective SuperToken, so we revert if the `flowVars.token` is not equal to the SuperToken set for the NFT during initialization.
- Added a fix for the try catch: low level call is done to ensure the call is successful, if it is, we decode the data retrieved and we use this to determine whether or not to execute the hook. (had to use a different pattern because of stack too deep).
- Cleaned up the interfaces a bit
- Added tests to ensure these fixes are addressed


0xdavinchee:
Closing this PR in favor of the `flow-nft-refactor` branch which will include some of the applicable fixes here:
- low level staticcall to NFT contract
- try/catch

PR #1343, opened 2023-03-30
[SDK-redux] Small type fix

kasparkallas:
It completely overlaps with the previous but is a little less strict on what fields to enforce.

Background: This initially came up with wagmi interop as wagmi prefers to send transactions "unchecked" which means that they skip an ethers.js follow-up RPC call to fill in more data about the transaction: https://docs.ethers.org/v5/api/providers/jsonrpc-provider/#JsonRpcSigner-connectUnchecked


0xdavinchee (approved):
👍🏼

hellwolf (approved):

PR #1348, opened 2023-04-05
[WORKFLOW/ETHEREUM-CONTRACTS] Fix Up Coverage

0xdavinchee:
## Summary
The ethereum-contracts test coverage has "dropped" to around 85% on codecov, however, this is largely due to several factors:
- new test files written specifically for foundry 
- new foundry test code coverage results not being merged and uploaded to codecov
- removal of previous test files
- commented out `lcov --add-tracefile` in `ci.canary`.

## Changes
- Ignore test contracts and other contracts as part of coverage
- Add some tests for BatchLiquidator/SuperAppBaseCFA foundry
- Recover CFAv1Library and IDAv1Library tests
- Add missing SuperTokenV1Library tests (w/ userData)
- Add missing SuperfluidGovernanceII test (setPPPConfig with invalid liq/pat period)
- extract more foundry specific tests
- fix typo in SuperToken.sol
- _revertWhenOutOfGas can and should be a view function
  - Add `_revertWhenOutOfGas` prop test


hellwolf:
```
codecov:
  require_ci_to_pass: yes
```

Does this prevent us from having the result if canary is red (which seems a common case atm) ?

packages/ethereum-contracts/.solcover.js
```diff
@@ -5,6 +5,11 @@ module.exports = {
     skipFiles: [
         "mocks/",
         "apps/SuperAppBase.sol",
+        "utils/SuperfluidFrameworkDeployer",
```
hellwolf, line 16: Is SuperfluidFrameworkDeployer really a good candidate to skip?

Same goes for SuperTokenDeployer

packages/ethereum-contracts/.solcover.js
```diff
@@ -5,6 +5,11 @@ module.exports = {
     skipFiles: [
         "mocks/",
         "apps/SuperAppBase.sol",
```
hellwolf, line 11: what's the role of SuperAppBase nowadays

0xdavinchee:
> ```
> codecov:
>   require_ci_to_pass: yes
> ```
> 
> Does this prevent us from having the result if canary is red (which seems a common case atm) ?

Yeah, I believe so.

packages/ethereum-contracts/.solcover.js
0xdavinchee: It depends, we don't have an explicit definition of what files should be included in our coverage.
In my opinion, we should reduce noise as much as possible and it should solely be the core contracts which are deployed in production.
These are both peripheral contracts that are not used in production and we do already have tests for them.

packages/ethereum-contracts/.solcover.js
0xdavinchee: I believe it is used by the `SuperAppBaseCFA` and it's usually overridden, we could create two simple SuperApp's  which inherits from the respective bases and override the functions, but just use super.function, although not sure if what we are testing here is interesting.

hellwolf:
installed codecov to github for the org.

packages/ethereum-contracts/test/foundry/libs/SafeGasLibrary.prop.t.sol
```diff
@@ -0,0 +1,24 @@
...
+
+/// @title SafeGasLibraryProperties
+/// @author Superfluid
+/// @notice This contract tests that the SafeGasLibrary behaves as expected.
+/// @dev If the gasleft() inside the function is less than 1/63 of the gasleft() before the function call,
+/// then the function ran out of gas and we revert the whole transaction.
+/// We mock this here by using a multiple of the gasleft() before the function call.
+contract SafeGasLibraryProperties is Test {
+
+    function test_SafeGasRevertsAsExpected(uint16 multiple) public {
```
hellwolf: This is a bad test, you are testing exactly how it is coded so you are not actually testing anything useful.

But what you really want to test is the property that: when you give an arbitrary gas limit to a call and if it reverts OUT_OF_GAS it is less than 63*.

packages/ethereum-contracts/.solcover.js
hellwolf: I thought they are used in some of the test cases, hence they should have been covered. I guess my assumption was wrong?

packages/ethereum-contracts/test/foundry/libs/SafeGasLibrary.prop.t.sol
0xdavinchee: removed 😅

hellwolf (approved):
Thank you! This is great. Back to 98%+

PR #1375, opened 2023-04-26
[ETHEREUM-CONTRACTS] use internal host

0xdavinchee:
## Summary
I noticed that in the CFAv1 we are relying on an external call to getHost() on the SuperToken which is unnecessary given that all agreement's inherit from `AgreementBase` and set the host on initialization.
This mainly saves some gas.


hellwolf (approved):
Okay, let's go with this.

Original intent was that this logic contract could conceptually reused by any "host". But that ship had long sailed.

PR #1377, opened 2023-04-26
[ETHEREUM-CONTRACTS] Add simple test benchmarking

0xdavinchee:
## Summary
To give us a better idea of where the bottlenecks exist in our testing suite and provides a less yolo approach in addressing them.
- This PR records the time each test case takes (in milliseconds) and sorts them by time taken in the final after hook
- It spits them out into a json file and in the CI it will `cat` the file.
- It uses the built-in `performance` library


hellwolf:
Can you do a bit more investigation of hardhat / mocha integration?

These sort of stuff is natively provided by the test framework such as mocha.

Also in node20, I heard that mocha like test primitives are natively supported, so those test stats output should also be done probably differently.

hellwolf:
"reporters" https://mochajs.org/#reporters

hellwolf:
"reporters" https://mochajs.org/#reporters

hellwolf:
is this okay now?

hellwolf (approved):

PR #1380, opened 2023-04-28
[ETHEREUM-CONTRACTS] Remove slowest tests

0xdavinchee:
## Summary
This PR just tackles the low hanging fruit and removes the slowest tests in our existing testing suite.


packages/ethereum-contracts/test/foundry/superfluid/Superfluid.t.sol
```diff
@@ -0,0 +1,64 @@
...
+} from "../../../contracts/interfaces/superfluid/ISuperAgreement.sol";
+import {
+    ISuperfluid
+} from "../../../contracts/interfaces/superfluid/ISuperfluid.sol";
+import { AgreementMock } from "../../../contracts/mocks/AgreementMock.sol";
+
+contract SuperfluidIntegrationTest is FoundrySuperfluidTester {
+    using SuperTokenV1Library for SuperToken;
+
+    uint32 private constant _MAX_NUM_AGREEMENTS = 256;
```
hellwolf: is this not a constant that's exported from Superfluid host contract?

hellwolf: >         if (_agreementClasses.length >= 256) revert HOST_MAX_256_AGREEMENTS();

Okay, but it's better if we improve the host contract in the meantime to make this a constant in it.

0xdavinchee: it is not: https://github.com/superfluid-finance/protocol-monorepo/blob/dev/packages/ethereum-contracts/contracts/superfluid/Superfluid.sol#L155

packages/ethereum-contracts/test/foundry/superfluid/Superfluid.t.sol
```diff
@@ -0,0 +1,64 @@
...
+    uint32 private constant _NUM_AGREEMENTS = 2;
```
hellwolf, line 20: a bit fragile logic, when GDA included it would be 3, it seems the test would fail that's good (but if not, then it could become a dummy test case)

packages/ethereum-contracts/test/foundry/superfluid/Superfluid.t.sol
0xdavinchee: In the IDA, it is a constant too, but it's private, maybe it should be public?

packages/ethereum-contracts/test/foundry/superfluid/Superfluid.t.sol
hellwolf: But indeed `_agreementClasses` is current not exposing .length or anything, something to think about if you want to be perfectionist here.

0xdavinchee: it will break, the original test broke in the GDA branch when running this

packages/ethereum-contracts/test/contracts/superfluid/Superfluid.test.ts
```diff
@@ -337,48 +337,7 @@ describe("Superfluid Host Contract", function () {
...
-                        superfluid.address,
-                        badMock.address
-                    ),
-                    superfluid,
-                    "HOST_MAX_256_AGREEMENTS"
-                );
-            });
-
-            it("#2.6 agreement must be registered first", async () => {
+            it("#2.5 agreement must be registered first", async () => {
```
hellwolf, line 342: maybe a comment // #2.5 moved to foundry test case path blah blah?

hellwolf (changes requested):
great stuff, some change request.

packages/ethereum-contracts/test/contracts/superfluid/Superfluid.test.ts
0xdavinchee: addressed the changes above

packages/ethereum-contracts/contracts/agreements/InstantDistributionAgreementV1.sol
```diff
@@ -74,15 +74,16 @@ contract InstantDistributionAgreementV1 is
 
     address public constant SLOTS_BITMAP_LIBRARY_ADDRESS = address(SlotsBitmapLibrary);
 
+    /// @dev Maximum number of subscriptions a subscriber can have
+    uint32 public constant MAX_NUM_SUBSCRIPTIONS = 256;
```
hellwolf: `SlotsBitmapLibrary._MAX_NUM_SLOTS`

hellwolf (approved):
one more change then go ahead and merge.

PR #1387, opened 2023-05-03
[SUBGRAPH] Entity FlowOperator - New field "Allowance" Added

### LINKED ISSUE #1382: [SUBGRAPH] Add a new entity aggregating ERC-20 Allowance and Superfluid ACL for Super Tokens
# Improve Readability of Super Tokens Access Settings

## Background
Currently, on-chain security access for Super Tokens is managed using [ERC-20 Allowance](https://academy.bit2me.com/en/what-is-token-allowance/) and [Superfluid ACL](https://docs.superfluid.finance/superfluid/developers/constant-flow-agreement-cfa/cfa-access-control-list-acl).

## Problem
Efficiently inspecting the latest state of access settings for Super Tokens is challenging, as it involves making multiple RPC calls to different endpoints and aggregating the latest state based on numerous events. This process is cumbersome and impractical when displaying a list in the UI. Potentially for multiple networks at the same time.

## Proposed Solution
To simplify and streamline this process, we suggest implementing Subgraph indexing, which will make querying the necessary data much more efficient.

### Example Schema
```graphql
query AccountAccessSettings {
    flowOperators(where: { sender: "0x..." }) {
        token
        operator
        allowance
        flowRateAllowanceRemaining
        permissions
    }
}
```
NOTE: This returns an array of entities.

### Events Involved
* Approve (ERC-20)  <-- Might create the entity.
* ~~Transfer (ERC-20)~~ <-- Transfer event does not need special handling because ` transferFrom` always fires Approve event as well.
* FlowOperatorUpdated (CFA) <-- Might create the entity.
* FlowUpdated (CFA)

### Entity ID
{token}-{account}-{operator}

## Other Information
* For ERC-20 allowance, [Revoke.cash](https://revoke.cash/) is commonly used.
* Subgraph already has `FlowOperator` entity which we can use.
* Might consider renaming "FlowOperator" to something else to clarify that it's not just ACL Flow Operator related entity anymore.

msoni89:


packages/subgraph/src/mappings/cfav1.ts
```diff
@@ -202,11 +203,20 @@ export function handleFlowOperatorUpdated(event: FlowOperatorUpdated): void {
         event.params.token,
         event.params.sender
     );
-
     flowOperator.permissions = event.params.permissions;
     flowOperator.flowRateAllowanceGranted = event.params.flowRateAllowance;
     flowOperator.flowRateAllowanceRemaining = event.params.flowRateAllowance;
     flowOperator.flowOperator = event.params.flowOperator;
+
```
msoni89: > Not sure, do we need to call (RPC) allowance here or not.

0xdavinchee: this is not necessary with an RPC call, the handleApproved mapping function should catch all cases of approval amounts changing

packages/subgraph/test/validation/hol/flowOperatorValidator.ts
```diff
@@ -65,6 +65,10 @@ const validateFlowOperator = (
         flowOperatorEntity.flowRateAllowanceGranted,
         "Flow Operator: flowRateAllowanceGranted error"
     ).to.equal(expectedFlowOperator.flowRateAllowanceGranted);
+    expect(
```
0xdavinchee: please remove this, we ideally want to remove code from `packages/subgraph/test`

packages/subgraph/test/interfaces.ts
```diff
@@ -242,6 +242,7 @@ export interface IFlowOperator extends IBaseEntity {
     readonly flowRateAllowanceGranted: string;
     readonly flowRateAllowanceRemaining: string;
     readonly flowOperator: string;
+    readonly allowance: string;
```
0xdavinchee: please remove this, we ideally want to remove code from `packages/subgraph/test`

packages/subgraph/test/helpers/initializers.ts
```diff
@@ -81,6 +81,7 @@ export const getOrInitFlowOperator = (
             updatedAtBlockNumber,
             permissions: 0,
             flowRateAllowanceGranted: "0",
+            allowance: "0",
```
0xdavinchee: please remove this, we ideally want to remove code from `packages/subgraph/test`

packages/subgraph/src/mappings/superToken.ts
```diff
@@ -447,3 +450,14 @@ function _createTransferEventEntity(event: Transfer): void {
     ev.token = event.address;
     ev.save();
 }
+
+export function handleApproval(event: Approval): void {
+    const flowOperatorID = getFlowOperatorID(event.params.spender, event.address, event.params.owner);
```
0xdavinchee, line 455: I would call `getOrInitFlowOperator` here and also create a new `ApprovalEvent` entity maybe.

packages/subgraph/subgraph.template.yaml
```diff
@@ -288,6 +288,9 @@ templates:
         - event: Transfer(indexed address,indexed address,uint256)
           handler: handleTransfer
           receipt: true
+        - event: Approval(indexed address,indexed address,uint256)
```
0xdavinchee, line 292: Please add the new entities to be added above under `mapping.entities`

packages/subgraph/tests/cfav1/hol/cfav1.hol.test.ts
```diff
@@ -88,7 +91,15 @@ describe("ConstantFlowAgreementV1 Higher Order Level Entity Unit Tests", () => {
 
         handleFlowOperatorUpdated(flowOperatorUpdatedEvent);
 
-        
+        const approvalEvent = createApprovalEvent(
```
0xdavinchee: Please remove this from here and put it in its own unit test in `supertoken/hol/supertoken.hol.test.ts`

packages/subgraph/src/mappings/superToken.ts
0xdavinchee: disregard given that this allowance field is only for flow operators

packages/subgraph/src/mappings/cfav1.ts
msoni89: Moved into getOrInitFlowOperator, it will only be called when the entry does not exist. Previously, it was being called every time the user updated their permission or flow allowance.

packages/subgraph/src/mappings/cfav1.ts
```diff
@@ -1,4 +1,4 @@
-import { BigInt } from "@graphprotocol/graph-ts";
+import { Address, BigInt } from "@graphprotocol/graph-ts";
```
0xdavinchee: last thing: this looks unused now?

0xdavinchee (commented):
Looks good, great stuff! 👍🏼  One last minor thing.

packages/subgraph/src/mappings/cfav1.ts
0xdavinchee: nice!

0xdavinchee (approved):
great stuff! 🙌🏼

packages/subgraph/src/mappings/superToken.ts
kasparkallas, line 455: What if flow operator entity doesn't exist yet?

packages/subgraph/src/mappings/superToken.ts
```diff
@@ -447,3 +450,14 @@ function _createTransferEventEntity(event: Transfer): void {
...
+    const flowOperator = FlowOperator.load(flowOperatorID);
+    if (!flowOperator) return;
```
kasparkallas, line 457: Well, now the information is incomplete when querying to find out the ERC-20 allowances. Can happen if there wasn't a ACL event before.

packages/subgraph/src/mappingHelpers.ts
```diff
@@ -338,6 +340,16 @@ export function getOrInitFlowOperator(
         flowOperatorEntity.flowOperator = flowOperatorAddress;
         flowOperatorEntity.updatedAtBlockNumber = block.number;
         flowOperatorEntity.updatedAtTimestamp = currentTimestamp;
+
+        // It will only be called when the entry does not exist. Previously, it was being called every time the user updated their permission or flow allowance
+        const superTokenContract = ISuperToken.bind(
+            tokenAddress
+        );
+        const currentAllowance = superTokenContract.try_allowance(senderAddress, flowOperatorAddress);
```
kasparkallas, line 348: RPC call added. Would be nice to avoid these.

kasparkallas (commented):
I can't see the handling of Transfer to reduce the allowance?

EDIT: Allowance is fired on Transfer.

packages/subgraph/src/mappingHelpers.ts
msoni89: This is for the first event when the user doesn't exist.  What if someone gives allowance before ACL permission then because of handleApprove implementation if ignore `if (!flowOperator) return` and maybe we will see the `allowance` discrepancy.

packages/subgraph/src/mappings/superToken.ts
msoni89: As we discussed it's only for vesting/flow allowances given (when transfer cliff amount). So we can't store all approval as FlowOperators that's why there is a check if the flow operator exists then only update the `allowance`.

if there wasn't an ACL event before -  it will skip here and handle allowance in `if(!getOrInitFlowOperator) and update its correct state into the entity.

packages/subgraph/src/mappings/superToken.ts
msoni89: So here we are updating the `flowOperator.allowance` of the super token given to the flow operator, if not exist it simply ignores the event and doesn't store anything. 

in the next line, there is if the case. `if (!flowOperator) return;`

packages/subgraph/src/mappings/superToken.ts
kasparkallas: I don't recall discussing having this only for vesting/flow allowances.

I'd say the entity should always be created when ERC-20 allowance is given. This might warrant an entity name change for more clarity (hence the talk about "access settings" in the issue).

Then the entity will have full clarity over both ERC-20 allowances and ACL permissions & allowances. We can also throw away the initial RPC call then.

PR #1393, opened 2023-05-08
[SUBGRAPH] rename networks.json to hosted-service-networks

0xdavinchee:
## Problem

Sepolia doesn't have hosted service subgraph support. This is currently breaking builds (daily check and SDK-release).

## Solution

A quick solution is to remove eth-sepolia from the `networks.json` file.
In addition, we rename `networks.json` to `hosted-service-networks.json` to differentiate it from the multiple subgraph sources we currently support and will support down the line:
- hosted service
- satsuma
- graph studio (staging)
- decentralized subgraph
- self-hosted


kasparkallas (approved):
Note: Sepolia subgraph compatibility is effectively not tested right now.

PR #1403, opened 2023-05-11
Refactor auto-wrap and scheduler tests to use FoundrySuperfluidTester

### LINKED ISSUE #1395: [ETHEREUM-CONTRACTS] Refactor SuperfluidTester.sol to reuse FoundrySuperfluidTester.sol
## Summary

The `SuperfluidTester.sol` contracts include logic which is already utilized in `FoundrySuperfluidTester.sol`. Most of the assumptions are the same:
- deploy the framework
- deploy a token
- upgrade some tokens for some users
- start testing

The difference in the `SuperfluidTester.sol` files in `autowrap` and `scheduler` is that these contracts also deploy the contracts to be tested: `Manager`, `WrapStrategy`, etc. 

This deployment occurs in: `setUp`, a foundry specific function which is called before each test, this function is virtual which means you can use the underlying implementation and add more functionality in an inheriting test contract:
```
contract TestContract is FoundrySuperfluidTester {
function setUp() public {
  super.setUp();
  // logic for deploying the tested contracts for example
}
}
```

## Final outcome
- [ ] Fewer LOC
- [x] SuperfluidTester.sol tests should be removed
- [x] Most of the setUp code in the other test files (.t.sol) should be removed (token minting + upgrading)
- [x] FoundrySuperfluidTester is used in place of SuperfluidTester.sol
- [x] Tests passing

elvijsTDL:


0xdavinchee (approved):
LGTM

vmichalik:
/// @notice Look at me , I am the captain now - Elvijs
this was a test and you failed @0xdavinchee

PR #1417, opened 2023-05-19
[SDK-core] make BatchCall convertible to Operation

kasparkallas:
Experimented with this the other day.

Why?
* Requested by community member to retrieve _calldata_ of a batch call. As a solution, instead of adding logic for this on the `BatchCall`, convert it to `Operation` and use the logic from there.
* The gas estimation multiplication is currently not applied to `BatchCall` because of an oversight caused by duplication of code in `BatchCall` and `Operation`. Solve it by converting `BatchCall` to `Operation` and run through the same logic path.

Kept it as a non-breaking change.


hellwolf:
is this ready for review?

kasparkallas:
> is this ready for review?

I'd say so.

@0xdavinchee check it out

0xdavinchee (approved):
looks good - LGTM

hellwolf:
need to fix the build error

PR #1420, opened 2023-05-22
[SUBGRAPH] config files with canonical name, xdai -> gnosis

d10r:
xdai is now gnosis: https://github.com/graphprotocol/graph-tooling/blob/main/packages/cli/src/protocols/index.ts#L85

canonical config file names would help with automation of self-hosted subgraphs.
Would that break anything else?


0xdavinchee:
gm - in its current state, yes, it would break a few things, check out `testSchemasAndQueries.sh`, `hosted-service-networks.json` and `deploy-all-hosted-service-networks.sh`.

d10r:
Added symlinks with old names for backwards compatibility, that passes tests.
@0xdavinchee do you think that's safe?

As for the renaming of xdai -> gnosis I'm a bit confused. 
In https://thegraph.com/docs/en/developing/supported-networks/ it says gnosis is the new name, but xdai should still work.
However when trying to deploy a subgraph locally, that seems not to be the case:
```
Deploying to Graph node http://localhost:27021/UNCAUGHT EXCEPTION: Error: Failed to deploy to Graph node http://localhost:27021/: network not supported by registrar: no network xdai found on chain ethereum
```

Thus the renaming.
I guess that's safe?!

0xdavinchee (approved):
LGTM

PR #1427, opened 2023-05-24
[ETHEREUM-CONTRACTS] Foundry test framework

0xdavinchee:
## Summary 
This PR is the first implementation of the test plan. More specifically, the test framework rewritten in Foundry which can be used to write tests moving forward as well as replacing some of the existing tests.

## In practice
To see how the test framework would change the way we write our tests currently, look at `ConstantFlowAgreementV1.t.sol` and `InstantDistributionAgreementV1.t.sol`.

The helper functions do a few things:
1. the actual action, e.g. create flow, update sub units, etc.
2. abstract away the assertions that should be checked after different actions, e.g. flow is created, units are updated, etc.
3. update test state so we can assert global invariants and validate other things, e.g. take balance snapshots, set the last updated sub index value to get the diff between the index's index value and the sub's index value (as this is not exposed)

A test scenario should be named with a sufficiently descriptive function name, we want to ensure everything holds in this scenario.

## New Features
- The FoundrySuperfluidTester contract will enable testing with Native Asset SuperToken, Pure SuperToken, Custom Wrapper SuperToken and Custom Pure SuperToken's in addition to Wrapper SuperToken's this can be dictated by setting: `TOKEN_TYPE` env variable.
  - This can eventually be used as the base to be packaged and given to external parties to test their custom super token's
- New invariants added to be verified
- Helper functions for CFA, IDA in FoundrySuperfluidTester

## Changes
- The existing tests have been modified to use the test framework where applicable and reasonable


packages/ethereum-contracts/test/foundry/FoundrySuperfluidTester.sol
```diff
@@ -82,136 +166,1170 @@ contract FoundrySuperfluidTester is Test {
...
+        ISuperToken superToken_,
+        address sender,
+        address flowOperator,
+        int96 expectedFlowRateAllowance,
+        uint8 expectedPermissionsBitmask
+    ) internal {
+        (bool canCreate, bool canUpdate, bool canDelete, int96 allowance) =
+            superToken_.getFlowPermissions(sender, flowOperator);
+
+        bool expectedAllowCreate = expectedPermissionsBitmask & 1 == 1 ? true : false;
```
0xdavinchee: `? true : false` not necessary

packages/ethereum-contracts/test/foundry/agreements/InstantDistributionAgreementV1.t.sol
```diff
@@ -2,11 +2,11 @@
 pragma solidity 0.8.19;
 
 import "../FoundrySuperfluidTester.sol";
-import { SuperToken } from "../../../contracts/superfluid/SuperToken.sol";
+import { ISuperToken, SuperToken } from "../../../contracts/superfluid/SuperToken.sol";
```
hellwolf: SuperToken is no longer referenced

packages/ethereum-contracts/test/foundry/FoundrySuperfluidTester.sol
```diff
@@ -65,13 +128,35 @@ contract FoundrySuperfluidTester is Test {
...
-        resolver = sf.resolver;
-
         require(nTesters <= TEST_ACCOUNTS.length, "too many testers");
         N_TESTERS = nTesters;
+
+        // Set the token type being tested
+        string memory tokenType = vm.envOr(TOKEN_TYPE_ENV_KEY, DEFAULT_TEST_TOKEN_TYPE);
+        bytes32 hashedTokenType = keccak256(abi.encode(tokenType));
+
+        testSuperTokenType = hashedTokenType == keccak256(abi.encode("WRAPPER_SUPER_TOKEN"))
```
hellwolf, line 149: hurts my eyes :) could have just used a 'if else train' too

packages/ethereum-contracts/test/foundry/FoundrySuperfluidTester.sol
```diff
@@ -82,136 +167,1290 @@ contract FoundrySuperfluidTester is Test {
...
+        (superToken) = sfDeployer.deployPureSuperToken("Super MR", "MRx", initialSupply);
+        _expectedTotalSupply = initialSupply;
+        for (uint256 i = 0; i < N_TESTERS; ++i) {
+            superToken.transfer(TEST_ACCOUNTS[i], INIT_SUPER_TOKEN_BALANCE);
+            _helperTakeBalanceSnapshot(superToken, TEST_ACCOUNTS[i]);
+        }
+    }
+
+    /// @notice Deploys a SuperToken based on the testSuperTokenType selected
+    function _setUpSuperToken() internal {
```
hellwolf, line 214: okay with this ergonomics (using raw int numbers)? how is the use sites look like?

alokeutpal:
Nice 👍

hellwolf (approved):
some minor comments, but let's go!

packages/ethereum-contracts/test/foundry/agreements/InstantDistributionAgreementV1.t.sol
0xdavinchee: removed

packages/ethereum-contracts/test/foundry/FoundrySuperfluidTester.sol
0xdavinchee: true, the ternary does seem a bit aggressive in this case 😅

0xdavinchee: unfortunately a necessity due to: "cannot initialize immutable in if block"

packages/ethereum-contracts/test/foundry/FoundrySuperfluidTester.sol
0xdavinchee: using enums instead

PR #1428, opened 2023-05-24
[ETHEREUM-CONTRACTS] Add Increase/DecreaseFlowRateAllowance to SuperTokenV1Library

0xdavinchee:
- add cache warm up
- add increase flow rate allowance
- add decrease flow rate allowance


packages/ethereum-contracts/contracts/apps/SuperTokenV1Library.sol
```diff
@@ -22,6 +22,16 @@ import {
  */
 library SuperTokenV1Library {
 
+    /**
+     * @notice Warms up the cache: get the host, cfa and ida from the token
+     * @dev Fixes the issue where expect revert fails because it expects this call to revert
+     * @param token The token used in the flow
+     */
+    function warmUpCache(ISuperToken token) external {
```
hellwolf: This code smells,  please explain more.

packages/ethereum-contracts/contracts/apps/SuperTokenV1Library.sol
```diff
@@ -22,6 +22,16 @@ import {
...
+        _getAndCacheHostAndCFA(token);
+        _getAndCacheHostAndIDA(token);
+    }
```
hellwolf: Why does Codecov think the new code is not covered by the coverage?

packages/ethereum-contracts/test/foundry/FoundrySuperfluidTester.sol
```diff
@@ -92,6 +92,7 @@ contract FoundrySuperfluidTester is Test {
             _expectedTotalSupply += INIT_SUPER_TOKEN_BALANCE;
             vm.stopPrank();
         }
+        superToken.warmUpCache();
```
hellwolf: Why should this be called by test code instead? If

hellwolf (changes requested):

packages/ethereum-contracts/contracts/apps/SuperTokenV1Library.sol
0xdavinchee: we need to use a mock contract to test this and all new functions on the library out

packages/ethereum-contracts/test/foundry/FoundrySuperfluidTester.sol
0xdavinchee: not sure where this would be called instead tbh

packages/ethereum-contracts/contracts/apps/SuperTokenV1Library.sol
0xdavinchee: If we look at this test:
```
    function testRevertIfDecreaseFlowRateAllowanceUnderflows() public {
        vm.startPrank(alice);
        vm.expectRevert(IConstantFlowAgreementV1.CFA_ACL_NO_NEGATIVE_ALLOWANCE.selector);
        superToken.decreaseFlowRateAllowance(bob, 10);
        vm.stopPrank();
    }
```
it will fail if we do not warm up the cache with an error stating that no revert happened. this is because you are expecting the next call to revert and the next call is to: `getHost` inside of `_getHostAndCFA`.
if we remove the revert, it also fails, with the custom error.

the warming up of the cache is to prevent this. 
an alternative is to just execute a view function before and write a comment about this.

hellwolf: I would prefer we don't have a special function in production code just for testing purpose. Instead, calling another function that incidentally warms up the catch sounds a clean workaround to me, so long as a comment in the test code explaining this "intended incidental warm up"

packages/ethereum-contracts/test/foundry/FoundrySuperfluidTester.sol
hellwolf: Instead of doing it for all tests, do the "intended warm up" code at the places where it's needed explicitly

packages/ethereum-contracts/contracts/apps/SuperTokenV1Library.sol
hellwolf: e.g. create a flow just for testing

d10r:
@0xdavinchee could you also comment on where the requirement to add delta fns for flowrate allowance came from / link to related conversations?

0xdavinchee:
> @0xdavinchee could you also comment on where the requirement to add delta fns for flowrate allowance came from / link to related conversations?

there was no hard requirement, but I added a TODO to add this to the library a while ago in one of the foundry tests-from my point of view, we generally want the library to have full feature parity with the cfa/ida/gda so that people won't have to use the library AND the cfa. 

this came up as I was working through the test framework.

do you have objections on adding this?

cc @d10r

hellwolf (approved):

packages/ethereum-contracts/contracts/apps/SuperTokenV1Library.sol
```diff
@@ -231,6 +205,82 @@ library SuperTokenV1Library {
         return true;
     }
 
+    /**
+     * @dev Increases the flow rate allowance for flow operator
+     * @notice allowing userData to be a parameter here triggered stack to deep error
```
d10r: stack "too" deep, not "to"

packages/ethereum-contracts/contracts/apps/SuperTokenV1Library.sol
```diff
@@ -231,6 +205,82 @@ library SuperTokenV1Library {
...
+    function increaseFlowRateAllowance(ISuperToken token, address flowOperator, int96 addedFlowRateAllowance)
+        internal
+        returns (bool)
+    {
+        return increaseFlowRateAllowance(token, flowOperator, addedFlowRateAllowance, new bytes(0));
+    }
+
+    /**
+     * @dev Increases the flow rate allowance for flow operator
+     * @notice allowing userData to be a parameter here triggered stack to deep error
```
d10r: same

packages/ethereum-contracts/contracts/apps/SuperTokenV1Library.sol
```diff
@@ -1323,22 +1247,7 @@ library SuperTokenV1Library {
...
-                    publisher,
-                    indexId,
-                    subscriber,
-                    new bytes(0) // ctx placeholder
-                )
-            ),
-            "0x"
-        );
-        return true;
+        return claim(token, publisher, indexId, subscriber);
```
d10r: new bytes(0) arg missing?

d10r (dismissed):
see comments

good refactoring! wasn't aware we have so much code repetition here

packages/ethereum-contracts/contracts/apps/SuperTokenV1Library.sol
0xdavinchee: yes, good catch

packages/ethereum-contracts/contracts/apps/SuperTokenV1Library.sol
0xdavinchee: resolved for all

PR #1446, opened 2023-06-01
[ETHEREUM-CONTRACTS] Fixing issue in `deleteWrapScheduleByIndex` function

### LINKED ISSUE #1441: [ETHEREUM-CONTRACTS] - Auto Wrapper
Hi @msoni89,

Thanks for your observation.

When you call the deleteWrapScheduleByIndex function with an index that doesn't correspond to an existing wrap schedule, the code attempts to emit a WrapScheduleDeleted event with zeroed-out data. 

I agreed that this behaviour is confusing overall.

To prevent this, we should add a condition to check if the wrap schedule exists before proceeding. We can use user field to check is data is there or not.

```
/// @dev IManager.deleteWrapSchedule implementation.
    function deleteWrapScheduleByIndex(bytes32 index) public {
        WrapSchedule storage wrap = wrapSchedule[index];

        address user = wrap.user;

       if(user == address(0)) revert ZeroAddress(); // Fix to avoid empty events

        if (user != msg.sender && wrap.expiry >= block.timestamp)
            revert UnauthorizedCaller(msg.sender, user);

        emit WrapScheduleDeleted(
            index,
            wrap.user,
            address(wrap.superToken),
            address(wrap.strategy),
            wrap.liquidityToken
        );

        delete wrapSchedule[index];
    }
```

shreyaspapi:
- Currently, the deleteWrapScheduleByIndex function emits a WrapScheduleDeleted event with zeroed-out data when called with an index that doesn't correspond to an existing wrap schedule.
- The fix includes the addition of a check for a zero user address.
- By implementing this fix, the function now reverts if the user address is zero, preventing the confusing event data.
- With this fix, the WrapScheduleDeleted event is emitted only when a valid wrap schedule is being deleted, ensuring accurate event logs.

fixes #1441


hellwolf (approved):

hellwolf:
@all-contributors please add @shreyaspapi for code

PR #1447, opened 2023-06-01
[ETHEREUM-CONTRACTS] Deployer v2

### LINKED ISSUE #1430: SuperfluidFrameworkDeployer ergonomic problems
## Purpose
1. Reduce as much friction as possible to get from 0 to Superfluid deployed in a local test environment (including local blockchains such as anvil, hardhat node and ganache).
2. Add some more customization via configuration.
3. Make the deployer more modular by explicitly splitting different parts of the deployment based on category.

See [here](https://hackmd.io/@0xdavinchee/B1v5TZfLn) for a more in-depth spike based on Axe's comments below.

0xdavinchee:
## Purpose
[To improve the ergonomics of our deployer in various testing environments.](https://app.zenhub.com/workspaces/protocol-team-6405e4d5014c0d0036e14f49/issues/gh/superfluid-finance/protocol-monorepo/1430)


saflamini (approved):
Hey @0xdavinchee this looks good. I specifically like the following:

1) Overloading for deployTestFramework is a good choice IMO. however, under what circumstances would the configs actually be used? my guess is that hardly anyone will mess with these. The deployer is used mostly in testing - and the configs aren't really needed there.
2) I like that the deploySuperToken functions have been moved into the deployer. one less thing to import
3) I like the 3 separate functions for Super Token deployment as well. People should understand what they're deploying (I'm not sure what the current deployer supports)

0xdavinchee:
> Hey @0xdavinchee this looks good. I specifically like the following:
> 
> 1. Overloading for deployTestFramework is a good choice IMO. however, under what circumstances would the configs actually be used? my guess is that hardly anyone will mess with these. The deployer is used mostly in testing - and the configs aren't really needed there.

- True, will change this. We may be adding another option here which may confuse. Agree there is hardly anyone who would modify these as most just want to create streams, etc. 
- Configs can always be changed via governance.
- Another reason not to unintuitive overloading in JS: `deployTestFramework["(uint256,...)"]();`

> 2. I like that the deploySuperToken functions have been moved into the deployer. one less thing to import
> 3. I like the 3 separate functions for Super Token deployment as well. People should understand what they're deploying (I'm not sure what the current deployer supports)

Current SuperTokenDeployer has the same functions as well.

packages/ethereum-contracts/README.md
```diff
@@ -174,69 +177,70 @@ Additionally, we use foundry's assert functions to make debugging easier by prov
...
-```sh
-CREATE_NEW_RESOLVER=1 DISABLE_NATIVE_TRUFFLE=true truffle --network ganache exec "node_modules/@superfluid-finance/ethereum-contracts/ops-scripts/deploy-test-environment.js"
-```
+> NOTE: you must deploy erc1820 registry before deploying the Superfluid protocol. You can do this by running `npx hardhat run dev-scripts/deploy-erc1820.js --network localhost` in the terminal.
+
+#### Hardhat
+You can import the `deployContractsAndToken` function from the `dev-scripts/deploy-contracts-and-token.js` file and use it in your testing scripts. It will deploy the Superfluid protocol and deploy Wrapper, Native Asset and Pure SuperToken's for you. 
+Alternatively, you can execute `npx hardhat run dev-scripts/run-deploy-contracts-and-token.js --network localhost` to deploy the contracts and tokens to a local hardhat node, you will know this worked if you see the terminal window with your local hardhat node running spit out a bunch of things.
+
+#### Foundry
```
hellwolf, line 193: Good stuff

packages/ethereum-contracts/contracts/utils/SuperfluidFrameworkDeploymentSteps.sol
```diff
@@ -41,7 +27,16 @@ import { IResolver } from "../interfaces/utils/IResolver.sol";
 /// @author Superfluid
 /// @notice A contract which splits framework deployment into steps.
```
hellwolf, line 28: maybe a few notes on why this was needed in the first place, I don't think it's evident for people that have not experienced this issue.

hellwolf (approved):
Lgtm. Added a few comments where we can benefit from putting a few more lines of comments.

PR #1457, opened 2023-06-07
added simple Dapp for closing optimism streams via L1 portal contract

d10r:
also changed existing gateway links to use cloudflare-ipfs which is more reliable than ipfs.io

related to https://github.com/superfluid-finance/superfluid-stories/issues/317


hellwolf (approved):
we should think about its packaging a little going forward.

PR #1465, opened 2023-06-11
Superapp base testing

ngmachado:
WIP - Following `froundry-test-framework` branch 

- [x] To review Invariance Strategy.
- [x] Expand more tests


ngmachado:
We have decided to replace the Fuzzer with invariance testing. 

This decision is based on the belief that focusing on a single SuperApp invariance - "SuperApp never jailed" - provides a more effective test for our SuperApps' interfaces.

Invariance testing can be challenging for a generic codebase without making specific assumptions. Superfluid's operational nature requires that callbacks to SuperApps come exclusively from `host` contracts, not direct interactions.

To avoid going out of scope on the tests we introduce Handler contracts to force a specific workflow that make sense to our use case.

At present, we have an IStreamHandler that outlines three operations for streams - create, update, and delete. This structure allows the framework to bypass other protocol areas that are not relevant to these tests.

The IStreamHandler interface is defined as follows:

```solidity
interface IStreamHandler {
    function createFlow(address sender, uint32 flowRate) external;
    function updateFlow(address sender, uint32 flowRate) external;
    function deleteFlow(address sender) external;
}
```

__Note that the receiver has been excluded, as the SuperApp being tested should always be the receiver.__

Additionally, within the test implementation, we have included a test counter for each function and integrated it with the invariance test. We have yet to find a way to force foundry to use console.log in these tests. However, for now, we've provided a method to print these counters also as a invariance test.

0xdavinchee:
looks good overall, one minor pedantic thing plz: `forge fmt` the files 😃

hellwolf (approved):

packages/ethereum-contracts/test/foundry/apps/SuperAppTester/handlers/Handler.sol
```diff
@@ -0,0 +1,74 @@
...
+    uint256 public _deleteFlowCallCount;
+    uint256 public _skipDeleteFlowCount;
+
+
+    constructor(address _superAppAddress, ISuperToken _superToken) {
+        superAppAddress = _superAppAddress;
+        superToken = _superToken;
+    }
+
+    function createFlow(address sender, uint32 flowRate) external override {
```
hellwolf: make the minimum bound configurable.

ngmachado: Done. We now have flowRate as int96 to be equal to SF code base.

`Handler` offers a full range default, taking into consideration @Didi values (82bits)

```solidity
// default flow rate bounds, can be overridden on constructor
    int96 public minFlowRate = 1;
    int96 public maxFlowRate = (type(int96).max) >> 14;
```

When using the framework users can configure it using the `Handler` constructor or leave as `0` to use default values.

`handler = new Handler(address(superApp), superToken, MIN_FLOW_RATE, MAX_FLOW_RATE);`

PR #1469, opened 2023-06-13
[ETHEREUM-CONTRACTS] IDAv1Forwarder and Remove CFAHook

0xdavinchee:
Given that we are moving away from the SDKs, we will need the forwarder contracts for nicer UX when not coding in solidity.
This also includes a base that is used by CFAv1Forwarder and will be used by GDAv1Forwarder as well.


hellwolf:
Should consider to simply having one "SuperTokenForwarder" going forward.

kasparkallas:
> Should consider to simply having one "SuperTokenForwarder" going forward.

Having them separate has benefits when it comes to readability.

packages/ethereum-contracts/contracts/utils/IDAv1Forwarder.sol
```diff
@@ -0,0 +1,312 @@
+// SPDX-License-Identifier: AGPLv3
+pragma solidity >= 0.8.4;
```
hellwolf: this should not be ">= 0.8.4".

packages/ethereum-contracts/contracts/utils/IDAv1Forwarder.sol
```diff
@@ -0,0 +1,312 @@
...
+
+/**
+ * @title IDAv1Forwarder
+ * @author Superfluid
+ * The IDAv1Forwarder contract provides an easy to use interface to
+ * InstantDistributionAgreementV1 specific functionality of Super Tokens.
+ * Instances of this contract can operate on the protocol only if configured as "trusted forwarder"
+ * by protocol governance.
+ */
+contract IDAv1Forwarder is ForwarderBase {
```
hellwolf, line 16: perhaps we could try to "implement IInstantAgreementV1".

hellwolf (approved):

hellwolf:
problematic version check pragmas:

```
contracts/interfaces/agreements/IConstantFlowAgreementHook.sol:pragma solidity ^0.8.0;

contracts/mocks/SuperAppBaseFlowTester.sol:pragma solidity >= 0.8.0;
contracts/utils/CFAv1Forwarder.sol:pragma solidity >= 0.8.4;
```

^-- being fixed in https://github.com/superfluid-finance/protocol-monorepo/pull/1478

Issue #15, opened 2020-05-17
organize test cases


hellwolf:
Miao will take this on and organize a bit the test cases.

Issue #14, opened 2020-05-17
experiment with field based agreement storage layout


hellwolf:
To further save some gas, we may replace the current bytes based agreement storage with field based instead.

Issue #13, opened 2020-05-17
embrace eslint max-len


hellwolf:
some of the files have too long lines, we will enable max line 100 or 120.

PR #1470, opened 2023-06-14
[ETHEREUM-CONTRACTS] Adding foundry toga tests

shreyaspapi:
Test -

- [x] Remaining `#11 - #13` 
- [x] Remaining `#14 - #15`

From https://github.com/superfluid-finance/protocol-monorepo/blob/dev/packages/ethereum-contracts/test/contracts/utils/TOGA.test.ts


hellwolf:
@d10r would you be interested in reviewing it too?

packages/ethereum-contracts/test/foundry/utils/TOGA.t.sol
```diff
@@ -0,0 +1,483 @@
...
+        assertTrue(address(0) == toga.getCurrentPIC(superToken));
+
+        vm.startPrank(alice);
+        superToken.send(
+            address(toga),
+            BOND_AMOUNT_1E18,
+            abi.encode(EXIT_RATE_1)
+        );
+
+        assertTrue(toga.getCurrentPIC(superToken) == alice);
```
hellwolf: missing `vm.expectEmit` testing that's present in typescript?

hellwolf: Annoyingly, would need to copy the event signature in this contract

hellwolf:
Perhaps we can add a fuzzing test case now that foundry is being used?

d10r:
hi @shreyaspapi , thx for that!

A few observations:
* using foundry, we could simplify the testing code by taking advantage of its fuzzing tests. Those are especially useful for figuring out edge cases. I added an example in https://github.com/superfluid-finance/protocol-monorepo/commit/7137c9488d6f0a795f8eb20ea9b82963bbefd059 - feel free to copy'n paste that & iterate on it
* `_collectRewards` was a kind of workaround in order to avoid triggering the ERC777 callback. We disabled calling those when using `ERC20.transfer()` later, thus this isn't needed anymore. If it remained like this, you should also migrate over the comments in the JS files, otherwise that context gets lost. But I suggest to instead remove that function altogether and instead use foundry's [deal](https://book.getfoundry.sh/cheatcodes/deal?highlight=deal#deal).
* you can also use `deal` instead of mint
* there's an edge case not covered by the tests (also missing in the JS tests): if the toga bond gets negative (possible if a token goes insolvent), all bids fail. You may try to add a fuzzing test case which uncovers that

packages/ethereum-contracts/test/foundry/utils/TOGA.t.sol
shreyaspapi: Looking into it 👀

shreyaspapi: Added `vm.expectEmit` to test all the events.

shreyaspapi:
@hellwolf @d10r

Thanks for the initial review. I have made some changes as suggested.

- Added fuzzing tests using Foundry to simplify testing and uncover edge cases.
- Removed the `_collectRewards` function and replaced it with Foundry's `deal` functionality.

> - there's an edge case not covered by the tests (also missing in the JS tests): if the toga bond gets negative (possible if a token goes insolvent), all bids fail. You may try to add a fuzzing test case which uncovers that

Regarding the missing edge case mentioned about the toga bond going negative, I believe that if the stream of toga -> PIC gets liquidated, the bond value goes to 0. Please clarify if there's something else I need to consider, can you explain a bit more about `(possible if a token goes insolvent)`.

Please review the changes and let me know if there's anything else to address other then toga going to negative.

packages/ethereum-contracts/test/foundry/utils/TOGA.t.sol
```diff
@@ -0,0 +1,553 @@
...
+/**
+ * @title TOGATest
+ * @dev A contract for testing the functionality of the TOGA contract.
+ */
+contract TOGATest is FoundrySuperfluidTester {
+    using SuperTokenV1Library for ISuperToken;
+
+    TOGA internal toga;
+
+    uint256 internal constant MIN_BOND_DURATION = 3600 * 24 * 7; // 7 days
```
0xdavinchee: you can use `sfDeployer.DEFAULT_TOGA_MIN_BOND_DURATION` to retrieve the canonical min bond duration.
also, you can use `7 days` or `1 week` instead of `3600 * 24 * 7` in solidity

packages/ethereum-contracts/test/foundry/utils/TOGA.t.sol
```diff
@@ -0,0 +1,553 @@
+// SPDX-License-Identifier: AGPLv3
```
0xdavinchee, line 1: please do a forge fmt on this file: 
`forge fmt test/foundry/utils/TOGA.t.sol`

packages/ethereum-contracts/test/foundry/utils/TOGA.t.sol
```diff
@@ -0,0 +1,553 @@
...
+    // Helper
+
+    /**
+     * @dev Checks the net flow of an asset for an account.
+     * @param token The Super Token representing the asset.
+     * @param account The address of the account to check.
+     * @param expectedNetFlow The expected net flow.
+     */
+    function _assetNetflow(
+        ISuperToken token,
```
0xdavinchee: please use `superToken_` instead of `token` as this shadows an existing declaration in storage.

packages/ethereum-contracts/test/foundry/utils/TOGA.t.sol
```diff
@@ -0,0 +1,553 @@
...
+        address account,
+        int96 expectedNetFlow
+    ) internal {
+        int96 flowRate = sf.cfa.getNetFlow(token, account);
+        assertTrue(flowRate == expectedNetFlow);
+    }
+
+    function _deleteFlow(
+        ISuperToken token,
```
0xdavinchee: please refer to `FoundrySuperfluidTester.sol`, there are helper functions there which do a lot more assertions.
see `_helperDeleteFlow`

packages/ethereum-contracts/test/foundry/utils/TOGA.t.sol
```diff
@@ -0,0 +1,553 @@
...
+    function _emptyAccount(
+        address account,
+        ISuperToken token
+    ) internal {
+        vm.startPrank(account);
+        token.transfer(admin, token.balanceOf(account));
+        vm.stopPrank();
+    }
+
+    function _startStream(
```
0xdavinchee: same as the above with `_deleteFlow`

packages/ethereum-contracts/test/foundry/utils/TOGA.t.sol
0xdavinchee: ditto w/ token comment from above.

packages/ethereum-contracts/test/foundry/utils/TOGA.t.sol
```diff
@@ -0,0 +1,553 @@
...
+        address receiver
+    ) internal {
+        vm.startPrank(sender);
+        token.deleteFlow(sender, receiver);
+        vm.stopPrank();
+    }
+
+    function _emptyAccount(
+        address account,
+        ISuperToken token
```
0xdavinchee: ditto w/ token comment from above.

packages/ethereum-contracts/test/foundry/utils/TOGA.t.sol
```diff
@@ -0,0 +1,553 @@
...
+        address receiver,
+        int96 flowRate
+    ) internal {
+        vm.startPrank(sender);
+        superToken.createFlow(receiver, flowRate);
+        vm.stopPrank();
+    }
+
+    function _transferToken(
+        IERC20 token,
```
0xdavinchee: ditto w/ token comment from above. you can use `token_` here

packages/ethereum-contracts/test/foundry/utils/TOGA.t.sol
```diff
@@ -0,0 +1,553 @@
+// SPDX-License-Identifier: AGPLv3
+pragma solidity 0.8.19;
```
0xdavinchee, line 2: the tests are 100% coverage.
please delete the `TOGA.test.ts` file and also remove the line in `testsuites/all-contracts.ts` which references this file to be run for testing.

packages/ethereum-contracts/test/foundry/utils/TOGA.t.sol
```diff
@@ -0,0 +1,553 @@
...
+
+    // Helper
+
+    /**
+     * @dev Checks the net flow of an asset for an account.
+     * @param token The Super Token representing the asset.
+     * @param account The address of the account to check.
+     * @param expectedNetFlow The expected net flow.
+     */
+    function _assetNetflow(
```
0xdavinchee: You might not need this function with the helper functions after

packages/ethereum-contracts/test/foundry/utils/TOGA.t.sol
```diff
@@ -0,0 +1,553 @@
...
+    /**
+     * @dev Sends a PIC bid.
+     * @param sender The address of the sender.
+     * @param token The Super Token representing the asset.
+     * @param bond The bond amount.
+     * @param exitRate The exit rate.
+     */
+    function _sendPICBid(
+        address sender,
+        ISuperToken token,
```
0xdavinchee: ditto w/ token comment from above (use `superToken_`)

packages/ethereum-contracts/test/foundry/utils/TOGA.t.sol
```diff
@@ -0,0 +1,553 @@
...
+        int96 exitRate
+    ) internal {
+        vm.startPrank(sender);
+        token.send(address(toga), bond, abi.encode(exitRate));
+        vm.stopPrank();
+    }
+
+    function _sendPICBid(
+        address sender,
+        ISuperToken token,
```
0xdavinchee: ditto w/ token comment from above (use `superToken_`)

packages/ethereum-contracts/test/foundry/utils/TOGA.t.sol
```diff
@@ -0,0 +1,553 @@
...
+        uint256 bond,
+        bytes memory exitRate
+    ) internal {
+        vm.startPrank(sender);
+        token.send(address(toga), bond, exitRate);
+        vm.stopPrank();
+    }
+
+    function _shouldDefaultExitRate(
```
0xdavinchee: I see that this is a reimplementation `getDefaultExitRateFor`, please use `toga.getDefaultExitRateFor()` instead.

packages/ethereum-contracts/test/foundry/utils/TOGA.t.sol
```diff
@@ -0,0 +1,553 @@
...
+        uint256 bondAmount
+    ) internal pure returns (int96) {
+        return capToInt96(int256(bondAmount / (MIN_BOND_DURATION * 4)));
+    }
+
+    function capToInt96(int256 value) internal pure returns(int96) {
```
0xdavinchee: same as above, best not to duplicate

packages/ethereum-contracts/test/foundry/utils/TOGA.t.sol
```diff
@@ -0,0 +1,553 @@
...
+        return value < type(int96).max ? int96(value) : type(int96).max;
+    }
+
+    function _shouldMaxExitRate(
```
0xdavinchee: same comment as above

packages/ethereum-contracts/test/foundry/utils/TOGA.t.sol
```diff
@@ -0,0 +1,553 @@
...
+            toga.minBondDuration() == MIN_BOND_DURATION,
+            "minBondDuration"
+        );
+    }
+
+    /**
+     * @dev Tests that Alice becomes the PIC.
+     */
+    function testAliceBecomesPIC(uint256 bond, int96 exitRate) public {
+        assertTrue(address(0) == toga.getCurrentPIC(superToken));
```
0xdavinchee: you can alternatively use assertEq(), this makes it more immediately clear that you are asserting that two values are equal than having to do an extra mental step of seeing what is true

packages/ethereum-contracts/test/foundry/utils/TOGA.t.sol
```diff
@@ -0,0 +1,553 @@
...
+        vm.assume(bond < uint256(type(uint64).max)); // Test token only contains 64 bits
+
+        // with small bonds, opening the stream can fail due to CFA deposit having a flow of 1<<32 due to clipping
+         vm.assume(bond > 1<<32 || exitRate == 0);
+
+         vm.assume(exitRate >= 0);
+         // satisfy exitRate constraints of the TOGA
+         vm.assume(exitRate <= toga.getMaxExitRateFor(superToken, bond));
+         // the clipped CFA deposit needs to fit into 64 bits - since that is flowrate multiplied by
+         // liquidation perdion, 14 bits are added for 14400 seconds, so we can't use the full 96 bits
```
0xdavinchee: I would look at some of the assumes and you can move them out into an `_assume` function, see `_assumeValidFlowRate` in `FoundrySuperfluidTester.sol`.

Also a small typo: `perdion` => `period`

packages/ethereum-contracts/test/foundry/utils/TOGA.t.sol
```diff
@@ -0,0 +1,553 @@
...
+
+        _sendPICBid(alice, superToken, bond, exitRate);
+
+        assertTrue(toga.getCurrentPIC(superToken) == alice);
+
+        (address pic, uint256 bond, int96 exitRatePIC) = toga.getCurrentPICInfo(
+            superToken
+        );
+
+        assertTrue(pic == alice);
```
0xdavinchee: use `assertEq` when possible and add an error string at the end of the assert functions for easier debugging. see some of the other test files for the conventions:
`"testAliceBecomesPIC: Alice is PIC"`
The convention is: `"{NAME_OF_FUNCTION}: {BRIEF_DESCRIPTION_OF_ASSERTION}"`

packages/ethereum-contracts/test/foundry/utils/TOGA.t.sol
0xdavinchee: if it is needed, move this to `FoundrySuperfluidTester.sol`, also a minor typo: `_assertNetFlow` => `assertNetFlow`

packages/ethereum-contracts/test/foundry/utils/TOGA.t.sol
```diff
@@ -0,0 +1,553 @@
...
+     */
+    function testBobOutBidsAlice(uint256 bobBond, uint256 aliceBond) public {
+        vm.assume(aliceBond > 0);
+        vm.assume(bobBond > aliceBond);
+        vm.assume(bobBond < uint256(type(uint64).max)); // User only has 64 bits test super tokens
+
+        // Send PIC bid from Alice
+        _sendPICBid(alice, superToken, aliceBond, 0);
+
+        vm.startPrank(alice);
```
0xdavinchee: this prank may not be needed?

packages/ethereum-contracts/test/foundry/utils/TOGA.t.sol
```diff
@@ -0,0 +1,553 @@
...
+        // Assert Alice is the current PIC
+        assertTrue(toga.getCurrentPIC(superToken) == alice);
+
+        // Get the bond amount
+        (, uint256 bond, ) = toga.getCurrentPICInfo(superToken);
+        assertTrue(bond == aliceBond);
+
+        vm.stopPrank();
+
+        vm.expectRevert("TOGA: bid too low");
```
0xdavinchee: please move this expectRevert function into its own test:
`function testRevertIfBidTooLow(...)`
You can do a global search for: `testRevertIf` to see examples of this

packages/ethereum-contracts/test/foundry/utils/TOGA.t.sol
```diff
@@ -0,0 +1,553 @@
...
+        address implementer1 = _ERC1820_REG.getInterfaceImplementer(
+            address(toga),
+            keccak256("TOGAv1")
+        );
+        address implementer2 = _ERC1820_REG.getInterfaceImplementer(
+            address(toga),
+            keccak256("TOGAv2")
+        );
+
+        assertTrue(
```
0xdavinchee: can use assertEq here too

packages/ethereum-contracts/test/foundry/utils/TOGA.t.sol
```diff
@@ -0,0 +1,553 @@
...
+            implementer1 == address(toga),
+            "TOGA should be registered as TOGAv1"
+        );
+        assertTrue(
+            implementer2 == address(toga),
+            "TOGA should be registered as TOGAv2"
+        );
+    }
+
+    function testEnforceMinExitRateLimit() public {
```
0xdavinchee: just a convention thing, but use `testRevertIf__` naming convention and try to keep the tests as small as possible, that is if this test is expecting a revert, it should only be doing that.

So this test could be split in two, if the second part where you send the pic bid to get a 0 net flow rate is not tested elsewhere already.

packages/ethereum-contracts/test/foundry/utils/TOGA.t.sol
```diff
@@ -0,0 +1,553 @@
...
+        assertTrue(
+            _shouldMaxExitRate(bond) ==
+                toga.getMaxExitRateFor(superToken, bond)
+        );
+        assertTrue(
+            _shouldDefaultExitRate(bond) ==
+                toga.getDefaultExitRateFor(superToken, bond)
+        );
+        // the default exit rate needs to be equal or greater than the max exit rate
+        assertTrue(
```
0xdavinchee: you can also use `assertGe()` :)

packages/ethereum-contracts/test/foundry/utils/TOGA.t.sol
```diff
@@ -0,0 +1,553 @@
...
+
+        _sendPICBid(alice, superToken, bond, abi.encode());
+        _assetNetflow(
+            superToken,
+            alice,
+            toga.getDefaultExitRateFor(superToken, bond)
+        );
+    }
+
+    function testCannotBecomePICWithBidSmallerThanCurrentPICBond(uint256 bond, uint256 smallerBond) public {
```
0xdavinchee: please use the `testRevertIf` naming

packages/ethereum-contracts/test/foundry/utils/TOGA.t.sol
```diff
@@ -0,0 +1,553 @@
...
+        _sendPICBid(alice, superToken, bond, exitRate);
+
+        assertTrue(toga.getCurrentPIC(superToken) == alice);
+
+        (address pic, uint256 bond, int96 exitRatePIC) = toga.getCurrentPICInfo(
+            superToken
+        );
+
+        assertTrue(pic == alice);
+        assertTrue(bond == bond);
```
0xdavinchee: I would change the name of the param here to `bond_`

packages/ethereum-contracts/test/foundry/utils/TOGA.t.sol
```diff
@@ -0,0 +1,553 @@
+// SPDX-License-Identifier: AGPLv3
+pragma solidity 0.8.19;
+
```
0xdavinchee, line 3: I'd highly recommend `Solidity Visual Developer` if you're using VSCode as well :)

packages/ethereum-contracts/test/foundry/utils/TOGA.t.sol
```diff
@@ -0,0 +1,553 @@
...
+        vm.expectRevert("TOGA: exitRate too high");
+        toga.changeExitRate(superToken, _shouldDefaultExitRate(aliceBond));
+        vm.stopPrank();
+
+        deal(address(superToken), address(toga), 1e15);
+
+        (, uint256 bond2, ) = toga.getCurrentPICInfo(superToken);
+        assertTrue(bond2 >= 1e12);
+
+        vm.startPrank(alice);
```
0xdavinchee: the changing of the exit rate likely can be placed in a helper function where you assertNetFlow after.

0xdavinchee (commented):
gm! 

This looks very good overall, the logic is being tested well and you have 100% coverage for the TOGA, so props! We are very grateful that you've written these tests.

Almost all of my comments are regarding convention and the way of writing tests at Superfluid, you can review the [README](https://github.com/superfluid-finance/protocol-monorepo/blob/dev/packages/ethereum-contracts/README.md) in ethereum-contracts to get a better idea of the how and why.

Some high level comments, I've referenced some examples in the code with comments as well:
- aim to not duplicate logic as this leads to maintenance overhead and possible bugs down the line
- utilize existing helper functions in `FoundrySuperfluidTester.sol` when possible
- be careful with shadowed declarations, the plugin I recommended will make this visible immediately
- use `assertEq`, `assertGe`, etc. when possible and use `assertTrue` when the other options aren't available.
- there are `assume's` for the different fuzzing tests which are repeated, try to see which ones can be placed in an `_assumeXYZ` function, this follows from the duplication point above
- keep the tests specific to testing one thing, if it's supposed to test a passing case, only test that, if it's supposed to revert, only test that, you may need several steps to test certain cases (e.g. testMultiplePICsInParallel), but the reverts in there can be taken out into their own revert tests
  - use `testRevertIf` naming convention

hellwolf:
should also remove the original test file then we are ready to merge.

hellwolf (approved):
remove js test then we should be good!

shreyaspapi:
@0xdavinchee I see you have made a commit, I also had some uncommitted changes 😕, that addresses the issues.

0xdavinchee:
@shreyaspapi oh okay, sorry about that! I wanted to help you with some of the changes because I requested a lot of different things.

PR #1473, opened 2023-06-19
Add support for celo-alfajores network

nestorbonilla:


vmichalik:
Hi @nestorbonilla.
Unfortunately, adding new networks is not as simple as making a PR to our monorepo, as there are many peripheral pieces required to ensure good developer experience.
We do not currently believe that Celo's testnet is sufficiently differentiated or in-demand by our community for us to prioritize supporting the deployment there. Instead we are going to supply a developer faucet and remove other restrictions on Celo mainnet, since gas costs there are very low.
Let us know if you have further comments or questions.

Issue #1472, opened 2023-06-19
[ETHEREUM-CONTRACTS] add support for Safe as additional multisig type for gov actions


d10r:
Status quo: gov can be owned either by an EOA or by a [classic multisig wallet](https://github.com/gnosis/MultiSigWallet).

In order to move towards a governance process involving more parties and allowing easier integration with the evolving ecosystem around managing governance, we add support for Safe as governance owner.
Scope of this task:

* allow to run gov scripts with `GOVERNANCE_ADMIN_TYPE=SAFE`
* set up a testnet deployment with a Safe as gov owner

d10r:
testing with:
```
$ GOVERNANCE_ADMIN_TYPE=SAFE RELEASE_VERSION=safe npx truffle exec --network base-goerli ops-scripts/gov-set-trusted-forwarder.js : 0x0000000000000000000000000000000000000000 "0x$(ethkey generate random | tail -n 1 | cut -d ' ' -f2)" 1
```
what can be improved:
* execute if the proposer alone can (1 of 1)
* add auto-detect support

PR #1477, opened 2023-06-26
[ETHEREUM-CONTRATS] added support for SAFE in ops-scripts

d10r:
see #1472


hellwolf:
We talked about using the next testnet or mainnet as testbed for this new protocol-ops setup.

hellwolf (approved):

PR #1478, opened 2023-06-26
License Cleanups

hellwolf:
Overall Superfluid protocol uses AGPLv3 license. But it is also intended that the contracts used for development on Superfluid Protocol should use more relaxed open source license, namely MIT.

This is the pull request that fixes this:

- All licenses file are updated re their copy right status.
- Top-level license file is adjusted to reflect the licenses of each package correctly.
- ethereum-contracts now is dual licensed "(AGPLv3 OR MIT)":
   - LICENSE file to include both licenses,
   - individual source files to state their license type correctly,
   - added a license section to README to further clarify this.


0xdavinchee (approved):
LGTM

PR #1479, opened 2023-06-26
Added zkEVM testnet

d10r:
This PR
* adds support for zkevm-testnet
* adds a subgraph config for base-goerli (piggy-backing)
* uses metadata in getConfig
* reduces config duplication in various files


packages/js-sdk/src/getConfig.js
```diff
@@ -9,17 +9,16 @@ const getConfigData = (chainId) => {
     if (!networkData) {
         console.log(`no metadata found for network with chainId ${chainId}`);
         return {
-            nativeTokenSymbol: "",
-            resolverAddress: "",
+            isTestnet: true,
```
hellwolf, line 12: Observation: this introduces changes to js-sdk which will be unpublished.

It's time to piggyback this change:

- Remove ""ops-scripts/**/*"," from package.json
- dev-scripts remain published, since it should be for developers.
- to use ops-scripts, one should always check out the git repo branch.

hellwolf (approved):

packages/js-sdk/src/getConfig.js
hellwolf: I changed it.

PR #1483, opened 2023-06-29
Remove js-sdk

d10r:
tentative PR probing how difficult/messy (or not) it would be to remove js-sdk


d10r:
outcome: won't remove, but may merge into ethereum-contracts package

PR #1505, opened 2023-07-12
Added missing metadata types into list.d.ts

msoni89:


packages/metadata/module/networks/list.d.ts
```diff
@@ -9,6 +9,9 @@ interface ContractAddresses {
     readonly superTokenFactory: string;
     readonly superfluidLoader: string;
     readonly toga: string;
```
msoni89, line 11: Not sure, how its working but toga is an option field but here it's string types.

0xdavinchee (approved):
thank you - LGTM

PR #1506, opened 2023-07-12
[WORKFLOWS] Fix Subgraph Versions Check

0xdavinchee:
Subgraph does not have a released NPM package, therefore we need to use different logic otherwise it breaks when doing `npm show`.

### DIFF
--- a/.github/workflows/cd.packages-stable.create-release-drafts.yml
+++ b/.github/workflows/cd.packages-stable.create-release-drafts.yml
@@ -65,7 +65,7 @@ jobs:
       - name: Check package versions
         run: |
           SUBGRAPH_NEW_VERSION=$(jq -r .version packages/subgraph/package.json)
-          if git tag | grep -e "^subgraph" | grep v"$SUBGRAPH_NEW_VERSION";then
+          if git tag | grep -e "^subgraph@v${SUBGRAPH_NEW_VERSION}$";then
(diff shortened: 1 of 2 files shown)


.github/workflows/cd.packages-stable.create-release-drafts.yml
```diff
@@ -63,7 +63,16 @@ jobs:
           registry-url: https://registry.npmjs.org/
 
       - name: Check package versions
-        run: ./tasks/check-package-version.sh subgraph SHOULD_PUBLISH_SUBGRAPH SUBGRAPH_NEW_VERSION >> "$GITHUB_ENV"
+        run: |
+          SUBGRAPH_NEW_VERSION=$(jq -r .version packages/subgraph/package.json)
+          if git tag | grep -e "^subgraph" | grep v"$SUBGRAPH_NEW_VERSION";then
```
hellwolf: could also just be more strict: `git tag | grep -e "^subgraph@v${SUBGRAPH_NEW_VERSION}$"`

hellwolf (approved):
minor change,

.github/workflows/cd.packages-stable.create-release-drafts.yml
0xdavinchee: done

PR #1519, opened 2023-07-14
Add contract verification check to daily slack message

### LINKED ISSUE #1502: [BUG] Contracts unverified on some networks
@ngmachado noticed that AVAXx is not verified on Fuji, I verified the underlying SuperToken logic but it appears that we haven't verified the SETHProxy on all networks. 

We took this as an opportunity to scope out what other contracts may be unverified on all networks.

Next steps: 
- [ ] [verify unverified contracts](https://github.com/superfluid-finance/protocol-monorepo/issues/1503)

elvijsTDL:
Screenshot 2023-07-14 at 13 07 45
Screenshot 2023-07-14 at 13 08 43

### DIFF
--- a/.github/workflows/daily-slack-message.yml
+++ b/.github/workflows/daily-slack-message.yml
@@ -1,3 +1,5 @@
+name: Daily protocol-monorepo slack message
+
 on:
   workflow_dispatch:
   schedule:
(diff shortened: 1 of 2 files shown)


hellwolf (approved):
NICE !

Issue #1549, opened 2023-07-25
[WORKFLOW] Upgrade package versions for hardhat


mmd-afegbua:
This PR caused this error in CI: https://github.com/superfluid-finance/protocol-monorepo/pull/1390/files

`WARNING: You are using a version of Node.js that is not supported, and it may work incorrectly, or not work at all. See https://hardhat.org/nodejs-versions`

https://github.com/superfluid-finance/protocol-monorepo/actions/runs/5620894738/job/15230659806?pr=1390

PR #1550, opened 2023-07-25
#1549 Upgrade hardhat packages

mmd-afegbua:


hellwolf:
This will be continued in the `update_nix_flake_inputs` branch instead.

PR #1551, opened 2023-07-26
test codeccov forked PR

mmd-afegbua:


hellwolf:
looks good now?

mmd-afegbua:
Yes. Codecov is working as expected. I'll wait for someone that doesn't have write access to create a forked PR so all doubts will be cleared.

PR #1553, opened 2023-07-28
Issue 1487 - add total number of holders counter and total number of accounts into token statistic

### LINKED ISSUE #1487: [SUBGRAPH] Add `totalNumberOfHolders` counter to `TokenStatistic`

msoni89:
Added two fields now: unique total accounts - interacted (will not decrease if the user's balance reaches 0), and total holders - current balance holder (will decrease if the user's balance reaches 0, and increase if the user's balance is greater than 0).


packages/subgraph/schema.graphql
```diff
@@ -2111,6 +2111,16 @@ type TokenStatistic @entity {
     """
     totalSupply: BigInt!
 
+    """
+    The total number of 'token' accounts.
+    """
+    totalNumberOfAccounts: Int!
+
+    """
+    The total number of 'token' accounts holding a balance.
```
kasparkallas: Specify "non-zero".

As in: "The total number of accounts holding a non-zero balance of the token."

packages/subgraph/schema.graphql
```diff
@@ -2111,6 +2111,16 @@ type TokenStatistic @entity {
     """
     totalSupply: BigInt!
 
+    """
+    The total number of 'token' accounts.
+    """
+    totalNumberOfAccounts: Int!
```
kasparkallas, line 2117: Maybe specify that this means who've interacted with the token.

As in: "The total number of accounts that have interacted with the token (but might not hold a balance anymore)."

packages/subgraph/src/mappingHelpers.ts
```diff
@@ -808,6 +818,9 @@ export function updateATSStreamedAndBalanceUntilUpdatedAt(
         tokenAddress,
         block
     );
+
+    let balanceUntilUpdatedAtBeforeUpdate = accountTokenSnapshot.balanceUntilUpdatedAt;
```
kasparkallas: Can use `const`? Otherwise you signal that this value changes later in the code flow.

packages/subgraph/src/mappingHelpers.ts
```diff
@@ -864,6 +877,31 @@ export function updateATSStreamedAndBalanceUntilUpdatedAt(
     );
     accountTokenSnapshot.save();
 
+    let balanceUntilUpdatedAtAfterUpdate = accountTokenSnapshot.balanceUntilUpdatedAt;
```
kasparkallas: `const`

packages/subgraph/tests/superToken/event/superToken.event.test.ts
```diff
@@ -267,6 +267,28 @@ describe("SuperToken Mapper Unit Tests", () => {
...
+                0, // totalSubscriptionsWithUnits
+                0, // totalApprovedSubscriptions
+                BIG_INT_ZERO, // totalDeposit
+                BIG_INT_ZERO, // totalOutflowRate
+                BIG_INT_ZERO, // totalAmountStreamedUntilUpdatedAt
+                BIG_INT_ZERO, // totalAmountTransferredUntilUpdatedAt
+                BIG_INT_ZERO, // totalAmountDistributedUntilUpdatedAt
+                BigInt.fromI32(1000000), // totalSupply = 100
+                3, // totalNumberOfAccounts
+                3 // totalNumberOfHolders
```
kasparkallas, line 290: Use different values if possible, otherwise they might be mixed up and the test wouldn't catch this high likelihood mistake.

msoni89: For the above test case, it is expected to return "3" as totalNumberOfHolders and can't pass anything else or a different value.


Done - added new test case to show "TokenStatistic::totalNumberOfHolders should decrease its count when a user transfers tokens and the balance reaches 0."

0xdavinchee:
keep in mind while this likely will catch a majority of cases, it is possible we don't count a few recipients of IDA distributions as this isn't picked up by the subgraph and therefore the logic here.

kasparkallas (approved):

hellwolf:
ready to merge?

PR #1556, opened 2023-08-03
Make public and external methods virtual to facilitate creation of custom SuperToken logic

d10r:
Customized versions of the SuperToken logic currently cannot just inherit `SuperToken.sol`, because it doesn't declare its methods as `virtual`.

Examples:
* In https://github.com/superfluid-finance/custom-supertokens/pull/26, `updateCode` changes.
* In https://github.com/GoodDollar/GoodProtocol/blob/master/contracts/token/superfluid/SuperToken.sol, the implementation of several methods (e.g. `transfer`) is changed.

In this PR, all public and external methods of SuperToken (and SuperfluidToken which it inherits from) are declared as `virtual` in order to facilitate creation of such customizations.

Additionally a few internal methods (for mint, burn, transfer, upgrade and downgrade) were also made virtual in order to allow changing the behaviour of that operations with less additional code.

### DIFF
--- a/packages/ethereum-contracts/contracts/superfluid/SuperToken.sol
+++ b/packages/ethereum-contracts/contracts/superfluid/SuperToken.sol
@@ -637,7 +637,7 @@ contract SuperToken is
         uint256 amount,
         bytes memory userData,
         bytes memory operatorData
-    ) private {
+    ) internal virtual {
         if (address(_underlyingToken) == address(0)) revert SUPER_TOKEN_NO_UNDERLYING_TOKEN();
 
         (uint256 underlyingAmount, uint256 adjustedAmount) = _toUnderlyingAmount(amount);
@@ -661,7 +661,8 @@ contract SuperToken is
         address to,       // the account receiving the underlying tokens
         uint256 amount,
         bytes memory data,
-        bytes memory operatorData) private {
+        bytes memory operatorData
+    ) internal virtual {
         if (address(_underlyingToken) == address(0)) revert SUPER_TOKEN_NO_UNDERLYING_TOKEN();
 
         (uint256 underlyingAmount, uint256 adjustedAmount) = _toUnderlyingAmount(amount);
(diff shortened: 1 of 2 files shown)


packages/ethereum-contracts/contracts/superfluid/SuperToken.sol
```diff
@@ -192,7 +193,7 @@ contract SuperToken is
      * For more context, see https://github.com/superfluid-finance/protocol-monorepo/wiki/About-ERC-777
      */
     function _transferFrom(address spender, address holder, address recipient, uint amount)
-        internal returns (bool)
+        internal virtual returns (bool)
```
hellwolf: We are doing a much bigger surgery than I heard of.

This is code smell and anti-pattern, in general one should not change the behaviour of these "_*" functions for it is crafted logic for the reusing site.

Instead, a more idiomatic OO way (not really promoting it anymore, but within the conext of this codebase) is to override the site that uses those code, rewiring the logic etc.

hellwolf: let's discuss it separately tomorrow.

hellwolf (changes requested):
Code organization principles:

- `_foo` internal function as building blocks for reusability.
  - they should not be altered by subclasses.
  - "private" scope is generally avoided for simplicity, since test-harness may benefit from accessing them directly by subclassing it. This is often the case for our test cases.
  - The building blocks become "monadic", if it calls other abstract virtual functions. The naming convention for these functions are "_doSomeDamage".
- `external` functions can be created through assembling these internal logic together.
  - `public` functions is a shortcut used in code base to combine the definition of  `_foo` and `foo`. This style should be gradually rid of from this code base, in favor of separation of definitions from day 0.


There are generally two ways of code logic reassembling:

```
function foo() external virtual override {
   // inject some logic before...
   super.foo();
   // inject some logic after...
}
```

or a complete rewrite of `foo` using `_foo` etc.

packages/ethereum-contracts/contracts/superfluid/SuperToken.sol
d10r: changed to only have external and public methods virtual

hellwolf (approved):

hellwolf (approved):
I still find it a bit stinky to be able to override so many things, but it is at least better than being able to override internal logic blocks.

PR #1559, opened 2023-08-04
[ETHEREUM-CONTRACTS|SDK-CORE] 1544 new flow operator updates

### LINKED ISSUE #1544: [ETHEREUM-CONTRACTS] Add addFlowOperatorPermissions which takes increaseflowrateAllowance as an input to ACL
## Problem

Many of our core user groups use gnosis safes, which means tx preparation and signing are done completely separately - many transactions may be prepared before a single one is signed
this breaks a key assumption of `updateFlowOperatorPermissions`, which requires `flowrateAllowance` as an input - you have to know the total `flowrateAllowance` to set
we fixed the issue with `flowrateAllowance` by creating `increaseFlowRateAllowance` where we can just add extra permissions without knowing the exact total after all transactions
BUT we still need to fix `flowOperatorPermissions`

## Solution

We add `addFlowOperatorPermissions` which takes `increaseflowrateAllowance` as an input

0xdavinchee:
## Ethereum Contracts
- `increaseFlowAllowanceWithPermissions` and `decreaseFlowAllowanceWithPermissions` added to `ConstantFlowAgreementV1.sol`
- Slight refactoring of validation logic, **modifying existing deployed code path** (see `_validateUpdateFlowOperatorDataInput` function)

## SDK-Core
- `increaseFlowAllowanceWithPermissions` and `decreaseFlowAllowanceWithPermissions` added to `ConstantFlowAgreementV1` class


kasparkallas (changes requested):
The increasing and decreasing of flow operator permissions should also function delta-based, not snapshot-based.

For example, if I increase a permission by 1 (Create) then it should never remove other permissions (Delete or Update).

Some other examples:
1. If my permissions are 1 (Create) and I increase by 1 then it stays at 1.
2. If my permissions are 1 (Create) and I increase by 2 (Update) then it becomes 3 (Create or Update).
3. If my permissions are 5 (Create or Delete) and I increase by 1 (Create) then it stays 5.
4. If my permissions are 1 (Create) and I decrease by 5 (Create or Delete) then it should become 0.
5. If My permissions are 1 (Create) and I decrease by 6 (Delete or Update) then it should stay 1.

Link to permission values for quick reference: https://docs.superfluid.finance/superfluid/developers/constant-flow-agreement-cfa/cfa-access-control-list-acl#permissions-parameter

packages/ethereum-contracts/contracts/agreements/ConstantFlowAgreementV1.sol
```diff
@@ -934,6 +910,120 @@ contract ConstantFlowAgreementV1 is
...
+    function _validateUpdateFlowOperatorDataInput(
+        ISuperfluidToken token,
+        address flowOperator,
+        uint8 permissions,
+        int96 flowAllowance,
+        bytes calldata ctx
+    ) internal returns (ISuperfluid.Context memory currentContext) {
+        if (!FlowOperatorDefinitions.isPermissionsClean(permissions)) revert CFA_ACL_UNCLEAN_PERMISSIONS();
+        // [SECURITY] NOTE: we are holding the assumption here that ctx is correct and we validate it with
+        // authorizeTokenAccess:
```
kasparkallas: Nitpick: What is the point of this comment?

packages/ethereum-contracts/contracts/agreements/ConstantFlowAgreementV1.sol
```diff
@@ -934,6 +910,120 @@ contract ConstantFlowAgreementV1 is
...
+    {
+        return existingPermissions & (~permissionDelta);
+    }
+    
+    /// @dev This function ensures:
+    /// - token access
+    /// - passed permissions are "clean"
+    /// - no sender flow operator
+    /// - no negative allowance
+    function _validateUpdateFlowOperatorDataInput(
```
kasparkallas: Nitpick: It's also authorizing though...

packages/ethereum-contracts/contracts/agreements/ConstantFlowAgreementV1.sol
```diff
@@ -934,6 +910,120 @@ contract ConstantFlowAgreementV1 is
...
+
+    function deltaAddPermissions(uint8 existingPermissions, uint8 permissionDelta)
+        public
+        pure
+        returns (uint8)
+    {
+        return existingPermissions | permissionDelta;
+    }
+
+    function deltaRemovePermissions(uint8 existingPermissions, uint8 permissionDelta)
```
kasparkallas: Nitpick: IMO the "delta" prefix in the functions names is not necessary.

packages/ethereum-contracts/test/foundry/agreements/ConstantFlowAgreementV1.ACL.t.sol
```diff
@@ -155,17 +165,138 @@ contract ConstantFlowAgreementV1ACLTest is FoundrySuperfluidTester {
...
+        vm.assume(FlowOperatorDefinitions.isPermissionsClean(permissions));
+        vm.assume(flowAllowance > 0);
+        vm.assume(address(this) != flowOperator);
+
+        (bytes32 flowOperatorId, uint8 oldPermissions, int96 oldFlowRateAllowance) =
+            sf.cfa.getFlowOperatorData(superToken, address(this), flowOperator);
+
+        superToken.increaseFlowAllowanceWithPermissions(flowOperator, permissions, flowAllowance);
+
+        uint8 newPermissions = sf.cfa.deltaAddPermissions(oldPermissions, permissions);
```
kasparkallas: I'm not sure I like this...

Feels like testing the implementation with _the implementation_.

Is `deltaAddPermissions` covered elsewhere?

packages/ethereum-contracts/contracts/interfaces/agreements/IConstantFlowAgreementV1.sol
```diff
@@ -143,6 +143,42 @@ abstract contract IConstantFlowAgreementV1 is ISuperAgreement {
...
+     * @dev msgSender from `ctx` increases flow rate allowance for the `flowOperator` by `addedFlowRateAllowance`
+     * @dev if `addedFlowRateAllowance` is negative, we revert with CFA_ACL_NO_NEGATIVE_ALLOWANCE
+     * @param token Super token address
+     * @param flowOperator The permission grantee address
+     * @param permissionsToAdd A bitmask representation of the granted permissions to add as a delta
+     * @param addedFlowRateAllowance The flow rate allowance delta
+     * @param ctx Context bytes (see ISuperfluid.sol for Context struct)
+     * @return newCtx The new context bytes
+     */
+    function increaseFlowAllowanceWithPermissions(
```
kasparkallas: It's a tiny detail but there is a bit of a new term "flow allowance" used here as opposed to "flow rate allowance".

Fun fact, in the products, we've been calling it "stream allowance".

kasparkallas: Yea, some other functions are named "increaseFlow**Rate**Allowance" and "decreaseFlow**Rate**Allowance".

packages/ethereum-contracts/contracts/agreements/ConstantFlowAgreementV1.sol
```diff
@@ -934,6 +910,120 @@ contract ConstantFlowAgreementV1 is
...
+
+        FlowOperatorData memory flowOperatorData;
+        flowOperatorData.permissions = deltaRemovePermissions(oldPermissions, permissionsToRemove);
+        int96 newFlowRateAllowance = oldFlowRateAllowance - subtractedFlowRateAllowance;
+
+        // @note this defends against negative allowance
+        if (newFlowRateAllowance < 0) revert CFA_ACL_NO_NEGATIVE_ALLOWANCE();
+        flowOperatorData.flowRateAllowance = newFlowRateAllowance;
+
+        token.updateAgreementData(flowOperatorId, _encodeFlowOperatorData(flowOperatorData));
```
kasparkallas: Wouldn't it be better to use `_updateFlowRateAllowance` here or what's the point of that function?

packages/ethereum-contracts/contracts/agreements/ConstantFlowAgreementV1.sol
```diff
@@ -826,30 +826,6 @@ contract ConstantFlowAgreementV1 is
...
-        flowOperatorData.permissions = permissions;
-        flowOperatorData.flowRateAllowance = flowRateAllowance;
-        bytes32 flowOperatorId = _generateFlowOperatorId(currentContext.msgSender, flowOperator);
-        token.updateAgreementData(flowOperatorId, _encodeFlowOperatorData(flowOperatorData));
-
-        emit FlowOperatorUpdated(token, currentContext.msgSender, flowOperator, permissions, flowRateAllowance);
-    }
-
     /// @dev IConstantFlowAgreementV1.increaseFlowRateAllowance implementation
     function increaseFlowRateAllowance(
```
kasparkallas, line 830: Shouldn't this function delegate to `increaseFlowAllowanceWithPermissions` with `permissionsToAdd` set as 0?

Same goes for `decreaseFlowRateAllowance` .

Seems to me like all the code here could be de-duplicated.

packages/ethereum-contracts/contracts/agreements/ConstantFlowAgreementV1.sol
```diff
@@ -934,6 +910,120 @@ contract ConstantFlowAgreementV1 is
...
+
+        token.updateAgreementData(flowOperatorId, _encodeFlowOperatorData(flowOperatorData));
+
+        emit FlowOperatorUpdated(
+            token, currentContext.msgSender, flowOperator, flowOperatorData.permissions, newFlowRateAllowance
+        );
+    }
+
+    function deltaAddPermissions(uint8 existingPermissions, uint8 permissionDelta)
+        public
```
kasparkallas, line 928: Is there any value in _not_ making this public?

packages/ethereum-contracts/contracts/agreements/ConstantFlowAgreementV1.sol
0xdavinchee: we add this everywhere we do authorizeTokenAccess as a reminder-primarily for ourselves or auditors.

0xdavinchee: primarily as a reminder to ourselves and auditors, but it is not copied everywhere so will remove and add in `AgreementLibrary.sol`

packages/ethereum-contracts/contracts/agreements/ConstantFlowAgreementV1.sol
0xdavinchee: using the more verbose: `_validateAndAuthorizeUpdateFlowOperatorDataInput`

packages/ethereum-contracts/contracts/agreements/ConstantFlowAgreementV1.sol
0xdavinchee: fairly indifferent on this one, so will remove `delta`

packages/ethereum-contracts/test/foundry/agreements/ConstantFlowAgreementV1.ACL.t.sol
0xdavinchee: we are not explicitly testing `deltaAddPermissions`.

packages/ethereum-contracts/contracts/interfaces/agreements/IConstantFlowAgreementV1.sol
0xdavinchee: This was done consciously, mainly due to verbosity of the function name, but I see your point here, it is inconsistent and may lead one to question if there is a difference in behavior between the two functions.

packages/ethereum-contracts/contracts/agreements/ConstantFlowAgreementV1.sol
0xdavinchee: it was used previously in the cases where we modified the flow rate allowance, but kept the permissions the same (create/updateByFlowOperator and increase/decreaseAllowance) - but I will be changing the name of the function (it's fine because the function internal) to `_updateFlowOperatorData` as this is what it really does.

packages/ethereum-contracts/contracts/agreements/ConstantFlowAgreementV1.sol
0xdavinchee: This is public mainly for ease of testability, but given the earlier comment regarding testing the implementation with implementation, the value may lie there in not making it public.

However, testing bitwise operations with fuzzing seems a bit pedantic here.

packages/ethereum-contracts/contracts/agreements/ConstantFlowAgreementV1.sol
0xdavinchee: good catch, done

kasparkallas (approved):

packages/ethereum-contracts/testing-benchmark.json
```diff
@@ -0,0 +1,2986 @@
+[
```
hellwolf: what is this file

packages/ethereum-contracts/package.json
```diff
@@ -29,8 +29,8 @@
     "main": "./dev-scripts/index.js",
     "typings": "./types/index.d.ts",
     "scripts": {
-        "dev": "tasks/dev.sh",
-        "dev-forge": "nodemon -e sol -x yarn run-forge test --hardhat",
+        "dev": "nodemon -e sol -x yarn run-forge test --hardhat",
```
hellwolf: `yarn run-forge` still and `--hardhat`?

hellwolf: tweaked

hellwolf (approved):

PR #1565, opened 2023-08-08
[ETHEREUM-CONTRACTS] allow to update to custom SuperToken logic

d10r:
Allows SF gov to update SuperTokens to a custom logic.

### DIFF
--- a/packages/ethereum-contracts/contracts/gov/SuperfluidGovernanceBase.sol
+++ b/packages/ethereum-contracts/contracts/gov/SuperfluidGovernanceBase.sol
@@ -97,11 +97,11 @@ abstract contract SuperfluidGovernanceBase is ISuperfluidGovernance
         onlyAuthorized(host)
     {
         for (uint i = 0; i < tokens.length; ++i) {
-            host.updateSuperTokenLogic(tokens[i]);
+            host.updateSuperTokenLogic(tokens[i], address(0));
         }
     }
 
-    function batchUpdateSuperTokenLogicCustom(
+    function batchUpdateSuperTokenLogic(
         ISuperfluid host,
         ISuperToken[] calldata tokens,
         address[] calldata tokenLogics
@@ -111,7 +111,7 @@ abstract contract SuperfluidGovernanceBase is ISuperfluidGovernance
     {
         if (tokens.length != tokenLogics.length) revert SF_GOV_ARRAYS_NOT_SAME_LENGTH();
         for (uint i = 0; i < tokens.length; ++i) {
-            host.updateSuperTokenLogicCustom(tokens[i], tokenLogics[i]);
+            host.updateSuperTokenLogic(tokens[i], tokenLogics[i]);
         }
     }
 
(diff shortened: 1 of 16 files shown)


packages/ethereum-contracts/contracts/gov/SuperfluidGovernanceBase.sol
```diff
@@ -101,6 +101,20 @@ abstract contract SuperfluidGovernanceBase is ISuperfluidGovernance
         }
     }
 
+    function batchUpdateSuperTokenLogicCustom(
```
hellwolf: We could also overload and generalize it as a replacement to the previous version:

```
address[] calldata customTokenLogics
```

if `customTokenLogics[i]` is zero, we apply the regular `updateSuperTokenLogic` otherwise.

hellwolf: oh, so we need to update governance to all the network, right? Scary :)

hellwolf (changes requested):

packages/ethereum-contracts/contracts/gov/SuperfluidGovernanceBase.sol
d10r: yeah. If we can't do it without governance update anyway, I'd go with the non-breaking change of leaving the previous methods and adding one.
In case of the host contract however it may be a good idea to replace with such a generalized version. It's a breaking change too, but since it can only be invoked by gov and they're updated (or not) together, it wouldn't break anything.
Also in light of that contract being close to the contract size limit.

d10r: I pushed a commit which
* changes Superfluid.updateSuperTokenLogic to a generalized function (as you proposed)
* adds an overloaded version of SuperfluidGovernanceBase.batchUpdateSuperTokenLogic with an additional argument

Since this method is consumed by the ops-scripts, the UX issue of frameworks with handling overloaded functions is imo not an issue.

@hellwolf What do you think about this version?

packages/ethereum-contracts/contracts/superfluid/Superfluid.sol
```diff
@@ -292,14 +292,17 @@ contract Superfluid is
         emit SuperTokenFactoryUpdated(_superTokenFactory);
     }
 
-    function updateSuperTokenLogic(ISuperToken token)
+    function updateSuperTokenLogic(ISuperToken token, address newLogicOverride)
         external override
         onlyGovernance
     {
-        address code = address(_superTokenFactory.getSuperTokenLogic());
+        address newLogic = newLogicOverride != address(0) ?
```
hellwolf, line 299: put a comment

packages/ethereum-contracts/contracts/interfaces/superfluid/ISuperfluidGovernance.sol
```diff
@@ -53,6 +53,15 @@ interface ISuperfluidGovernance {
     function batchUpdateSuperTokenLogic(
         ISuperfluid host,
         ISuperToken[] calldata tokens) external;
+
+    /**
+     * @dev Update supertoken logic contract to the provided logic contracts
+     * Note that this is an overloaded version taking an additional argument `tokenLogics`
```
hellwolf: use @custom: note

hellwolf (approved):

PR #1568, opened 2023-08-08
[ETHEREUM-CONTRACTS] optional hooks

d10r:


hellwolf (changes requested):
Let's do a impact analysis of adding user data param to `downgradeTo`, in order to be insync with upgradeTo.

Also, check the param name "data", and disambiguate them with either userData or operatorData.

d10r:
> Let's do a impact analysis of adding user data param to `downgradeTo`, in order to be insync with upgradeTo.

After a few conversations and thinking this through I decided to leave `downgradeTo` as is.
Changing the function signature would mean that for some (maybe a very long) time we'd have SuperTokens out there with diverging signature.  
Since `downgradeTo` can only be invoked by the account holder itself (no ACL around it), it's hard to come up with use cases for that send hook.
It's just not worth any potential trouble to change or overload it.

> 
> Also, check the param name "data", and disambiguate them with either userData or operatorData.

renamed all remaining occurances of`data` to `userData` in SuperToken.sol.

hellwolf (approved):

PR #1587, opened 2023-08-15
[METADATA] added autowrap contract addresses

d10r:
having `flowScheduler` and  `vestingScheduler` in metadata, we should also have autowrap.
This PR adds it.
Since there's 2 contracts (see https://github.com/superfluid-finance/protocol-monorepo/tree/dev/packages/automation-contracts/autowrap), I mapped that to a nested object with 2 fields.

Contract addresses are taken from https://github.com/superfluid-finance/widget/blob/master/packages/widget/wagmi.config.ts#L28 which I assume is complete.

Note: the [autowrap README](https://github.com/superfluid-finance/protocol-monorepo/tree/dev/packages/automation-contracts/autowrap) contains an incomplete list of addresses, could be removed and pointed to metadata instead in order to avoid such inconsistencies going forward.

### DIFF
--- a/packages/automation-contracts/autowrap/package.json
+++ b/packages/automation-contracts/autowrap/package.json
@@ -15,6 +15,6 @@
     "dependencies": {
         "@openzeppelin/contracts": "4.9.3",
         "@superfluid-finance/ethereum-contracts": "1.7.2",
-        "@superfluid-finance/metadata": "1.1.10"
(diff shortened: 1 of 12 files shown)


hellwolf:
changelog update needed

hellwolf (approved):

hellwolf:
if plan to release, bump the package version too

PR #1590, opened 2023-08-16
[ETHEREUM-CONTRACTS] add invariants check at the end of helper functions

### LINKED ISSUE #1592: [ETHEREUM-CONTRACTS] Invariants + Coverage
see https://github.com/superfluid-finance/protocol-monorepo/pull/1590/files

0xdavinchee:
- Add invariants check at the end of the helper functions
- Add toga to list of users


.github/workflows/call.test-ethereum-contracts.yml
```diff
@@ -84,6 +84,8 @@ jobs:
         env:
           # NOTE: This is currently unset and fork tests are not being run
```
hellwolf: What is unset and what is not being run?

If it's really the case, perhaps comment out the line instead?

0xdavinchee: I'm going to delete this because we ended up removing those fork test files anyways

hellwolf (approved):

PR #1586, opened 2023-08-15
[TOOLING] ops-scripts improvements for 1.7+

d10r:
* Allows governance upgrade if governance is not owned by EOA
* Write a versionString to Resolver after successful deployment
* Consistent order of COFNFT/CIFNFT

PR #1602, opened 2023-08-21
GDA tooling

d10r:
This branch is not on top of dev, but on top of https://github.com/superfluid-finance/protocol-monorepo/pull/1586 - thus not directly committing to gda branch.
To be merged after PR1586.


hellwolf:
https://github.com/superfluid-finance/protocol-monorepo/wiki/Smart-Contracts-Version-Tagging

PR #1624, opened 2023-08-22
[WORKFLOWS] [SUBGRAPH] Subgraph workflow simplification

d10r:
follow-up of #1580 

Simplify the subgraph deployment script.

With the addition of a 3rd deploy target (aka vendor), duplication of script and workflow code went too far.  
After looking at it, I determined that 
* the scripts doing the actual job (prepare network specific schema and invoke `graph deploy`) can easily be merged into one (eliminating redundant code)
* the vendor specific workflow files can easily be merged into one, adding an input argument "vendor"
* the workflow scripts don't need to distinguish between specific network vs all network deployment, as that's handled by the script anyway

This PR also switches the network list in `packages/subgraph/hosted-service-networks.json` to canonical names, adding the then necessary mapping to non-canonical names (due to historical reasons) to the scripts using that file. Please double-check if that breaks anything @0xdavinchee

PR #1628, opened 2023-08-23
[WORKFLOWS] [GDA] [SUBGRAPH] Subgraph workflow simplification gda

d10r:
gda version of the branch (in order to allow deploying gda subgraphs)


hellwolf:
seems redundant to me, see #1624

PR #1632, opened 2023-08-23
[METADATA] - Include ExistentialNFTCloneFactory address in Metadata for all networks

surewht:
As the title says, I have added the ExistentialNFTCloneFactory address to all networks in the metadata package.


d10r (changes requested):
please also add the new field to module/networks/list.d.ts and re-build.
And please add the existentialNFTCloneFactory fields to the end of the list (after autowrap).

surewht:
> please also add the new field to module/networks/list.d.ts and re-build. And please add the existentialNFTCloneFactory fields to the end of the list (after autowrap).

done. We should probably do alphanumeric ordering though.

d10r (approved):

PR #1656, opened 2023-09-03
[ETHEREUM-CONTRACTS] ops-scripts improvements

d10r:
improved Safe support, save gas on eth-mainnet and base-mainnet (truffle defaults overpay in priority fees)


packages/ethereum-contracts/contracts/interfaces/utils/ISafe.sol
```diff
@@ -0,0 +1,23 @@
+// SPDX-License-Identifier: MIT
+pragma solidity >=0.8.0;
```
hellwolf: ">=0.8.11" is what we test with.

hellwolf (changes requested):
a small change request, but otherwise looks good.

hellwolf (approved):

PR #1662, opened 2023-09-06
[ETHEREUM-CONTRACTS] make batchCall payable

d10r:
`Superfluid.batchCall` not being payable blocks some otherwise possible ways to use it.
This PR contains a minimal changeset which makes it payable, forwarding the value provided to the first operation of type `OPERATION_TYPE_SUPERFLUID_CALL_APP_ACTION`.

A more complete solution would allow to define the value to be forwarded per operation, [as done e.g. in Multicall3](OPERATION_TYPE_SUPERFLUID_CALL_APP_ACTION).  
That would however require major code changes and make only limited sense for SF batch calls as long as it doesn't support arbitrary calls.
This PR could be a stopgap solution for unlocking a usecase blocked by it, and with a future addition supporting arbirary call targets (see https://github.com/superfluid-finance/protocol-monorepo/issues/1038) we could provide Multicall interface compatibility which includes arbitrary value splitting.

Making `batchCall` payable shouldn't risk any breakage.

### DIFF
--- a/packages/ethereum-contracts/contracts/superfluid/Superfluid.sol
+++ b/packages/ethereum-contracts/contracts/superfluid/Superfluid.sol
@@ -625,7 +625,7 @@ contract Superfluid is
             appCreditToken: ISuperfluidToken(address(0))
         }));
         bool success;
-        (success, returnedData) = _callExternalWithReplacedCtx(address(agreementClass), callData, ctx);
+        (success, returnedData) = _callExternalWithReplacedCtx(address(agreementClass), callData, 0, ctx);
         if (!success) {
             CallUtils.revertFromReturnedData(returnedData);
         }
@@ -647,6 +647,7 @@ contract Superfluid is
     function _callAppAction(
         address msgSender,
         ISuperApp app,
+        uint256 value,
         bytes memory callData
     )
         internal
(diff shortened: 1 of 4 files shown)


packages/ethereum-contracts/contracts/superfluid/Superfluid.sol
```diff
@@ -784,6 +785,7 @@ contract Superfluid is
     )
        internal
     {
+        bool valueForwarded = false;
```
hellwolf, line 788: We should send all msg.value back if no app action is in the batch call that uses it.

packages/ethereum-contracts/contracts/superfluid/Superfluid.sol
```diff
@@ -842,7 +844,9 @@ contract Superfluid is
                 _callAppAction(
                     msgSender,
                     ISuperApp(operations[i].target),
+                    valueForwarded ? 0 : msg.value,
```
kasparkallas, line 847: The assumption is that this will revert when the given `SUPERFLUID_CALL_APP_ACTION` it's passed to is not `payable`.

packages/ethereum-contracts/contracts/superfluid/Superfluid.sol
d10r: added

packages/ethereum-contracts/contracts/superfluid/Superfluid.sol
d10r: I checked again with a dedicated test. Does actually not fail, it's really just a compiler level construct. There must have been a misleading red herring yesterday when I checked.
Thinking about it, that makes sense, because otherwise the introduction of payable in Solidity would have been a breaking change.

d10r: yes, I added a test case for that.

hellwolf (approved):

hellwolf:
Update changelog

PR #1664, opened 2023-09-07
[ETHEREUM-CONTRACTS] Make TestToken support permit

0xdavinchee:
- inherit from OZ's ERC20Permit instead of just ERC20


hellwolf (approved):
good.

PR #1666, opened 2023-09-08
[ETHEREUM-CONTRACTS] Additional SuperTokenV1Library comment

0xdavinchee:
This PR adds a comment for the `SuperTokenV1Library` documentation which hints to the user to use vm.startPrank over vm.prank when using the library if the cache isn't warmed up.
We encountered this during the internal hackathon and realized it would be good to provide a hint for developers here.


hellwolf (approved):
k

PR #1681, opened 2023-09-18
[ETHEREUM-CONTRACTS/SDK-CORE] Remove typechain from build output

### LINKED ISSUE #1610: [ETHEREUM-CONTRACTS] Remove typechain-types as shipped build
Currently the ethereum-contracts package is responsible for shipping both the ABI and other things such as the typechain (for ethers-v5) types consumed by SDK-Core, etc. 
This is breaking.
The issue with this is we are assuming that downstream packages will be using ethers-v5. With ethers-v6 released, wagmi, viem and new frameworks which will likely come out and replace these in the future, the most pragmatic thing we can do is to ship the most fundamental building block: the ABIs. This allows us to be as framework agnostic as possible.
Internally, the responsibility of handling the ABIs for outputting the required file type will be placed on the consuming package (e.g. SDK-Core, etc.)

0xdavinchee:
## High level notes
- ethereum-contracts no longer ships typechain types in the build (BREAKING)
- as part of the sdk-core build process, we have a new script (`tasks/build-types.sh`):
  - copy typechain-types folder from ethereum-contracts (when contracts are built) to sdk-core
  - compile the types using tsc


hellwolf (approved):

packages/sdk-core/tasks/build-types.sh
```diff
@@ -0,0 +1,22 @@
+#!/usr/bin/env bash
+
+# make sure that if any step fails, the script fails
+set -xe
+
+if [ -d "./src/typechain-types" ]; then
+  rm -rf ./src/typechain-types
```
hellwolf: this is redundancy, `rm -rf` is sufficient and idomatic.

packages/sdk-core/tasks/build-types.sh
```diff
@@ -0,0 +1,22 @@
...
+fi
+
+# if the typechain files do not exist, we build
+# hardhat so that it does exist
+if [ ! -d "../ethereum-contracts/typechain-types" ]; then
+  cd ../ethereum-contracts
+  yarn build:contracts:hardhat
```
hellwolf: leaking abstraction: a nicer error message when typechain-types not available suffices.

PR #1690, opened 2023-09-21
[ETHEREUM-CONTRACTS] 1685 devolved supertokens

### LINKED ISSUE #1685: [ETHEREUM-CONTRACTS] Add support for devolution of SuperTokens
When deploying a SuperToken, one currently only has a choice between it being upgradable by SF gov, or not upgradable.
In order to retain the role of upgrade admin, one has to deploy a full copy of SuperToken, with `updateCode` slightly changed. (see https://github.com/superfluid-finance/protocol-monorepo/issues/1436)

We can allow new and existing SuperTokens to retain the upgrade admin role:
* add a storage slot for `ADMIN_OVERRIDE` (not sure about naming). Could also be a struct which for now has only an address field, may use the remaining 12 bytes later e.g. to add flags.
* add a modifier `onlyAdmin` which resolves to `ADMIN_OVERRIDE` if non-zero, to the host address otherwise.
* modify `SuperToken.updateCode` to be permissioned with `onlyAdmin` instead of hardcoded to the host.
* add an argument `admin` to `SuperToken.initialize()` (or override it in order not to break existing code).
This is enough to allow creation of devolved SuperTokens, e.g. by OpenDollar.

In order to additionally allow devolution of existing SuperTokens, and to allow devolved SuperTokens to change the admin, we also need something along the lines of:
* `SuperToken.changeAdmin() onlyAdmin ...`
* `Superfluid.changeSuperTokenAdmin() onlyGovernance`
* `SuperfluidGovernance.changeSuperTokenAdmin()`

With that in place, we should also add an example for how to deploy a devolved SuperToken, and maybe also mention it in the docs.

Stretch goal: Add support for devolved SuperTokens in `SuperTokenFactory`.
On that occasion we could also add a method for deploying Pure SuperTokens (to be specified in separate ticket).

0xdavinchee:
See https://github.com/superfluid-finance/protocol-monorepo/issues/1685


packages/ethereum-contracts/contracts/mocks/SuperTokenMock.sol
```diff
@@ -59,8 +59,8 @@ contract SuperTokenStorageLayoutTester is SuperToken {
         require (slot == 18 && offset == 0, "_operators changed location");
         // uses 4 slots
 
-        assembly { slot:= _reserve22.slot offset := _reserve22.offset }
-        require (slot == 22 && offset == 0, "_reserve22 changed location");
+        assembly { slot:= _reserve23.slot offset := _reserve23.offset }
```
hellwolf, line 65: `_adminOverride` field should be tested.

packages/ethereum-contracts/test/contracts/superfluid/SuperToken.NonStandard.test.ts
```diff
@@ -92,13 +92,18 @@ describe("SuperToken's Non Standard Functions", function () {
...
                 superToken,
-                "SUPER_TOKEN_ONLY_HOST"
+                "SUPER_TOKEN_ONLY_ADMIN"
             );
         });
 
         it("#1.4 only can initialize once", async () => {
             await expectRevertedWith(
-                superToken.initialize(ZERO_ADDRESS, 18, "name", "symbol"),
+                superToken["initialize(address,uint8,string,string)"](
```
d10r: Are we really going there? Feels a bit like a breaking change, although technically it isn't.

packages/ethereum-contracts/test/foundry/gov/SuperfluidGovernanceII.t.sol
d10r, line 1: foundry ftw!

packages/ethereum-contracts/contracts/superfluid/Superfluid.sol
```diff
@@ -304,6 +304,10 @@ contract Superfluid is
         emit SuperTokenLogicUpdated(token, newLogic);
     }
 
+    function changeSuperTokenAdmin(ISuperToken token, address newAdmin) external onlyGovernance {
```
hellwolf, line 307: shouldn't the admin be the manager here when it's set? test cases should rehearse  that transition of power over time too.

packages/ethereum-contracts/test/contracts/superfluid/SuperToken.NonStandard.test.ts
0xdavinchee: yup, hard to avoid this tbh

packages/ethereum-contracts/contracts/mocks/SuperTokenMock.sol
0xdavinchee: good catch 🔍

packages/ethereum-contracts/contracts/superfluid/Superfluid.sol
hellwolf: okay, so the `changeAdmin` is correct then. nvm

packages/ethereum-contracts/contracts/interfaces/superfluid/ISuperfluid.sol
```diff
@@ -233,6 +233,12 @@ interface ISuperfluid {
      */
     event SuperTokenLogicUpdated(ISuperToken indexed token, address code);
 
+    /**
+     * @notice Change the SuperToken admin override
+     * @dev The admin override is the only account allowed to update the token logic
```
hellwolf, line 238: >> For backward compatibility, the "host" usually is the default "admin"

packages/ethereum-contracts/test/contracts/superfluid/SuperToken.NonStandard.test.ts
hellwolf: avoid name conflict is also an option sometimes: name the new function initializeWithAdminOverride

0xdavinchee: fair-we can do this instead

hellwolf (approved):
LET"S GO

PR #1693, opened 2023-09-22
Add listed super tokens and superTokenLogic contracts to the verification message

elvijsTDL:


hellwolf:
@elvijsTDL wanna give it  a trial spin before we merge?

tasks/package.json
```diff
@@ -0,0 +1,5 @@
+{
+    "dependencies": {
+        "ethers": "6.7.1"
```
hellwolf: huh??

elvijsTDL: Using it for getting superTokenLogic contract addresses from superTokenFactory contracts ,because they are not available in metadata and @0xdavinchee suggested to better query the contract as the logic addresses might change

elvijsTDL:
> @elvijsTDL wanna give it a trial spin before we merge?

Yup, ill manually trigger it first

tasks/package.json
hellwolf: okay, I don't like that we do it here. Let me suggest another way.

.github/workflows/daily-slack-message.yml
```diff
@@ -16,8 +16,14 @@ jobs:
...
 
+      - name: Install dependencies
+        working-directory: tasks
+        run: npm install
+        
+
       - name: Send slack message
-        run: node tasks/daily-slack-bot.js 
+        working-directory: tasks
+        run: node daily-slack-bot.js
```
hellwolf:
```
run: |
   npm install ethers
   node daily-slack-bot.js
```

hellwolf (approved):
yes ser

PR #1700, opened 2023-09-26
[ETHEREUM-CONTRACTS] optional logic extension via eip-1967-adjacent storage slot for custom super tokens

d10r:
Problem this could solve (if it works as expected):
Allow custom SuperTokens to add upgradable logic without cloning the whole SuperToken contract or implementing custom call routing.


d10r:
As @hellwolf correctly pointed out, we can't use the eip-1967 storage slot, as it's already taken for the primary implementation pointer.  
Since the eip-1967 storage slot is
```
bytes32(uint256(keccak256('eip1967.proxy.implementation')) - 1)
```
we now use
```
bytes32(uint256(keccak256('eip1967.proxy.implementation')) - 2)
```
for the 2nd level implementation pointer.  
While this isn't a storage slot defined by eip-1967, it could potentially be a natural extension to the eip for supporting multi-level proxies.

hellwolf:
@hellwolf will have a look this week.

hellwolf:
>  While this isn't a storage slot defined by eip-1967, it could potentially be a natural extension to the eip for supporting multi-level proxies.

This seems too farfetched, and the downside of having conflict with 1967 in the future outweighs the possibility of merging with our thinking.

I suggest we choose an entirely different way.

But before we decide one way or the other, please make a attempt of suggesting such idea at https://ethereum-magicians.org/t/eip-1967-standard-proxy-storage-slots/3185 and we let's give it a week or two.

d10r:
asked for feedback in https://ethereum-magicians.org/t/eip-1967-standard-proxy-storage-slots/3185/25

PR #1701, opened 2023-09-27
[SUBGRAPH] Updated sf and satsuma supported network lists

mmd-afegbua:
PR adds to the list of networks supported by self hosted subgraph nodes. Also adds base-mainnet to list of networks for Satsuma.


0xdavinchee (approved):
LGTM

PR #1705, opened 2023-09-28
[ETHEREUM-CONTRACTS] PureSuperToken downgrade logic fixed

### LINKED ISSUE #1258: [ETHEREUM-CONTRACTS] ERC-777 burn won't work for pure SuperTokens
always invokes downgrade().
should instead just burn if underlying is zero

0xdavinchee:
- We no longer revert when downgrading pure super tokens


hellwolf (commented):
Looks good to me.

@d10r please have a look too.

PR #1706, opened 2023-09-29
[ETHEREUM-CONTRACTS] SuperApp registerApp deprecation

### LINKED ISSUE #1660: [ETHEREUM-CONTRACTS] [DEVX] simplify SuperApp registration
The registration key doesn't serve any purpose (anymore) after switching to an expire date based logic instead of one-time keys.
As a devx improvement, we could now simplify the logic to this:

* `registerApp(configWord)` for registration through the constructor, invoked by a deployer.
* `registerApp(app, configWord)` for registration through a factory contract.

Overloading isn't an issue in the Solidity API (it's a devx issue only with e.g. ethers.js).

`registerAppWithKey` and `registerAppByFactory` can be kept for a while marked as deprecated (for backwards compatibility) and later removed. But docs and examples shouldn't use them anymore.

Internally (in governance contract and scripts) we can either keep the logic as is (declaring e.g. the empty string as default key) or simplify the code. It could be simplified to just whitelist deployers, regardless of EOA or contract.  
If we want to keep the distinction and associated constraints (EOA: call only from constructor), that could be achieved implicitly by checking if the msg.sender is an EOA or contract in `registerApp`.

0xdavinchee:


packages/ethereum-contracts/contracts/interfaces/superfluid/ISuperfluid.sol
```diff
@@ -73,15 +73,13 @@ interface ISuperfluid {
     // uses SuperAppDefinitions' App Jail Reasons as _code
     error APP_RULE(uint256 _code);                              // 0xa85ba64f
 
-    error HOST_INVALID_OR_EXPIRED_SUPER_APP_REGISTRATION_KEY(); // 0x19ab84d1
     error HOST_NOT_A_SUPER_APP();                               // 0x163cbe43
-    error HOST_NO_APP_REGISTRATION_PERMISSIONS();               // 0x5b93ebf0
+    error HOST_NO_APP_REGISTRATION_PERMISSION();                // 0x5b93ebf0
```
hellwolf: update the enclosed four bytes:

double check:

```
$ cast sig 'HOST_NO_APP_REGISTRATION_PERMISSIONS()'
0x5b93ebf0

$ cast sig 'HOST_NO_APP_REGISTRATION_PERMISSION()'
0xb56455f0

```

0xdavinchee: changed

d10r (approved):

hellwolf (approved):
Please add a CHANGELOG entry then we are good to go

PR #1714, opened 2023-10-06
[SUBGRAPH-OPS] Update deploy.sh

0xdavinchee:
The list of `SATSUMA_NETWORKS` should include the canonical names as we have a `legacyNetworkNames` which maps it to the correct name anyways.

Our `buildNetworkConfig.ts` script (we removed hardcoded configs and are building it based off metadata) expects the canonical name (the one in metadata) otherwise it won't create the .json file and the subgraph deployment will fail.


hellwolf (approved):
I wonder if you could do those shellcheck disabling in that file level.

I guess they were disabled because this script was "sourced" in another file?

PR #1718, opened 2023-10-16
[SDK-CORE] patch getUnderlyingToken

0xdavinchee:
- this fixes an issue first reported on slack [here](https://superfluidhq.slack.com/archives/C02K8UDDLR0/p1697137370913019?thread_ts=1697123027.550069&cid=C02K8UDDLR0) where supertoken creation is broken because of the assumption that getUnderlyingToken exists


hellwolf (approved):

kasparkallas:
I'd like to see the try-catch behaviour only for code "CALL_EXCEPTION" with a comment. Otherwise there's a chance that a random RPC issue will define _any_ token as a "pure super token" which could create race-condition-y tough to debug issues.

PR #1730, opened 2023-10-20
[ETHEREUM-CONTRACTS] make deploy script compatible with ethers v6

0xdavinchee:


packages/ethereum-contracts/dev-scripts/deploy-test-framework.js
```diff
@@ -65,21 +76,36 @@ const _getFactoryAndReturnDeployedContract = async (
...
+ * NOTE: This only works with Hardhat + ethers v5/ethers v6 currently.
+ * @param privateKey NEVER USE A PRIVATE KEY WITH REAL FUNDS - a test account private key
+ * @param provider an ethers provider
+ * @param ethersV5Signer an ethers v5 signer
  * @returns
  */
-const deployTestFramework = async () => {
-    const signer = (await ethers.getSigners())[0];
-    await deployERC1820(ethers.provider);
+const deployTestFramework = async (privateKey, provider, ethersV5Signer) => {
```
d10r: I think this interface is a bit confusing. Can we change it such that it better conveys that either privateKey or ethersV5Signer is needed?

0xdavinchee: one of the downsides of javascript :/

do you have something in mind?

d10r (approved):

hellwolf (approved):
fantastic!

PR #1754, opened 2023-12-01
[SDK-core] map the name from subgraph to unknown entity

kasparkallas:
**Why?**
Vastly improves the look of the events tab for the Console's GDA PR: https://github.com/superfluid-finance/superfluid-console/pull/190 (all the GDA events are displayed as "_Unknown" otherwise)


0xdavinchee (approved):
nice - LGTM

PR #1765, opened 2023-12-22
[ETHEREUM-CONTRACTS] Protocol Ops Fixes

d10r:
* [info-print-contract-addresses.js](https://github.com/superfluid-finance/protocol-monorepo/compare/script-fixes?expand=1#diff-3d06557d774ed3284e8a2a99bda0e9ec31d345a129baa4423af3eac9911f140b): skip FlowNFTs if not deployed instead of crashing
* [buildNetworkConfig.ts](https://github.com/superfluid-finance/protocol-monorepo/compare/script-fixes?expand=1#diff-0b015c7c3dfa5c9afe6cf3666ddd65ac14311a328c27fc2a6a530792dc4b18cb): don't create wrong path if "subgraph" is somewhere in the path below where it's expected


hellwolf (approved):

d10r:
explanation to what was the issue with [buildNetworkConfig.ts](https://github.com/superfluid-finance/protocol-monorepo/compare/script-fixes?expand=1#diff-0b015c7c3dfa5c9afe6cf3666ddd65ac14311a328c27fc2a6a530792dc4b18cb):

I was running it with the monorepo root being `~/src/protocol-monorepo/b/subgraph-deploy` (using git worktree).
The old code 
```
    const writeToDir = __dirname.split("subgraph")[0] + `subgraph/config/${networkName}.json`;
```

had an implicit assumption that the string "subgraph" would not occur in the path, other than what it was looking for, and would this result in a wrong result and failure if that assumption didn't hold.

PR #1797, opened 2024-01-22
[ETHEREUM-CONTRACTS] Forked truffle-plugin-verify so we can verify the new networks

d10r:
eventually we need to migrate away, but for now we need a short-term fix for the new networks, which is to use a fork of the plugin.


hellwolf:
Keep truffle alive! :D

hellwolf (approved):

PR #1818, opened 2024-01-31
fix: divide by zero subgraph indexing error

kasparkallas:
the error:
image


0xdavinchee (approved):
thanks @kasparkallas!

PR #1823, opened 2024-02-06
[sdk-core] use different semver for metadata

kasparkallas:
### Why?
It' unnecessary to lock down the version to a specific version, and couple every new metadata release to a new sdk-core release.

EDIT:
* I can't update metadata for older sdk-cores. Currently I'm forced into the dev-package to get the latest information. Even though the newer metadat is not breaking to the older SDK.
* If I have metadata specified as a dependency in my app then there's a high likelihood that sdk-core will get a different version from the app.

### What?
Use looser semver for specifying metadata from sdk-core.

### How?
Allow minor and patch updates.


packages/sdk-core/package.json
```diff
@@ -57,7 +57,7 @@
     },
     "dependencies": {
         "@superfluid-finance/ethereum-contracts": "1.9.0",
```
kasparkallas: This is also locked down, and interestingly 1.9.0 is not released yet

hellwolf: We have talked about this. This is the currently how we test sdk-core compatibility with the unreleased ethereum-contracts.

Happy to open this topic in the future again, but not changing workflow of this for now.

hellwolf:
Can you restate the "why" with less opinion but more the problem to be fixed, so that we can figure out a way forward?

kasparkallas:
1. I can't update metadata for older sdk-cores. Currently I'm forced into the dev-package to get the latest information. Even though the newer metadat is not breaking to the older SDK.
2. If I have metadata specified as a dependency in my app then there's a high likelihood that sdk-core will get a different version from the app.

hellwolf:
> I can't update metadata for older sdk-cores. Currently I'm forced into the dev-package to get the latest information. Even though the newer metadat is not breaking to the older SDK.

yes, this is a problem.

I think the current way of putting metadata part of the work space is wrong. Let me think of a change to it.

hellwolf:
```
$ git grep -- '@superfluid-finance/metadata' | grep package.json
packages/automation-contracts/autowrap/package.json:        "@superfluid-finance/metadata": "1.1.27"
packages/automation-contracts/scheduler/package.json:        "@superfluid-finance/metadata": "1.1.27"
packages/ethereum-contracts/package.json:        "@superfluid-finance/metadata": "1.1.27",
packages/js-sdk/package.json:        "@superfluid-finance/metadata": "1.1.27",
packages/metadata/package.json:  "name": "@superfluid-finance/metadata",
packages/metadata/package.json:    "build": "./build.sh && echo '@superfluid-finance/metadata build successfully'",
packages/sdk-core/package.json:        "@superfluid-finance/metadata": "1.1.27",
packages/subgraph/package.json:        "@superfluid-finance/metadata": "1.1.27",
```

One possibility is to make all these dependencies peer dependencies instead.

hellwolf (approved):

hellwolf (approved):

Issue #1782, opened 2024-01-18
[ETHEREUM-CONTRACTS] TrustedMacros: A TrustedForwarder that rules them all


hellwolf:
## Overview
**What?**

This is a trusted forwarder with high degree of extensibility through permission-less and user-defined "macro contracts", where:

* Macro contracts encapsulates a high level ABI call wrapper, which compresses call data as a result.
  * Benefit: it is good for L2 where calldata size matters when it comes to transaction cost.
* Additionally, it makes it a much easier and consistent interface for integrating eip-712.
   *Benefit: the transaction is more readable than otherwise still having to users to read less legible eip-712 transaction with naive eip-712 integration.

An example user-defined macro request using eip-712:

![image](https://github.com/superfluid-finance/protocol-monorepo/assets/186660/41b4d8a0-2cd4-4165-953d-03e37ffda16a)

**How?**

1. Upon requested by the TrustedMacro contract, a user-defined macro should return a array of Superfluid batch operations, which will be send to trustedForwardBatchCall by the TrustedMacro.
2. A user-defined macro is stateless but still can query other external contracts, which eliminates a class of security issues without compromising extensibility completely.

[missing image]

**Why?**

* Allow apps to compress calldata for L2s (!), without having to register a bespoke trusted forwarder.
* Allow uniformed out-of-box integration of eip-712 legible signature!

**Deadline:**

* Prototype ready: soon.
* Integration to an application: end of Jan.

**Technical Design Document (TDD):**

## Acceptance criteria
 - [x] Prototype demoed
 - [x] Test-integration with an live app
 - [ ] Satisfies entire acceptance criteria
 - [ ] Is the appetite for the spike exhausted?
 - [ ] Documentation updated
 - [ ] Presentation to stakeholders
 - [ ] Output is a TDD
 - [ ] Recommendation on next steps

## Links/Resources
* https://superfluidhq.slack.com/archives/C04V40RFJQ7/p1704895433059129
* https://github.com/superfluid-finance/protocol-monorepo/issues/68
* https://weijiekoh.github.io/eip712-signing-demo/index.html
* https://github.com/delegatable/delegatable-sol

hellwolf:
An example user-defined macro request using eip-712:

![image](https://github.com/superfluid-finance/protocol-monorepo/assets/186660/41b4d8a0-2cd4-4165-953d-03e37ffda16a)

hellwolf:
[...]

PR #1828, opened 2024-02-13
[ETHEREUM-CONTRACTS] proposed changes for the macro forwarder

d10r:
Proposed changes:
* rename `TrustedMacrosVanilla` to `MacroForwarder`. Reasoning: _trusted_ is a property not inherent to the contract, but coming into effect by registering it as trusted by gov. _Vanilla_ isn't self-explanatory, I find it unnecessary. If we add a 712 variant, we can simply add _712_ to the name. Plural -> singular because the use of a variable `trustedMacros` for an instance in the test seemed odd.
* Removal of `simulateMacro: Initially I tried to find a better name (_simulate_ to me suggests that it would simulate the actual execution, not just retrieve the operations), but I came to the conclusion that it's more confusing than helpful and should instead be removed. It doesn't really add anything, because it just delegates to `macro.buildBatchOperations()`
* Added test case using `StatefulMacro`: https://github.com/superfluid-finance/protocol-monorepo/issues/1782 states: _A user-defined macro is stateless ..._ - I think that's a mistake. The macro may very well have state if that helps to reduce the calldata size of the macro call. It just can't modify state in `buildBatchOperations()`. This test case exemplifies that, reducing the calldata to the minimum, which currently is the macro address.


hellwolf:
> rename TrustedMacrosVanilla to MacroForwarder. Reasoning: trusted is a property not inherent to the contract, but coming into effect by registering it as trusted by gov. Vanilla isn't self-explanatory, I find it unnecessary. If we add a 712 variant, we can simply add 712 to the name. Plural -> singular because the use of a variable trustedMacros for an instance in the test seemed odd.

okay.

> Removal of simulateMacro: Initially, I tried to find a better name (_simulate_ to me suggests that it would simulate the actual execution, not just retrieve the operations), but I came to the conclusion that it's more confusing than helpful and should instead be removed. It doesn't really add anything, because it just delegates to macro.buildBatchOperations()`

I should probably still expose it, just for the merit of being able to get those batch operations call data. But it wouldn't work if it became stateful.

> Added test case using StatefulMacro: https://github.com/superfluid-finance/protocol-monorepo/issues/1782 states: A user-defined macro is stateless ... - I think that's a mistake. The macro may very well have state if that helps to reduce the calldata size of the macro call. It just can't modify state in buildBatchOperations(). This test case exemplifies that, reducing the calldata to the minimum, which currently is the macro address.

Not necessarily a mistake, but it was a conscious and defensive choice.

However, we can rethink here: was the defensiveness justified?

Let me think a bit.

hellwolf (approved):

PR #1842, opened 2024-02-28
[ETHEREUM-CONTRACTS] never increase buffer when decreasing flowrate

d10r:


hellwolf:
Thanks!

Now, we need to fix tests and add new tests!

hellwolf:
Some fixes needed and need a lot of strengthening of tests: `CFA_FLOW_ALREADY_EXISTS` errors start to occur a lot, suspecting logic bug here, and should be easy to reproduce with appropriate tests.

hellwolf:
outdated PR.

PR #1843, opened 2024-02-29
[SUBGRAPH] fix resolver entry

0xdavinchee:
we move this: `resolverEntry.isToken = Token.load(target.toHex()) != null;` out of the `resolverEntry == null` and only update it if the `target` is not null.

We need to do this because of sepolia where the token entity is created after the resolver set event is emitted (triggered), thus we want to be able to call set again and fix this. Edited: This makes ETHx not shown as listed in sepolia subgraph.


packages/subgraph/src/mappings/resolver.ts
```diff
@@ -47,14 +42,34 @@ export function handleRoleRevoked(event: RoleRevoked): void {
...
+        resolverEntry.createdAtTimestamp = block.timestamp;
+        resolverEntry.targetAddress = target;
+    }
+    const isListed = target.notEqual(ZERO_ADDRESS);
+
+    // we only update this if the target is not equal to the zero address
+    if (isListed) {
+        resolverEntry.isToken = Token.load(target.toHex()) != null;
+    }
+    resolverEntry.updatedAtBlockNumber = block.number;
```
kasparkallas, line 60: is it ok to always update when `getOrInit` is called? it's called in very few places?

kasparkallas (approved):
acknowledged

hellwolf:
how is this one doing

hellwolf:
the fix did not work, will be investigated.

hellwolf:
@elvijsTDL please write a concrete bug report on the issue related to this.

elvijsTDL:
The subgraph shows some of the tokens as unlisted even if they should be e.g. ETHx , fUSDCx

If you query the listed tokens you can see that they are missing:

``` query MyQuery {
  tokens(where: {isListed: true}) {
    isListed
    isSuperToken
    isNativeAssetSuperToken
    name
    symbol
  }
}
```

Screenshot 2024-03-22 at 11 56 16

At some point it has flipped from listed to unlisted as it used to work fine 

This might be related to GoodDollar G$ getting unlisted from Celo too

PR #1848, opened 2024-03-04
[PROTOCOL-MONOREPO] remove deprecated networks

d10r:
See https://github.com/superfluid-finance/superfluid-stories/issues/642


packages/metadata/main/networks/list.cjs
```diff
@@ -3,6 +3,7 @@ module.exports =
 [
     {
         "name": "eth-goerli",
+        "isDeprecated": true,
```
hellwolf: for actual deprecation, there needs a grace period. In absence of it, we either use a ticket to track, or we simply remove it here.

packages/subgraph/README.md
```diff
@@ -20,51 +20,10 @@ Get started using the Subgraphs with an introduction and query examples:
 
 https://docs.superfluid.finance/superfluid/docs/subgraph
 
-# 👨‍🚀 Hosted Subgraphs
```
hellwolf, line 23: What was the rationale behind the removal? I am for it, but it would be good to have a written notes from your thoughts @d10r .

packages/metadata/main/networks/list.cjs
d10r: Originally we said that eth-goerli will be removed by end of March, but that detail seems to have got lost.
fait accompli now, so will undo this.

packages/subgraph/README.md
d10r: A few reasons:
* the info here (_All subgraphs are available via The Graph's hosted service:_) was stale and has become wrong since last updated, because not all subgraphs are available via the hosted service anymore
* there's no / very little explanation of the concepts used (v1 endpoits, dev endpoints, feature endpoints)
* the dev endpoints are at least partially unmaintained / lagging behind v1 endpoints

I could have replaced by something else, but don't currently have enough information to do so and didn't want to get derailed from the actual task of deprecating networks. We can anytime do in a dedicated PR.

I also briefly checked with @0xdavinchee if he's ok with this.

hellwolf (approved):

hellwolf:
image

d10r:
image

PR #1878, opened 2024-03-06
[SUBGRAPH] approval mapping

### LINKED ISSUE #1834: [SUBGRAPH] Investigate issues with `Approval` event mapping
### What?
The ERC-20 `Approval` event should be mapped to the subgraph. The `addresses` field should contain the relevant addresses for the user products to show the event in activity history views.

Currently, the event doesn't show up. At a quick glance, there is something off with the mapping. There seems to be two events with the same name:
* approval for ERC-20 allowance
* approval for FlowNFT

The mapping for FlowNFT approval event definitely has `addresses` missing.

I didn't find the mapping for ERC-20 approval event. I could only find it being handled for the `FlowOperator` entity: https://github.com/superfluid-finance/protocol-monorepo/blob/fa5a75dbaf85af9f60a03b224a9c29f1bf214fa7/packages/subgraph/src/mappings/superToken.ts#L454

When querying from Mumbai, this error shows up:

### Why?
The ERC-20 approval event is important and should be properly mapped to the subgraph.

0xdavinchee:
- create `ApprovalEvent` when `Approval` event is emitted from a SuperToken contract.

closes #1834


packages/subgraph/src/mappings/superToken.ts
```diff
@@ -452,6 +453,17 @@ function _createTransferEventEntity(event: Transfer): void {
 }
 
 export function handleApproval(event: Approval): void {
+    const eventId = createEventID("Approval", event);
+    const ev = new ApprovalEvent(eventId);
+    initializeEventEntity(ev, event, []);
```
kasparkallas: relevant addresses need to go here (token, owner, to)

same goes for the other approval event

kasparkallas:
We use the same event entity for the ERC-20 Approval and the NFT Approval? Why not use another entity?

0xdavinchee:
> We use the same event entity for the ERC-20 Approval and the NFT Approval? Why not use another entity?

we can, no strong reason to use the same for both

kasparkallas:
> > We use the same event entity for the ERC-20 Approval and the NFT Approval? Why not use another entity?
> 
> we can, no strong reason to use the same for both

we decided to re-use the event

packages/subgraph/src/mappings/superToken.ts
kasparkallas: Didn't change yet? Should do a quick search of `initializeEventEntity(ev, event, []);` over the code base to quickly find the spots where adding the related addresses might be missing.

0xdavinchee: yeah good point

kasparkallas (approved):

hellwolf:
test is broken

kasparkallas:
> test is broken

ah, it's the test that's currently supposed to fail, will ignore

image

hellwolf:
> ah, it's the test that's currently supposed to fail, will ignore

So we shall wait for the fix, from some future PR. Right?

kasparkallas:
> > ah, it's the test that's currently supposed to fail, will ignore
> 
> So we shall wait for the fix, from some future PR. Right?

future PR, we're already working it

created a feature branch (`2024-03-subgraph-gda-feature-branch`) to merge all the latest updates/fixes we want to finally test and release together

PR #1880, opened 2024-03-06
[ETHEREUM-CONTRACTS] Fix nfthooks outofgas

### LINKED ISSUE #1879: [ETHEREUM-CONTRACTS]  flow nft hooks can fail with outofgas
## What & Why

The FlowNFT hooks were implemented wrapped in try/catch out of caution, because we don't want the whole tx to fail for any reason related to NFT minting/burning.  
The intention was to have the same behaviour as SuperApp hooks: if the hook invocation reverts with outofgas, the whole tx should revert. If it reverts for other reasons, the tx should not revert.

Yet we see non-reverting transactions with the NFT hook reverting due to outofgas, for example [this](https://dashboard.tenderly.co/tx/optimistic/0xf8c9fe43d1836afb280dfece4c8806f4604b1a6e6657685971782d77a1736e51/debugger?trace=0.1.0.2.4.1.0.1.0.1.5).

This is likely to happen due to the way geth gas estimation works: it does a binary search for the lower limit of gas required (see [source](https://github.com/ethereum/go-ethereum/blob/master/eth/gasestimator/gasestimator.go#L53)).

## AC

- [ ] test case reproducing the issue, ideally by fuzzing
- [ ] fix

d10r:
The fuzz test nicely reproduces the issue.
I couldn't fix it with the try/catch in place.
My proposal is to remove the try/catch. The execution is conditional on the FlowNFT addresses being set, so won't be attempted if none is set. As far as I remember, the decision to try/catch those hooks was not taken out of conviction, but more in a "why not" way, assuming it won't cause any issues.


packages/ethereum-contracts/test/foundry/agreements/ConstantFlowAgreementV1.t.sol
```diff
@@ -17,11 +18,55 @@ contract ConstantFlowAgreementV1IntegrationTest is FoundrySuperfluidTester {
...
+
         _warpAndAssertAll(superToken);
 
         _helperCreateFlow(superToken, bob, alice, flowRate);
 
         _warpAndAssertAll(superToken);
     }
+
+    // there should be no gas limit which causes the NFT hook to fail with the tx succeeding
+    function testNFTHookOutOfGasRevertsWholeTx(uint256 gasLimit) public {
```
hellwolf, line 30: did you observe any failure without the fix with such fuzzing?

d10r: yes, reliably failed.
Example:
```
[FAIL. Reason: CFA_NFT_INVALID_TOKEN_ID(); counterexample: calldata=0xa34c991a00000000000000000000000000000000000000000000000000000000000060c4 args=[24772 [2.477e4]]] testNFTHookOutOfGasRevertsWholeTx(uint256) (runs: 27, μ: 415430, ~: 419640)
Logs:
  Bound Result 424774

Test result: FAILED. 0 passed; 1 failed; 0 skipped; finished in 55.35ms
```

hellwolf:
Question without reviewing it holistically yet: if it deletes a NFT that didn't exist (minting failed), would this revert? @d10r @0xdavinchee

hellwolf (changes requested):
1. An ad-hoc fork testing of the case where NFT was missing uploaded as a gist or temp repo.
2. A code review from protocol engineers.

d10r:
https://github.com/d10r/sf-flownft-hook-change-test

```
[⠢] Compiling...
No files changed, compilation skipped

Running 1 test for test/CFAv1MakeFlowNFTHooksMandatoryTest.t.sol:SFAppTest
[PASS] testUpdateDeleteFlowWithMissingNFT() (gas: 5084163)
Logs:
  deploying new CFA
  updateContracts
  curFr 6390128600823050
  curFr 0

Test result: ok. 1 passed; 0 failed; 0 skipped; finished in 6.12s
```

hellwolf (approved):

PR #1883, opened 2024-03-07
[ETHEREUM-CONTRACTS] don't claim as a side effect of updateMemberUnits()

### LINKED ISSUE #1772: [GDA] [ETHEREUM-CONTRACTS] GDA Flow Distribution Semantics Friction
see https://github.com/d10r/sf-gda-issues

d10r:
Currently `pool.updateMemberUnits()` does also `claimAll()`.
I have commented before that I consider that to be a bad idea, as it's an unexpected side effect and imo has more cons than pros.

I sense no opposition against this change and think it should be done, thus attempting a PR.

PS: some tests are failing. Before spending time on those I want to understand if we agree on making this change.


kasparkallas:
What about doing it for the pool member only when the units go to 0?

d10r:
> What about doing it for the pool member only when the units go to 0?

Dunno, that would just replace one kind of special case with another one.

I'd instead consider adding some configurability around _claim_, e.g. to make it permissioned by the receiver, maybe with an option for the admin to recover unclaimed funds.
But that's out of scope of this PR.

hellwolf (approved):
goodbye, side effects:

image

PR #1903, opened 2024-03-21
[ETHEREUM-CONTRACTS] User Macro: added postCheck and recommendation for getParams

### LINKED ISSUE #1886: [ETHEREUM-CONTRACTS] MacroForwarder postCheck hook
## What & Why

After execution, the MacroForwarder shall invoked a hook `postCheck()`.
This allows user macros to verify the state post execution and allow them to revert the transaction if they _don't like_ (sloppy language here) the outcome.

## AC
- [x] extended interface
- [ ] test coverage

d10r:
see https://github.com/superfluid-finance/protocol-monorepo/issues/1886

missing: good test case


packages/ethereum-contracts/contracts/interfaces/utils/IUserDefinedMacro.sol
```diff
@@ -18,4 +18,34 @@ interface IUserDefinedMacro {
...
     function buildBatchOperations(ISuperfluid host, bytes memory params, address msgSender) external view
         returns (ISuperfluid.Operation[] memory operations);
+
+    /**
+     * @dev A post-check function which is called after execution.
+     * It allows to do arbitrary checks based on the state after execution,
+     * and to revert if the result is not as expected.
+     * Can be an empty implementation if no check is needed.
+     */
+    function postCheck() external view;
```
hellwolf: to be consistent, let's pass ISuperfluid host too.

packages/ethereum-contracts/contracts/utils/MacroForwarder.sol
```diff
@@ -33,6 +33,7 @@ contract MacroForwarder is ForwarderBase {
     function runMacro(IUserDefinedMacro m, bytes calldata params) external returns (bool)
     {
         ISuperfluid.Operation[] memory operations = buildBatchOperations(m, params);
+        m.postCheck();
```
hellwolf: move this down, it is a "post check".

packages/ethereum-contracts/test/foundry/utils/MacroForwarder.t.sol
```diff
@@ -87,6 +97,20 @@ contract MultiFlowDeleteMacro is IUserDefinedMacro {
...
+
+contract MacroWithRevertingPostCheck is IUserDefinedMacro {
+    function buildBatchOperations(ISuperfluid, bytes memory, address /*msgSender*/) external override pure
+        returns (ISuperfluid.Operation[] memory /*operation*/)
+    {
+        return new ISuperfluid.Operation[](0);
+    }
+
+    function postCheck(ISuperfluid host, bytes memory params, address msgSender) external pure {
+        revert("I'm a bad macro");
```
hellwolf: I would make it based on "params" passed, and have a positive and negative test cases to show how postCheck should be used.

d10r: > I would make it based on "params" passed, and have a positive and negative test cases to show how postCheck should be used.

done, please check

hellwolf: okay.

Actually, on the second thought, a more "realistic" use case of  `postCheck` should be checking the "effect" of the `bool retVal = _forwardBatchCall(operations);`

Because checking params could have been done during `buildBatchOperations` too.

We can leave that example to docs or inline comments, instead. It's up to you how much you want to spend more time in this PR.

hellwolf (approved):

hellwolf (changes requested):
Excellent. @d10r, I think we add some changelog then we should be good.

hellwolf (approved):
Thanks!

PR #1929, opened 2024-04-19
Workflow caching with nix & faster pre-commit hook

### LINKED ISSUE #1854: [WORKFLOWS] Support Workflow Caching With Github Cache
Current Problems
==============

- Package testing workflows are always run regardless if there are input files change.
- Some operational workflows (deploy subgraph, deploy smart contracts) rebuild the essential packages all the time. It usually takes 5-10 minutes. It includes these steps:
  - Nix store setup, deterministic from (flake.nix, flake.lock)
  - Yarn node_module setup. deterministic from (nix store, yarn.lock)
  - Package source paths: ethereum-contracts js-sdk sdk-core sdk-redux

Solution Space
============

## Cached Workflow Result (Selected Solution)

With this solution, workflow results will be cached, and the cash key will be based on the input file set. This prevents the workflow from being run excessively for the same files.

**Technology Used**

- [Github Cache](https://github.com/actions/cache) provides a simple hash-key based cache scheme natively available to github users.
- [mk-cache-key.nix](https://github.com/hellwolf/mk-cache-key.nix/) -  a nix expression to make cache keys for "half board" nix projects. This works well in tandem with the GitHub cache.

## Incremental Build (For Comparison)

With this solution, new build results can be incrementally built on top of previous build results. This requires fundamentally more principled setup of build pipeline.

**Technology Candidates**

- [Nix](https://nix.dev/) & nixpkgs
- [Bazel](https://bazel.build/)

hellwolf:
- [x] First stage of build speed improvement:
  - [x] composite workflow .github/actions/setup-workflow-base/action.yml
    - using github cache
    - using mk-cache-key.nix
  - refactored workflows:
    - [x] spec-haskell
    - [x] solidity-semantic-money
    - [x] ethereum-contracts
    - [x] sdk-core
    - [x] hot-fuzz
    - [x] automation
    - [x] subgraph
- [x] fix a pre-commit hook bug that made it slow
  - new pre-commit hook
    - remove pre-commit script from all workflows,
    - instead, putting workflow filtering in the pre-commit hook.
  - more work needed to use lint-staged; not doing the refactoring in this PR.


hellwolf:
image

hellwolf:
image

Now it takes around 1 minutes to restore the cached workflow result.

hellwolf:
An example of cached workflow:

image

packages/sdk-core/default.nix
```diff
@@ -0,0 +1,29 @@
+{
+  halfBoardModule = {
+    dependencies = [
+      ../..
+      ../ethereum-contracts
+    ];
+    outputs = [
+      "dist"
+      "coverage"
```
kasparkallas, line 9: There might be more outputs? https://github.com/superfluid-finance/protocol-monorepo/blob/dev/packages/sdk-core/.gitignore

kasparkallas (approved):
Acknowledged.

packages/sdk-core/default.nix
hellwolf: right, will check again once the cache output is actually needed in some workflow.

PR #1933, opened 2024-04-23
[sdk-core][fix] lowecase subgraph argument when address

kasparkallas:
Why?
The behaviour used to be `query.id.toLowerCase()` (removed only in the dev-branch) without the address check but that caused issues with some Subgraph IDs that had both lowercase and uppercase letter in the ID. When the ID is an address, that's always lowercase though.


hellwolf (approved):

hellwolf:
suspicious ..

PR #1941, opened 2024-05-16
[SUBGRAPH] Add mainnet and sepolia

mmd-afegbua:


packages/metadata/networks.json
```diff
@@ -213,7 +213,7 @@
         "logsQueryRange": 20000,
         "explorer": "https://gnosisscan.io",
         "subgraphV1": {
-            "cliName": "gnosis",
+            "cliName": "xdai",
```
d10r: I think this is a regression - it used to be xdai, was changed to gnosis.
If goldsky still requires xdai, they probably use an old version. Changing here may cause breakage with other vendors.

d10r (changes requested):

packages/metadata/networks.json
kasparkallas: Yep, needs an exception for Goldsky in the subgraph deploy script

d10r (approved):

PR #1950, opened 2024-06-04
[SUBGRAPH] Subgraph Deploy CI tests

mmd-afegbua:


hellwolf:
@mmd-afegbua @kasparkallas conflicts with the changes that were merged to dev. Can you check if this PR still makes sense?

packages/subgraph/tasks/deploy.sh
```diff
@@ -132,7 +132,7 @@ deploy_to_goldsky() {
     )
 
     local goldskyNetwork="${legacyNetworkNames[$network]:-$network}"
-    local subgraphName="protocol-$DEPLOYMENT_ENV-$goldskyNetwork/$VERSION_LABEL"
+    local subgraphName="protocol-$DEPLOYMENT_ENV-$goldskyNetwork/1.0.0"
```
kasparkallas: This should be `$VERSION_LABEL`?

packages/sdk-core/previous-versions-testing/queryTests.ts
```diff
@@ -5,14 +5,14 @@ import metadata from "@superfluid-finance/metadata";
...
- * otherwise, we use OP Sepolia
+ * otherwise, we use avalanche fuji
  * @returns chainId
  */
 export const getChainId = () => {
     // null coalesce, but this should NEVER return null for either
     return process.env.SUBGRAPH_RELEASE_TAG == "v1"
         ? metadata.getNetworkByShortName("matic")?.chainId ?? 0
-        : metadata.getNetworkByShortName("opsepolia")?.chainId ?? 0;
+        : metadata.getNetworkByShortName("fuji")?.chainId ?? 0;
```
kasparkallas, line 15: Oh, I ran into this as well, was currently debugging, I was thinking that the issue is old metadata in the SDK-core, so I'm forcing it to a newer one with yarn's `resolutions` and that worked as well.

packages/subgraph/tasks/deploy.sh
mmd-afegbua: The version label from the script is the commit SHA. It changes for each deployment and it will lead to different subgraph  version on Goldsky. Being that the version is added to the subgraph url, we'll be forced to change the url for every new deployment if a static version (i.e 1.0.0) is not used.

kasparkallas: Goldsky supports tagging, that's what we should point everything at, not the concrete versions.

packages/sdk-core/previous-versions-testing/queryTests.ts
kasparkallas: It's fine to keep OP Sepolia here now, but it doesn't really matter which one we use.

hellwolf (approved):

PR #1954, opened 2024-06-11
[subgraph] configurable history pruning

kasparkallas:
## Why?
We want to index subgraphs to Goldsky with less entities created in the database.

## How?
To do so, we will indroduce history pruning. 

## Other
Updated subgraph tooling and spec.

## Warning!
If you rely on time-travel queries then this will be a breaking change. Contact us if you need continued access to time-travel queries!

Example of a time-travel query:
```graphql
{
  challenges(block: { number: 8000000 }) {
    challenger
    outcome
    application {
      id
    }
  }
}
```

Read more:
- https://thegraph.com/docs/en/cookbook/pruning/
- https://thegraph.com/docs/en/querying/graphql-api/#time-travel-queries

### DIFF
--- a/packages/subgraph/scripts/buildNetworkConfig.ts
+++ b/packages/subgraph/scripts/buildNetworkConfig.ts
@@ -20,7 +20,7 @@ interface SubgraphConfig {
 const ADDRESS_ZERO = "0x0000000000000000000000000000000000000000";
 
 const vendorCliNameExceptions: Record<string, Record<string, string>> = {
-    ["goldsky"]: {
+    "goldsky": {
         "xdai-mainnet": "xdai"
     }
 }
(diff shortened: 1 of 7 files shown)


packages/subgraph/scripts/buildNetworkConfig.ts
```diff
@@ -41,6 +52,7 @@ function main() {
         nativeAssetSuperTokenAddress: networkMetadata.nativeTokenWrapper,
         constantOutflowNFTAddress: networkMetadata.contractsV1.constantOutflowNFT || ADDRESS_ZERO,
         constantInflowNFTAddress: networkMetadata.contractsV1.constantInflowNFT || ADDRESS_ZERO,
+        indexerHints_prune: vendorHistoryPruning[vendorName] || "never",
```
kasparkallas, line 55: default to "never" (previous behavior)

d10r (approved):

PR #1955, opened 2024-06-11
[subgraph] lowest hanging fruit to reduce a bit of RPC calls

kasparkallas:
## Why?
To speed up indexing and to reduce indexing time.

## What?
Tried to massively reduce RPC calls in the following branch: https://github.com/superfluid-finance/protocol-monorepo/pull/1948 but it needs more testing. This is the lowest hanging fruit fix that struck out from that endeavour, which is to not do an RPC call for a balance if it has already been done in the same block. 

## How?
Add a simple condition to check whether the entity has been updated in the same block.

## Other
The fix might not have much effect if the `balanceOf` calls are somehow efficiently cached in the RPC, although it would still reduce I/O.

### DIFF
--- a/packages/subgraph/src/mappingHelpers.ts
+++ b/packages/subgraph/src/mappingHelpers.ts
@@ -986,22 +986,28 @@ function updateATSBalanceAndUpdatedAt(
                 balanceDelta as BigInt
             );
     } else {
-        // if the account has any subscriptions with units we assume that
-        // the balance data requires a RPC call for balance because we did not
-        // have claim events there and we do not count distributions
-        // for subscribers
-        const newBalanceResult = superTokenContract.try_realtimeBalanceOf(
-            Address.fromString(accountTokenSnapshot.account),
-            block.timestamp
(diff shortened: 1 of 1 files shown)


d10r (approved):

hellwolf:
test failing. is it related?

kasparkallas:
> test failing. is it related?

Yep, it's related. It's caused by the fact that when `AccountTokenSnapshot` is initialized then the `updatedAt` is immediately set.

Even though every time the entity is initialized, later on the balance function is called outside of it. There are 3 ways to solve it:
1. Call the balance function when initializing the entity.
2. Remove the `updatedAt` update from the initialization
3. Save a field in style of "balanceLastUpdatedFromRpcAtBlockNumber" onto the entity and use that to check

It's probably best to do 3 as it's the safest and is needed anyway if we are to add more RPC optimizations.

I'll not add this to the next deployment because of this, so I'll close this PR for now and continue with the other optimization PR.

PR #1959, opened 2024-06-18
[SUBGRAPH] Goldsky version label

mmd-afegbua:


.github/workflows/handler.deploy-production-subgraphs.yml
```diff
@@ -17,7 +17,6 @@ on:
         type: choice
         options:
           - superfluid
-          - graph
```
kasparkallas, line 20: Remove from descriptions as well. Look for "graph" and 'graph'

packages/subgraph/tasks/deploy.sh
```diff
@@ -126,7 +126,11 @@ deploy_to_goldsky() {
     local network="$1"
     # TODO: use tagging?
 
-    local subgraphName="protocol-$DEPLOYMENT_ENV-$network/1.0.0"
+    #Get subgraph version from package.json
+    PACKAGE_JSON_PATH="package.json"
+    VERSION_LABEL=$($JQ -r '.version' $PACKAGE_JSON_PATH)
```
kasparkallas: It's slightly confusing that`VERSION_LABEL` is an input but gets overriden here; but yeah, otherwise this is the version we want here.

mmd-afegbua: Changed it to SUBGRAPH_VERSION.

kasparkallas (approved):

PR #1965, opened 2024-06-21
[automations] VestingSchedulerV2 add helpful view functions

kasparkallas:
### Added and exposed `getCreateVestingScheduleParamsFromAmountAndDuration`

The view function can be used to know up front what will be the stored `VestingSchedule` values. Useful for UI-s. It's effectively the previous `_createVestingScheduleFromAmountAndDuration` exposed as a view function without any logic changed.

### Exposed `ScheduleCreationParams`

The struct holds all relevant values to a vesting schedule in addition to just the ones being stored. Useful for UI-s. Separated "superfluid context" out of it as it's a technical requirement of the smart contract to enable batching transaction, not necessarily relevant to the core data of the vesting schedule.

### Deleted `ScheduleCreationFromAmountAndDurationParams`

Less is more.

### Added and exposed `getMaximumNeededTokenAllowance`

Added a helpful utility function to calculate what is the maximum needed ERC-20 allowance for the vesting schedule to execute successfully under all conditions. Not used by the core contract logic as allowance is meant to be handled outside, so no added risk, but useful for UI-s (that otherwise would need to calculate all of this on their own).

### Used `uint96` for `remainderAmount`

Initially I used `uint256` for simplicity sake but `uint96` is plenty enough.

### Re-ordered `VestingSchedule` packing

As there's no difference in the storage size used, I opted to order it in a more backwards compatible way to V1 vesting scheduler.

### Added more tests

### DIFF
--- a/packages/automation-contracts/scheduler/contracts/VestingSchedulerV2.sol
+++ b/packages/automation-contracts/scheduler/contracts/VestingSchedulerV2.sol
@@ -704,6 +704,8 @@ contract VestingSchedulerV2 is IVestingSchedulerV2, SuperAppBase {
         return flowRate != 0;
     }
 
+
+    /// @dev IVestingScheduler.getCreateVestingScheduleParamsFromAmountAndDuration implementation.
     function getCreateVestingScheduleParamsFromAmountAndDuration(
         ISuperToken superToken,
         address receiver,
@@ -762,6 +764,7 @@ contract VestingSchedulerV2 is IVestingSchedulerV2, SuperAppBase {
         }
     }
 
+    /// @dev IVestingScheduler.getMaximumNeededTokenAllowance implementation.
     function getMaximumNeededTokenAllowance(
         address superToken,
         address sender,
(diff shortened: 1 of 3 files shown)


packages/automation-contracts/scheduler/test/VestingSchedulerV2.t.sol
```diff
@@ -1696,15 +1746,101 @@ contract VestingSchedulerV2Tests is FoundrySuperfluidTester {
...
+        uint256 totalAmount,
+        uint32 totalDuration,
+        uint32 cliffPeriod,
+        uint32 startDate,
+        uint8 randomizer
+    ) public {
+        // Assume
+        vm.assume(randomizer != 0);
+        
+        vm.assume(startDate == 0 || startDate >= block.timestamp);
```
hellwolf, line 1771: generally vm.bound is preferred over the  vm.assume. check their documentation on why.

packages/automation-contracts/scheduler/contracts/VestingSchedulerV2.sol
```diff
@@ -751,4 +654,112 @@ contract VestingSchedulerV2 is IVestingSchedulerV2, SuperAppBase {
         (,int96 flowRate,,) = cfaV1.cfa.getFlow(superToken, sender, receiver);
         return flowRate != 0;
     }
+
+
+    /// @dev IVestingScheduler.getCreateVestingScheduleParamsFromAmountAndDuration implementation.
+    function getCreateVestingScheduleParamsFromAmountAndDuration(
```
hellwolf, line 660: a minor thing, a trick to make this function "pure", is to take block.timestamp as a input value, e.g. "timestampNow"

0xPilou (approved):

hellwolf (approved):

PR #1966, opened 2024-06-24
[SUBGRAPHS] Add goldsky subgraph tag script

mmd-afegbua:
This is a script that will be run manually to promote a new version of subgraph to prod.

### DIFF
--- a/packages/subgraph/tasks/tagGoldskySubgraph.sh
+++ b/packages/subgraph/tasks/tagGoldskySubgraph.sh
@@ -3,9 +3,12 @@
 # Default subgraph type
 subgraph_type="protocol-v1"
 
+# Default tag
+tag="prod"
+
 # Function to display usage instructions
 usage() {
-    echo "Usage: $0 --token <API_KEY> --version <version> [--network <network>|all]"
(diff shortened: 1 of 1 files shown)


kasparkallas:
There's no GHA to trigger this yet?

packages/subgraph/tasks/tagGoldskySubgraph.sh
```diff
@@ -0,0 +1,72 @@
...
+        exit 1
+    fi
+}
+
+# Check if the network argument is 'all'
+if [ "$network" == "all" ]; then
+    goldsky_commands=$(generate_goldsky_commands)
+    run_goldsky_commands_in_docker "$goldsky_commands"
+elif [ -n "$network" ]; then
+    run_goldsky_commands_in_docker "if ! goldsky subgraph tag create $subgraph_type-$network/$version --tag prod; then
```
kasparkallas: Let's make the "tag" an input as well? Then we can do "dev" and other adhoc versions.

kasparkallas:
In general, when it comes to tagging, one thing to keep in mind is not to do the tagging before the subgraph has finished indexing, i.e. it's caught up to the edge as Goldsky people put it.

kasparkallas (approved):

hellwolf (approved):

PR #1979, opened 2024-07-02
Lib CallbackUtils & Remove PoolNFT callback try/catch

### LINKED ISSUE #1897: [ETHEREUM-CONTRACTS] PoolNFTs: remove try/catch
## What & Why

In #1879 we fixed an issue with the FlowNFTs: the hooks for minting and burning were wrapped in a try/catch and could fail with outofgas. This happened despite code supposed to avoid exactly that.
The Pool NFTs use the same mechanism. We haven't yet observed such internal reverts, but currently can't be confident the same could happen.

## AC

- [x] Remove try/catch from PoolNFT hooks & make sure there's no code path where it could revert

### LINKED ISSUE #1898: [ETHEREUM-CONTRACTS] Investigate 64/64 rule failure with try/catch
## What & Why

When working on #1879, 2 mitigations other than removing the try/catch were attempted

1. Change of order of instructions
```
uint256 gasLeftBefore = gasleft();
```
originally was followed by a call
```
address constantOutflowNFTAddress = _canCallNFTHook(flowVars.token);
```
and by an `if` statement.

So the assumption was that taking the snapshot as last instruction before the try would solve the problem.
But this turned out to not be the case, the test still failed.

2. try/catch may add overhead making it not suited for use with the 63/64 rule

Thus I locally tried to replace it with low-level calls, following the implementation done for SuperApp callbacks (see [here](https://github.com/superfluid-finance/protocol-monorepo/blob/dev/packages/ethereum-contracts/contracts/superfluid/Superfluid.sol#L1026)). To my surprise, it still failed. However I was focused on getting the task done (vs understanding what exactly was happening), which in case of the FlowNFTs could be achieved by simply not allowing reverts. So what I was seeing  could have been a red herring.

However we want to keep track of this and eventually figure out what exactly was the issue here, so we can avoid it causing troubles in other occasions.

## AC

- [ ] reproduce the issue described in #1879, find out why it happened and how to mitigate it (other than allowing no reverts at all)

hellwolf:
- CallbackUtils replaces SafeGasLibrary.
- Remove try/catch in PoolNFT callbacks.


d10r (approved):
Looks good.
I was most concerned about `poolMemberNFT.onDelete` failing if minting hadn't succeeded. But that's safe because if not minted, `poolMemberNFT.poolMemberDataByTokenId(tokenId).member` will be the zero address.
This way we get rid of the issue of try/catch misleading the gas estimation, leading to failed transactions.

This PR also fixes the failing test in https://github.com/superfluid-finance/protocol-monorepo/pull/1982 (I tried locally).

PR #2003, opened 2024-08-08
[WORKFLOW] Subgraph deploy matrix

### LINKED ISSUE #1856: [SUBGRAPH WORKFLOWS] Use matrix for multi-vendor|network deployment
## What

If deploying to multiple networks and/or vendors via GH workflow, a [matrix](https://docs.github.com/en/actions/using-jobs/using-a-matrix-for-your-jobs) (ideally 2-dimensional) is used to map each deployment (vendor & network) to a job

## AC

* When running the workflow to deploy a subgraph, with network set to "all", the actual deploy step for each network is run as its own job
* no manually maintained redundant list of networks per vendor which could get out of sync
* it's still possible to deploy to all networks for a given vendor without using a GH workflow

## How

The workflow https://github.com/superfluid-finance/protocol-monorepo/blob/dev/.github/workflows/handler.deploy-production-subgraphs.yml currently just forwards the job of iterating through "all" networks to https://github.com/superfluid-finance/protocol-monorepo/blob/dev/packages/subgraph/tasks/deploy.sh.
That script contains then figures out what "all" means for the given vendor and iterates through them.
We instead want to dynamically create a matrix of networks to be used if "all" networks are selected, and map the deployment step to a job for each. The dynamic part is needed in order to avoid having to maintain redundant mappings in the subgraph package and in a workflow file.

mmd-afegbua:
Sample run: https://github.com/superfluid-finance/protocol-monorepo/actions/runs/10580677816


d10r (changes requested):
This moves logic from scripts (deploy.sh) to workflow files, specifically the list of networks per vendor.
That violates the principle that workflows shall not contain more logic than necessary, such that they don't become the only way to execute tasks.

mmd-afegbua:
> This moves logic from scripts (deploy.sh) to workflow files, specifically the list of networks per vendor. That violates the principle that workflows shall not contain more logic than necessary, such that they don't become the only way to execute tasks.

The logic introduced is the matrix for the list of network for each vendor, which had been previously handled by `deploy.sh`. `deploy.sh` can still execute the deployment task, just that if it is being used outside of the workflow, one is expected to know the vendor and networks that are supported there. This way, you dont't have to make changes to the deploy script if for instance, you want to experiment with a new network.

hellwolf:
@d10r

d10r:
We didn't manage to fix the [failing test](https://github.com/superfluid-finance/protocol-monorepo/actions/runs/10577003087/job/29304019724#step:13:353) here, thus it's not required anymore (after consulting with @kasparkallas ).

Could be improved: when deploying to multiple networks, run only the deploy step for each.

d10r (approved):

PR #2005, opened 2024-08-19
Small build pipeline improvements

hellwolf:


kasparkallas (approved):

d10r:
~~please don't yet merge, I want to add something~~
nvm

d10r (approved):
I don't know what motivates the change, hope it's worth it.
```
didi@P1G4:~/src/sf/protocol-monorepo$ time jq --version
jq-1.7.1

real    0m0.002s
user    0m0.001s
sys     0m0.000s
didi@P1G4:~/src/sf/protocol-monorepo$ time npx --no --package=node-jq -- jq --version
jq-1.7.1

real    0m2.892s
user    0m3.890s
sys     0m1.226s
```

hellwolf:
> I don't know what motivates the change, hope it's worth it.

The origin of having to add this $JQ is because ubuntu docker provided by github at some point removed jq from the pre-installs, and at that time the build system has not been utilizing nix yet.

The npx performance seems unnecessarily poor, indeed, though it's not the primary goal of the build system; the primary goal should still be reproducibility.

However, since we now already use nix, I might as well revert back and use `jq` directly assuming either it's run in nixified build system or locally that you have jq installed (nix or non-nix way).

PR #2008, opened 2024-08-24
[WORKFLOW] Fix the counting of files

glitched-w0rld:
```
find "pr-packages/$pr_number" -maxdepth 1 -type f -printf '.' | wc -c
```

The above is an overly complicated way to count files. This PR replaces it with a simpler method using `ls`.

```
ls -1 "pr-packages/$pr_number" | wc -l
```


hellwolf:
Thanks for the report, now I remember why I changed it to using `find`:

image

hellwolf:
Let me read into shellcheck and see if there is any merit of that recommendation

hellwolf:
https://github.com/koalaman/shellcheck/issues/2950  I will follow up this one and make an upstream contribution.

I will close this one for now.

Thanks!

PR #2017, opened 2024-09-06
[SUBGRAPH] fix missing token name/symbol

### LINKED ISSUE #1992: [SUBGRAPH] Native Super Token name empty
Example: ETHx on base-mainnet at [0x46fd5cfb4c12d87acd3a13e92baa53240c661d93](https://basescan.org/token/0x46fd5cfb4c12d87acd3a13e92baa53240c661d93#readProxyContract).

The subgraph returns an empty string for the name:
```
query MyQuery {
  tokens(where: {isNativeAssetSuperToken: true}) {
    id
    name
    symbol
  }
}
```

```
{
  "data": {
    "tokens": [
      {
        "id": "0x46fd5cfb4c12d87acd3a13e92baa53240c661d93",
        "name": "",
        "symbol": "ETHx"
      }
    ]
  }
}
```

The onchain name is set:
image

Affected chains:
eth-mainnet, base-mainnet, scroll-mainnet, optimism-mainnet, arbitrum-one, avalanche-c, bsc-mainnet

Not affected:
polygon-mainnet, xdai-mainnet

So there seems to be something different with the OG networks which triggers this behaviour.

d10r:
fixes #1992

### DIFF
--- a/packages/subgraph/src/mappingHelpers.ts
+++ b/packages/subgraph/src/mappingHelpers.ts
@@ -1,4 +1,4 @@
-import { Address, BigInt, ethereum } from "@graphprotocol/graph-ts";
+import { Address, BigInt, ethereum, log } from "@graphprotocol/graph-ts";
 import { FlowUpdated } from "../generated/ConstantFlowAgreementV1/IConstantFlowAgreementV1";
 import { ISuperfluid as Superfluid } from "../generated/Host/ISuperfluid";
 import {
@@ -43,7 +43,8 @@ import {
     getStreamID,
     getStreamRevisionID,
     getSubscriptionID,
-    handleTokenRPCCalls
+    handleTokenRPCCalls,
+    getTokenInfoAndReturn,
 } from "./utils";
 
 /**************************************************************************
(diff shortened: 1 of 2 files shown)


kasparkallas (approved):

d10r:
**what happened:**
Custom Super Tokens tend to register via factory _before_ initializing themselves via `SuperToken.initialize()`, this is the case for the native token wrapper too (see [deploy script](https://github.com/superfluid-finance/protocol-monorepo/blob/dev/packages/ethereum-contracts/ops-scripts/deploy-super-token.js#L101)). As a result, the first invocation of `getOrInitSuperToken()` queries a token proxy contract which points to the canonical SuperToken logic, but isn't yet initialized.  
Initialization typically happens in another transaction closely following.

The previous code handled later updating of the token symbol, but that's not the only field set on initialize, see [SuperToken._initialize()](https://github.com/superfluid-finance/protocol-monorepo/blob/dev/packages/ethereum-contracts/contracts/superfluid/SuperToken.sol#L230).  
The logic now is: if symbol and name have zero length (proxy condition for detecting an uninitialized token), sync all the entity data from RPC again. This should cover all special cases of this category, even those we don't yet see - e.g. a custom SuperToken wrapper.

PR #2023, opened 2024-09-23
[ETHEREUM-CONTRACTS] more realistic fUSDC

### LINKED ISSUE #1654: [PROTOCOL-OPS] verification script failures go unnoticed
[https://github.com/superfluid-finance/protocol-monorepo/blob/dev/packages/ethereum-contracts/tasks/etherscan-verify-framework.sh](etherscan-verify-framework.sh) contains logic which prints failed verification at the end.  
This seems to have stopped working at some point - invocations of the truffle verify command don't (always) return an error code anymore when a verification fails. May be related to the addition of sourcify as verification target.

AC:
* investigate and document what exactly is happening
* implement and/or recommend a mitigation - may involve the [blackbox verification check](https://github.com/superfluid-finance/protocol-monorepo/blob/dev/tasks/daily-slack-bot.js#L162)
  * if no mitigation is implemented, remove the now misleading message about "failed verification" and add a reminder telling that not all verifications may have succeeded

Optional:
Use the opportunity to add some inline commends and/or script outputs explaining that proxies are often based on older version's bytecode (can be different even if just compiler version changed).  
Re-verifying then usually succeeds (just printing "already verified"), but there seem to be exceptions to that, e.g. SuperfluidGovernanceIIProxy has been observed causing failures even if already verified. Example:
```
+ npx truffle run --network base-mainnet verify SuperToken@0x46fd5cfB4c12D87acD3a13e92BAa53240C661D93 --custom-proxy SETHProxy
Verifying contracts on basescan
   Verifying SuperToken@0x46fd5cfB4c12D87acD3a13e92BAa53240C661D93
   Verifying custom proxy contract SETHProxy at 0x46fd5cfB4c12D87acD3a13e92BAa53240C661D93
   Verifying proxy implementation SuperToken at 0xfe5209ac341835670fa731383d5d28c8a093dda3
   Invalid constructor arguments provided. Please verify that they are in ABI-encoded format
   Failed to verify 1 contract(s): SuperToken@0x46fd5cfB4c12D87acD3a13e92BAa53240C661D93
Verifying contracts on sourcify
   Verifying SuperToken@0x46fd5cfB4c12D87acD3a13e92BAa53240C661D93
   Verifying custom proxy contract SETHProxy at 0x46fd5cfB4c12D87acD3a13e92BAa53240C661D93
   Verifying proxy implementation SuperToken at 0xfe5209ac341835670fa731383d5d28c8a093dda3
   Fail - Unable to verify: The deployed and recompiled bytecode don't match.
   Failed to verify 1 contract(s): SuperToken@0x46fd5cfB4c12D87acD3a13e92BAa53240C661D93
```
This despite both [proxy and impl being verified](https://basescan.org/address/0x46fd5cfB4c12D87acD3a13e92BAa53240C661D93#code).

### LINKED ISSUE #1853: [ETHEREUM-CONTRACTS DEVX] Shift fUSDC to 6 decimals
Problem: USDC has 6 decimals while fUSDC has 18.
Applications using fUSDC with the expectation that it behaves like the mainnet counterpart may get burned by this difference.

Solution: change fUSDC to have 6 decimals:
1. change for new deployments: adjust deployment script
2. change for existing deployments: requires a one-time "hack"
  a.  add logic to SuperToken which allows the owner to change
  b.  update fUSDC to this logic
  c.  trigger the logic to make the switch to 6 digits
  d.  update to canonical logic

To be considered: effect of that change on existing fUSDCx and upgrading/downgrading.
I currently can't think of any particular side effects. But didn't think about it deeply.

d10r:
* new testnet deployments: fUSDC with 6 decimals
* permit support

### DIFF
--- a/packages/ethereum-contracts/CHANGELOG.md
+++ b/packages/ethereum-contracts/CHANGELOG.md
@@ -9,6 +9,7 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 
 * `IUserDefinedMacro`: added a method `postCheck()` which allows to verify state changes after running the macro.
 * `SuperfluidFrameworkDeployer` now also deploys and `MacroForwarder` and enables it as trusted forwarder.
+* `deploy-test-environment.js` now deploys fUSDC (the underlying) with 6 decimals (instead of 18) to better resemble the actual USDC.
 
 ### Fixed
 
--- a/packages/ethereum-contracts/ops-scripts/deploy-test-environment.js
+++ b/packages/ethereum-contracts/ops-scripts/deploy-test-environment.js
@@ -57,26 +57,32 @@ module.exports = eval(`(${S.toString()})()`)(async function (
     let tokens;
 
     if (args.length >= 1) {
-        tokens = args.pop().split(",");
+        const tokenSymbols = args.pop().split(",");
+        tokens = tokenSymbols.map((symbol) => ({ symbol, decimals: 18}));
     } else {
-        tokens = ["fDAI", "fUSDC", "fTUSD", config.nativeTokenSymbol];
+        tokens = [
+            { symbol: "fDAI", decimals: 18 },
+            { symbol: "fUSDC", decimals: 6 },
+            { symbol: "fTUSD", decimals: 18 },
+            { symbol: "ETH", decimals: 18 }
+        ];
     }
-    console.log("Tokens to be deployed", tokens);
+    console.log("Super Tokens to be deployed", tokens.map((t) => t.symbol + "x"));
 
     console.log("======== Deploying superfluid framework ========");
     await deployFramework(errorHandler, options);
     console.log("==== Superfluid framework deployed  ========");
 
     for (let i = 0; i < tokens.length; ++i) {
-        if (tokens[i] !== deploySuperToken) { // ???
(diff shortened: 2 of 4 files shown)


d10r:
permit support was already added in https://github.com/superfluid-finance/protocol-monorepo/pull/1664, is available on newer testnets

d10r:
piggy-backed fix for non-silent verification failure (https://github.com/superfluid-finance/protocol-monorepo/issues/1654)

d10r:
Note regarding permit: the latest [version 2.2](https://github.com/circlefin/stablecoin-evm/blob/master/contracts/v2/FiatTokenV2_2.sol) of `FiatToken` (USDC) added a variant of `permit` where the signature is just one parameter instead of 3. This has become common, but is non-standard (not as specified in https://eips.ethereum.org/EIPS/eip-2612).

hellwolf (commented):
I don't see the permit support?
    Missing update of the changelog.

d10r:
> * I don't see the permit support?
> 
>     * Missing update of the changelog.

permit was already there, see comments.
no changelog update because only internal tooling changed.

hellwolf:
>  no changelog update because only internal tooling changed.

It is a change to the deploy-test-network flow, so adding a line in changelog wouldn't hurt for future reference.

hellwolf (approved):
excellent, keep squashing issues!

PR #2035, opened 2024-11-05
[DEMO] how we could replace specific token operations with the generic 2771 call proxy

d10r:
Needed changes:
* Make SuperToken an `IRelayRecipient`, trusting the DMZForwarder
* Replace occurances of `msg.sender` with `_msgSender()` (which uses the 2771-encoded sender if present)

Here only `SuperToken.increaseAllowance()` was changed that way, with a new test case using it.

In order to do this, we'd need to
* use `_msgSender()` in all SuperToken methods
* mark the operationX methods as deprecated
* update SuperToken contracts

As long as there's SuperTokens not updated to this logic, we can't easily remove the then legacy operations types from batch call.  
However we could remove the code from SuperToken if we conditionally re-map legacy operations in `Superfluid.sol` to use 2771Forwarder if the target token contract supports it.


d10r:
note: `BaseRelayRecipient._getTransactionSigner()` does not do what it's natspec claims.

PR #2037, opened 2024-11-13
[ETHEREUM-CONTRACTS] split DMZForwarder into SimpleForwarder and ERC2771Forwarder

d10r:
The recently introduced DMZForwarder needs to be split into 2 contracts, otherwise it cannot be safely trusted as ERC-2771 forwarder.

After the change, we'll have
- `SimpleForwarder` for `OPERATION_TYPE_SIMPLE_FORWARD_CALL`. Is deployed by the constructor of Superfluid - new deployment on every contract upgrade, which is ok.
- `ERC2771Forwarder` for `OPERATION_TYPE_ERC2771_FORWARD_CALL`. Is deployed by the deployment script and set as immutable.

`ISuperfluid` now has a method `getERC2771Forwarder()`. This is to be used by target contracts of `OPERATION_TYPE_ERC2771_FORWARD_CALL`, they can use:
```
  function isTrustedForwarder(address forwarder) public view override returns(bool) {
      return forwarder == address(_host.getERC2771Forwarder());
  }
```

The only reason why `SIMPLE_FORWARDER` is public is to facilitate contract verification.

As a side effect of increasing the Superfluid contract size by a few bytes, the size of `SuperfluidHostDeployerLibrary` is now above the 24576 byte allowed by https://eips.ethereum.org/EIPS/eip-170.
Since this contract is just test harness and [hardhat node can now be configured to allow "oversized" contracts](https://hardhat.org/hardhat-network/docs/reference#allowunlimitedcontractsize), we can lift this requirement on test contracts.
(All the shenanigans of `SuperfluidFrameworkDeployer` and related complexity fallout thus become obsolete and could be undone going forward)

The PR also contains a fix for the `codeChanged` function which is supposed to tell us if a contract needs to be upgraded.
The previous version would look at the wrong portion of bytecode for contracts with nested contracts (typical case: a contract containing code for deploying a proxy).
A side effect of the fix is that upgrading will now be triggered much more often, even when there's no actual contract changes. That's because the metadata portion of the bytecode changes for the slightest change, even if it's just a change in a comment in an imported interface file.
The fixing of this function also required adjustment of some test code where immutable replacements were missing.


packages/ethereum-contracts/contracts/interfaces/superfluid/Definitions.sol
```diff
@@ -231,7 +231,7 @@ library BatchOperation {
      */
     uint32 constant internal OPERATION_TYPE_SUPERFLUID_CALL_APP_ACTION = 2 + 200;
     /**
-     * @dev DMZForwarder.forwardCall batch operation type
+     * @dev SimpleForwarder.forwardCall batch operation type
      *
      * Call spec:
```
hellwolf, line 236: consider to add some security notes too.

packages/ethereum-contracts/CHANGELOG.md
```diff
@@ -3,6 +3,11 @@ All notable changes to the ethereum-contracts will be documented in this file.
 
 This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [v1.11.2]
+
+### Changed
+* Added `getERC2771Forwarder()` to `ISuperfluid`, to be used by batch call targets who want to use ERC-2771 for msg sender preservation.
```
hellwolf: Two utility contracts are added.

packages/sdk-core/hardhat.config.ts
```diff
@@ -49,6 +49,7 @@ const config: HardhatUserConfig = {
     networks: {
         hardhat: {
             chainId: 31337,
+            allowUnlimitedContractSize: true,
```
hellwolf, line 52: this workround is an indication that we have broken the usability of SuperfluidFrameworkDeploymentSteps.sol in hardhat based projects.

packages/ethereum-contracts/package.json
```diff
@@ -74,7 +74,7 @@
         "build:post-contracts": "run-p -l build:post-contracts:*",
         "build:post-contracts:abi-bundle": "tasks/build-bundled-abi.sh",
         "build:post-contracts:dev-scripts-typings": "rm -rf dev-scripts/*.d.ts dev-scripts/*.d.ts.map; tsc -p tsconfig.scripts.json",
-        "build:post-contracts:contracts-size": "forge build --sizes > build/contracts-sizes.txt",
+        "build:post-contracts:contracts-size": "forge build && forge build --sizes > build/contracts-sizes.txt || true",
```
hellwolf: we should remove this workaround.

packages/sdk-core/hardhat.config.ts
d10r: proposal: https://github.com/d10r/sf-hardhat-example/

d10r: PS: the one SF + HH user we know about already has `allowUnlimitedContractSize` set: https://github.com/GoodDollar/GoodProtocol/blob/master/hardhat.config.ts#L105

packages/ethereum-contracts/package.json
d10r: I changed this task such that it ignores contracts used only in (ephemeral) test context.
That way it will still fail if any other contract gets "oversized".

d10r:
After some investigation, I recommend we stop enforcing the contract size limit for contracts only used in testing context.
Because:
- there's a simple workaround for keeping hardhat supported with the `allowUnlimitedContractSize` switch
- in case the compiler warnings aren't enough, it's easy to set a hardhat project up such that it will throw an error if a contract is too large, see https://github.com/d10r/sf-hardhat-example/blob/master/hardhat.config.js#L4
- this decision was made at a time when (afaik) the proposed workaround didn't exist, or we weren't aware of it
- hardhat isn't the default toolkit in the ecosystem anymore, so adding a simple requirement (one line config change) for compatibility shouldn't be a deal breaker

Note: for just this specific instance, it would be quite easy to work around the issue once more. But we are piling up extra complexity for imo not good enough reasons. Changing that would allow us not only to not make this PR more complicated, but also open the door for considerable simplification of the codebase in a future PR.

hellwolf (approved):

PR #2040, opened 2024-12-05
[ETHEREUM-CONTRACTS] [WIP] cheaper balance checks & more

d10r:
## cheaper balance checks

When doing a transfer or starting a stream, we check if the sender has enough balance.
Instead of calculating the exact balance, it's enough to check if there's enough balance.
That can often be done much more cheaply, that is when a sender account has a lot of GDA pools connected.
Since the connected pools can only _increase_ the balance, we can stop iterating as soon as the required balance threshold is met.

The current implementation violates the agreement abstraction because it needs to leave the GDA balance last.
This is made with the IDA deprecation already in mind (for this optimization to work well, we can have only 1 agreement with non-constant balance calculation cost).

This status is to be considered a PoC in order to test and show the idea, more attention should be paid to naming, avoidance of code reduncancy, testing, etc.

## SuperTokenV1Library

- added `flowWithCtx` and `flowFromWithCtx`

### DIFF
--- a/packages/ethereum-contracts/CHANGELOG.md
+++ b/packages/ethereum-contracts/CHANGELOG.md
@@ -7,7 +7,9 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 
 ### Added
 
-- `SuperTokenV1Library`: overloaded `claimAll` for the msg.sender to claim for themselves.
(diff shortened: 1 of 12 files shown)


d10r:
PR did its job of prototyping & conveying the idea

PR #2042, opened 2024-12-13
[sdk-core] new Operation types and ETH value forwarding

kasparkallas:
~~TODO: bump version & changelog~~

```
### Added
- Handle new Operation types with BatchCall
- Forward ETH value with BatchCall and Operation
```

### DIFF
--- a/packages/sdk-core/src/BatchCall.ts
+++ b/packages/sdk-core/src/BatchCall.ts
@@ -1,5 +1,5 @@
 import { JsonFragment } from "@ethersproject/abi";
-import { ethers } from "ethers";
+import { BigNumber, ethers } from "ethers";
 
 import Host from "./Host";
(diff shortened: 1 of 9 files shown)


hellwolf (approved):

packages/sdk-core/src/BatchCall.ts
```diff
@@ -89,10 +92,25 @@ export default class BatchCall {
...
         );
-        const tx =
-            this.host.contract.populateTransaction.batchCall(
-                operationStructArray
-            );
+
+        const values = operationStructArray
+            .filter((x) => x.value?.gt(BigNumber.from(0)))
+            .map((x) => x.value);
+
```
kasparkallas, line 99: A possible situation to protect against would be that the operation with the ETH value would be the first payable call in the set of operations.

d10r (approved):

PR #2043, opened 2024-12-16
[ETHEREUM-CONTRACTS] remove footgun from CFASuperAppBase

d10r:
## What & Why

The CFA hook `afterAgreementTerminated` is invoked when a flow to the SuperApp or a flow from the SuperApp is deleted.
The latter case is unexpected for many, because it exists only for the delete callback, not for the create and update callbacks.
For SuperApps creating outgoing flows and implementing this hook without being aware of it (e.g. not checking the sender argument), this creates the risk of unexpected states, unexpected state updates, and of jailing.

## How

- invoke `onFlowDeleted` only for deletion of incoming flows and remove the `receiver` argument. This breaks existing code using it, which we want in order to make sure the change can't cause unnoticed behaviour change of such contracts.
- invoke a new hook `onOutFlowDeleted` for deletion of outgoing flows. This allows to specifically handle this case, e.g. to implement _sticky_ streams (don't allow closing by reopening in the hook).

Note: the change is _breaking_ in the sense that code using it and updating the dependency needs to be changed.
Being just a library which is not used by the protocol itself, it's not breaking in the sense that deployed contracts change behaviour.


packages/ethereum-contracts/contracts/apps/CFASuperAppBase.sol
```diff
@@ -146,6 +158,24 @@ abstract contract CFASuperAppBase is ISuperApp {
     function onFlowDeleted(
```
hellwolf: Since it's a breaking change, changing it to `onInFlowDeleted` will make it less ambiguous and less reliant to checking comments before assuming its meaning.

hellwolf (approved):

hellwolf:
@d10r can you check if this can address https://github.com/superfluid-org/protocol-monorepo/issues/2069 ?

d10r:
> @d10r can you check if this can address #2069 ?

can do. But that makes the change considerably more breaking. Worth it?

hellwolf:
> > @d10r can you check if this can address #2069 ?
> 
> can do. But that makes the change considerably more breaking. Worth it?

It seems so.

PR #2044, opened 2024-12-17
[ETHEREUM-CONRACTS] fix simple forwarder

d10r:
Problem: SimpleForwarder is deployed by the constructor of Superfluid.
`SimpleForwarder.forwardCall` was made `onlyOwner` for no good reason, other than mitigating potential unspecified phishing type abuse. However this causes it not to work in deployments where Superfluid (host) is upgradable, which is all prod deployments.
The tests didn't "see" this issue because the foundry deployer doesn't use an upgradable host instance.

Solution: remove the `onlyOwner` clause.
Also `withdrawLostNativeTokens` was replaced by logic which makes it impossible for any native tokens to remain in SimpleForwarder. They either are taken by the target, or taken back by the sender, or the tx fails. TODO: add test coverage verifying this.

### DIFF
--- a/.github/workflows/handler.publish-pr-packages.yml
+++ b/.github/workflows/handler.publish-pr-packages.yml
@@ -39,7 +39,7 @@ jobs:
       - name: Download artifact
         continue-on-error: true
         id: download_artifact
-        uses: dawidd6/action-download-artifact@v4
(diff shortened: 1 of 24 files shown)


hellwolf (changes requested):
to be replaced by fix-simple-forwarder-2

PR #2045, opened 2024-12-20
[ETHEREUM-CONTRACTS] fix simple forwarder (alternative solution)

d10r:
https://github.com/superfluid-finance/protocol-monorepo/pull/2044 fixes the issue by removing the `onlyOwner` clause from SimpleForwarder.
This PR fixes it by giving up on the idea of having the contract deployed in the constructor of Superfluid, allowing the SimpleForwarder contract to remain unchanged, at the price of adding another constructor argument to Superfluid.


hellwolf (approved):

hellwolf (commented):
wait.

changelog?

PR #2048, opened 2024-12-31
[ETHEREUM-CONTRACTS] update SuperTokenV1Library

d10r:
- Add `createPoolWithCustomERC20Metadata`
- SuperTokenV1Library `distribute` / `distributeFlow` return `actualAmount` / `actualFlowRate`

### DIFF
--- a/packages/ethereum-contracts/CHANGELOG.md
+++ b/packages/ethereum-contracts/CHANGELOG.md
@@ -14,9 +14,11 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 
 ### Changed
 - Fixed deployment of SimpleForwarder (solved an issue which caused batch operation `OPERATION_TYPE_SIMPLE_FORWARD_CALL` to always revert)
-- `SuperTokenV1Library.distributeFlow`: return `actualFlowRate`
-- `SuperTokenV1Library.distribute`: return `actualAmount`
 - `SuperTokenV1Library.getFlowRate` and `SuperTokenV1Library.getFlowInfo` now also allow querying the flowrate between pools and pool members.
+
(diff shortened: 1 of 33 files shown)


packages/ethereum-contracts/CHANGELOG.md
```diff
@@ -7,10 +7,15 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
...
 - Superfluid Pools now implement `IERC20Metadata`, thus going forward have a name, symbol and decimals
-- `ISuperfluidPool.createPoolWithCustomERC20Metadata` for creating pools with custom ERC20 metadata
+- `IGeneralDistributionAgreementV1.createPoolWithCustomERC20Metadata` and `SuperTokenV1Library.createPoolWithCustomERC20Metadata` for creating pools with custom ERC20 metadata
+- `SuperTokenV1Library`
+  - overloaded `claimAll` for the msg.sender to claim for themselves
+  - added `flowWithCtx` and `flowFromWithCtx`
 
 ### Changed
 - Fixed deployment of SimpleForwarder (solved an issue which caused batch operation `OPERATION_TYPE_SIMPLE_FORWARD_CALL` to always revert)
+- `SuperTokenV1Library.distributeFlow`: return `actualFlowRate`
```
hellwolf: it could be breaking change for some; I think I might have a few places doing assert(whateverreturningbool)

hellwolf: but that's an okay breaking change, since it's not at ABI level, and code will not compile for the new function signature if they relying on returning bool.

hellwolf (approved):
consider to update CHANGELOG to mark some of the changes BREAKING

PR #2051, opened 2025-01-14
[ETHEREUM-CONTRACTS] emit Transfer event in updateMemberUnits()

d10r:
Problem: now that the pools have proper ERC20 metadata, we also want tooling like explorers to properly render the token state.
Since explorers heavily rely on the Transfer event as trigger for balance updates, it becomes a problem that `updateMemberUnits` doesn't emit them. This PR changes that.

`updateMemberUnits` to a higher than previous amount is effectively a mint, thus a transfer from zero to the member.
`updateMemberUnits` to a lower than previous amount is effectively a burn, thus a transfer from the member to zero.

Already deployed to base-sepolia for test purposes.


d10r:
the "fix" is working for etherscan: https://sepolia.basescan.org/token/0xbff63599fdf8b5225c98ee8767c1f177104ae8ec

hellwolf (approved):

PR #2052, opened 2025-02-18
switch to foundry v1 (stable)

d10r:


packages/automation-contracts/autowrap/foundry.toml
```diff
@@ -3,6 +3,8 @@ root = '../../../'
 libs = ['lib']
 src = 'packages/automation-contracts/autowrap'
 solc_version = "0.8.23"
```
hellwolf, line 5: shall we also mark evm_version explicitly?

packages/automation-contracts/scheduler/foundry.toml
```diff
@@ -3,6 +3,8 @@ root = '../../../'
 libs = ['lib']
 src = 'packages/automation-contracts/scheduler'
 solc_version = "0.8.23"
+optimizer = true
+optimizer_runs = 200
```
hellwolf, line 8: ditto.

hellwolf (approved):

PR #2077, opened 2025-06-16
[ETHEREUM-CONTRACTS] v1.13.0: SuperfluidPool: add [increase|decrease]MemberUnits and deprecate PoolMemberNFT

d10r:
This PR contains:
* https://github.com/superfluid-org/protocol-monorepo/issues/2080
* https://github.com/superfluid-org/protocol-monorepo/issues/2079

Additionally there's a small optimization: `GDA._isPool()` is replicated in `SuperfluidPool.sol`.
This shaves off a few thousand gas for `[update|increase|decrease]MemberUnit`, because it removes an indirection through the GDA which adds 2 external calls (proxy and logic contract).

The version is bumped to v1.13.0 because disabling the PoolMemberNFT update hooks is technically a breaking changes.

This PR doesn't yet remove all the PoolMemberNFT related code because for the deprecation its logic contract is updated one last time. After that, more of the related code can be removed.

v1.13.0 also contains the already merged https://github.com/superfluid-org/protocol-monorepo/pull/2057


packages/ethereum-contracts/contracts/agreements/gdav1/SuperfluidPool.sol
```diff
@@ -474,8 +454,19 @@ contract SuperfluidPool is ISuperfluidPool, BeaconProxiable {
             assert(GDA.appendIndexUpdateByPool(superToken, p, t));
         }
         emit MemberUnitsUpdated(superToken, memberAddr, oldUnits, newUnits);
+    }
 
-        _handlePoolMemberNFT(memberAddr, newUnits);
+    // replicates GDAv1._isPool in order to eliminate unnecessary gas cost (less external calls)
```
hellwolf: It's a low hanging fruit of not violating DRY: Extract the storage related function as a global "free" function so that reusing is strictly inlining same code in two different contracts (gda logic or the pool logic).

hellwolf (changes requested):

packages/ethereum-contracts/contracts/agreements/gdav1/SuperfluidPool.sol
hellwolf: if you want faster 1.13 progress, removing this particular change is also okay for me. But I ask for no code quality degradation by not repeating code.

d10r: Is it ok like this?
https://github.com/superfluid-org/protocol-monorepo/pull/2077/commits/35e54ee15fb4607c6633dbfc37eb7602d20c44fe

hellwolf:
I will propose a small improvement in the next commit.

hellwolf (approved):
I will continue the storage layout code refactoring in https://github.com/superfluid-org/protocol-monorepo/pull/2086

PR #2083, opened 2025-06-24
[ETHEREUM-CONTRACTS] Dedicated allowlist contract for SuperApp registration

d10r:
## What & Why

This PR introduces a dedicated simple allowlist contract for managing SuperApp registration permission.
This is done in order to make this rather frequent process more lightweight, as it does not require the same degree of security of other governance actions like contract upgrades.

## How

Adds an immutable constructor argument to `Superfluid.sol` for the allowlist contract.
Extends the precondition for SuperApp registration to first check if permission is given by this allowlist. If not, it also looks up the governance contract, such that pre-existing permissions stay in place.
This solution does not require any changes to the governance contract and only small and very localized changes to the host contract.

## Other

The PR also includes some replacements of `registerAppWithKey` and `registerAppByFactory` with `registerApp` in mock contracts. This is just cleanup of deprecated logic, not directly related to the changes.


packages/ethereum-contracts/contracts/utils/AllowList.sol
```diff
@@ -0,0 +1,20 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.23;
+
+import { Ownable } from "@openzeppelin/contracts/access/Ownable.sol";
```
hellwolf: Can also make a simple usage of OZ's roles contracts. Regardless of how simple it is, not having to rebuild the wheel should be the starting point.

packages/ethereum-contracts/contracts/interfaces/superfluid/ISuperfluid.sol
```diff
@@ -658,6 +658,12 @@ interface ISuperfluid {
     // solhint-disable func-name-mixedcase
     function getERC2771Forwarder() external view returns(address);
 
+    /**
+     * @dev returns the address of the ACL contract used for granular permissioning.
+     * @return address of the ACL contract
+     */
+    function getACL() external view returns(address);
```
hellwolf: Unclear naming: ACL of what. Calling `getSuperAppRegistrationACL` was fine.  

`function getSuperAppRegistrationACL() external view returns (IAccessControl)`

This would makes it even more clear that it has a `IAccessControl` interface :

```
Functions
[hasRole(role, account)](https://docs.openzeppelin.com/contracts/4.x/api/access#IAccessControl-hasRole-bytes32-address-)

[getRoleAdmin(role)](https://docs.openzeppelin.com/contracts/4.x/api/access#IAccessControl-getRoleAdmin-bytes32-)

[grantRole(role, account)](https://docs.openzeppelin.com/contracts/4.x/api/access#IAccessControl-grantRole-bytes32-address-)

[revokeRole(role, account)](https://docs.openzeppelin.com/contracts/4.x/api/access#IAccessControl-revokeRole-bytes32-address-)

[renounceRole(role, account)](https://docs.openzeppelin.com/contracts/4.x/api/access#IAccessControl-renounceRole-bytes32-address-)
````

File: packages/ethereum-contracts/contracts/utils/ACL.sol [hunk 3]
```diff
@@ -0,0 +1,16 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.23;
+
+import { AccessControl } from "@openzeppelin/contracts/access/AccessControl.sol";
+
+contract ACL is AccessControl {
```
hellwolf: Please always try to give it a distinctive name, it doesn't deserve such a undecorated name since it has specific implementation of how admin role are assigned and transferred. 

Any thing with an adjective is better: "TransferrableAdminACL", "SimpleACL", etc.

packages/ethereum-contracts/contracts/interfaces/superfluid/ISuperfluid.sol
d10r: more verbose documentation added

hellwolf (approved):
old registration method is no longer tested though.

PR #2084, opened 2025-06-26
[ETHEREUM-CONTRACTS] distinguish between unsettled and claimable balance in pools

d10r:
## What & Why

`getClaimable` should return the amount which will be added to the balance if calling `claimAll`. This is currently not the case, because the implementation mixes up the concepts of _claimable amount_ and _unsettled amount_.

The point of the claiming concept is to allow pool members to _fetch_ tokens allocated to them even when they are not connected to that pool. If the pool is connected, there's nothing to be claimed - the connected status could conceptually also be seen as being continuous auto-claiming.

If `getClaimable` returns a non-zero number, a call of `claimAll` is expected to add that amount to the user balance.
With the current implementation that is true only for unconnected pools.

## How

The inconsistency is solved by making `getClaimable` always return 0 for connected pools.
This is achieved by having distinct logic for internal settlement and for claiming.


hellwolf:
what prompted this issue? I prefer this be done after #2086 refactoring.

d10r:
> what prompted this issue? I prefer this be done after #2086 refactoring.

Surprising behaviour when working on tests in the context of autoconnect (not yet pushed). And yes, should be after #2086 then.

hellwolf:
UnsettledValue shouldn't be "leaked" as such.

I will reject this one for now.

I will continue from #2086, where we can touch this topic, too.

d10r:
this is now based on https://github.com/superfluid-org/protocol-monorepo/pull/2086

d10r:
> UnsettledValue shouldn't be "leaked" as such.

Note that it's not part of the interface. The PR doesn't change the public interface, it just fixes the behaviour of `getClaimable` and `getClaimableNow`. `getUnsettledValue` is conceptually private, but has public visibility because the GDA needs it too.

packages/ethereum-contracts/contracts/agreements/gdav1/SuperfluidPool.sol
```diff
@@ -459,10 +456,23 @@ contract SuperfluidPool is ISuperfluidPool, BeaconProxiable {
         emit MemberUnitsUpdated(superToken, memberAddr, oldUnits, newUnits);
     }
 
-    function _claimAll(address memberAddr, uint32 time) internal returns (int256 amount) {
-        amount = getClaimable(memberAddr, time);
+    function _settle(address memberAddr, int256 amount) internal {
```
hellwolf: I will have to reject this impelemtation, for design reason:

This function is effectful and has no  constraint on what "amount" can be provided to it. Because of this, now as a code reader, you will have to worry if all the call sites of this function provided this "amount" correctly. The more robust design is to avoid such possibility, keep reasoning of correctness local.

In this case, it is quite easy to do, I will do a change in the following commmit.

hellwolf (approved):
checkout the updated _settle, then we can merge.

PR #2093, opened 2025-07-17
[ETHEREUM-CONTRACTS] allow connecting pools on behalf of members (with some constraints)

d10r:
## What & Why

Onboarding new users to SuperTokens via GDA distributions is tainted if they have to do a claim transaction before seeing Super Tokens being sent to them.
This can be mitigated by allowing to connect pools on behalf of members.
At the same time this power shall be constraint in order to keep the "spam protection" aspect - 3rd parties shall not be able to _use up_ all slots, and they shall also not be able to increase gas costs for that account due to more expensive balance calculation. Thus autoconnect won't be allowed to use all slots, and a mechanism for accounts to fully opt out will be added.

In short:
* 3rd party can connect pool members using a subset of the slots
* member can disable autoconnect

## How

Add a function to the GDA which allows to connect pools on behalf of members.
Add a mechanism which allows members to opt out of autoconnect.

This can also be viewed as a kind of ACL: by default permission is given, but it can be withdrawn.

Implementation detail:
The pre-existing `connectPool` will revert if all slots are occupied.
The newly added `tryConnectPoolFor` does not revert, but returns a bool signalling if the attempt succeeded.

Why not using try/catch: because the gas estimation algorithm of RPC nodes uses binary search to find the smallest gas limit for which the transaction still succeeds, not caring about failing nested calls in try/catch.

## TBD
* [x] implement a mechanism which lets members disable autoconnect
* [ ] (maybe) ~~implement a mechanism for a subset of contracts to opt-out of autoconnect by default~~
* [x] change evm version to "shanghai" for all packages
* [x] add support for new method to SuperTokenV1Library


packages/ethereum-contracts/contracts/agreements/gdav1/GeneralDistributionAgreementV1.sol
```diff
@@ -305,49 +311,100 @@ contract GeneralDistributionAgreementV1 is AgreementBase, TokenMonad, IGeneralDi
...
+
+        bool isConnected = token.isPoolMemberConnected(this, pool, memberAddr);
+
+        if (doConnect != isConnected) {
             if (doConnect) {
-                uint32 poolSlotId =
-                    _findAndFillPoolConnectionsBitmap(token, msgSender, bytes32(uint256(uint160(address(pool)))));
+                if (onlyAutoConnectSlots) {
+                    // check if we're below the slot limit for autoconnect
+                    (uint32[] memory slotIds, ) = SlotsBitmapLibrary.listData(
```
hellwolf: I will consider a new function for the library that only counts the number of used slots.

packages/ethereum-contracts/test/foundry/agreements/gdav1/GeneralDistributionAgreement.t.sol
```diff
@@ -694,7 +695,7 @@ contract GeneralDistributionAgreementV1IntegrationTest is FoundrySuperfluidTeste
         assertEq(poolAdjustmentFlowRate, 0, "GDAv1.t: Pool adjustment rate is non-zero");
     }
 
-    function testDistributeFlowToUnconnectedMembers(
+    function skip_testDistributeFlowToUnconnectedMembers(
```
hellwolf: justify the "skip_"?

packages/solidity-semantic-money/foundry.toml
```diff
@@ -4,7 +4,7 @@ src = 'packages/solidity-semantic-money/src'
 out = 'packages/solidity-semantic-money/out/default'
 cache_path = 'packages/solidity-semantic-money/out/default.cache'
 solc_version = '0.8.26'
-evm_version = 'paris' # no PUSH0 for now
+evm_version = 'paris'
```
hellwolf: I will check in a separate PR.

packages/ethereum-contracts/test/foundry/agreements/gdav1/GeneralDistributionAgreement.t.sol
d10r: undone, was overlooked

packages/ethereum-contracts/contracts/agreements/gdav1/GeneralDistributionAgreementV1.sol
```diff
@@ -305,49 +311,100 @@ contract GeneralDistributionAgreementV1 is AgreementBase, TokenMonad, IGeneralDi
...
+                SuperfluidPool(address(pool)).operatorConnectMember(
+                    memberAddr, doConnect, uint32(currentContext.timestamp)
+                )
+            );
 
-    /// @inheritdoc IGeneralDistributionAgreementV1
-    function connectPool(ISuperfluidPool pool, bytes calldata ctx) external override returns (bytes memory newCtx) {
-        return connectPool(pool, true, ctx);
-    }
+            emit PoolConnectionUpdated(token, pool, memberAddr, doConnect, currentContext.userData);
```
hellwolf, line 404: Without adding anything, is there an easy way to know if the connection was an "auto-connect" from event logs?

packages/ethereum-contracts/contracts/agreements/gdav1/GeneralDistributionAgreementV1.sol
```diff
@@ -305,49 +311,100 @@ contract GeneralDistributionAgreementV1 is AgreementBase, TokenMonad, IGeneralDi
...
+        if (doConnect != isConnected) {
             if (doConnect) {
-                uint32 poolSlotId =
-                    _findAndFillPoolConnectionsBitmap(token, msgSender, bytes32(uint256(uint160(address(pool)))));
+                if (onlyAutoConnectSlots) {
+                    // check if we're below the slot limit for autoconnect
+                    (uint32[] memory slotIds, ) = SlotsBitmapLibrary.listData(
+                        token, memberAddr, _POOL_SUBS_BITMAP_STATE_SLOT_ID, _POOL_CONNECTIONS_DATA_STATE_SLOT_ID_START
+                    );
+                    if (slotIds.length >= MAX_POOL_AUTO_CONNECT_SLOTS) {
```
hellwolf: This should be `(nUsedSlots > MAX_POOL_AUTO_CONNECT_SLOTS)`, will be fixed with the `SlotsBitmapLibrary.countUsedSlots` in nex commit.

hellwolf: I did mess up with the logic, will revert.

packages/ethereum-contracts/contracts/agreements/gdav1/GeneralDistributionAgreementV1.sol
```diff
@@ -305,49 +311,102 @@ contract GeneralDistributionAgreementV1 is AgreementBase, TokenMonad, IGeneralDi
...
-        public
-        returns (bytes memory newCtx)
+    /// @inheritdoc IGeneralDistributionAgreementV1
+    function connectPool(ISuperfluidPool pool, bytes calldata ctx) external override returns (bytes memory newCtx) {
+        newCtx = ctx;
+        _setPoolConnection(pool, address(0), true /* doConnect */, ctx);
+    }
+
+    /// @inheritdoc IGeneralDistributionAgreementV1
+    function tryConnectPoolFor(ISuperfluidPool pool, address memberAddr, bytes calldata ctx)
```
hellwolf, line 315: Note that you can call `tryConnectPoolFor` with `address(0)`, it will act the same as calling `connectPool()`

Test might be confused now, if it doesn't ignore such inputs.

packages/ethereum-contracts/contracts/interfaces/agreements/gdav1/IGeneralDistributionAgreementV1.sol
```diff
@@ -214,6 +214,22 @@ abstract contract IGeneralDistributionAgreementV1 is ISuperAgreement {
...
+    /// @param memberAddr The member address
+    /// @param ctx Context bytes
+    /// @return success true if the member was (or remained) connected, false otherwise
+    /// @return newCtx the new context bytes
+    function tryConnectPoolFor(ISuperfluidPool pool, address memberAddr, bytes calldata ctx) 
+        external
+        virtual
+        returns (bool success, bytes memory newCtx);
+
+    /// @notice Lets accounts deny or allow 3rd parties to connect them to pools. The default is to allow.
```
hellwolf: typo

packages/ethereum-contracts/contracts/agreements/gdav1/GeneralDistributionAgreementV1.sol
```diff
@@ -305,49 +311,102 @@ contract GeneralDistributionAgreementV1 is AgreementBase, TokenMonad, IGeneralDi
...
+            success = false;
+        } else {
+            success = _setPoolConnection(pool, memberAddr, true /* doConnect */, ctx);
+        }
+    }
+
+    function setConnectPermission(bool allow) external override {
+        IAccessControl simpleACL = ISuperfluid(_host).getSimpleACL();
+        if (!allow) {
+            simpleACL.grantRole(ACL_POOL_CONNECT_EXCLUSIVE_ROLE, msg.sender);
```
hellwolf, line 338: who is the admins of simpleACL??

d10r: It's still the GH agent, but after deploying this it should be handed over to SF gov or SF gov owner.

packages/ethereum-contracts/contracts/agreements/gdav1/GeneralDistributionAgreementV1.sol
d10r: Right. But it already has `vm.assume(member != address(0));`, so should be good.

packages/ethereum-contracts/contracts/agreements/gdav1/GeneralDistributionAgreementV1.sol
d10r: I don't think there is.

hellwolf:
Also, auto-connect pools for pools should not be a valid operation, adding an error type and revert is needed.

AFAIR, in practice, due to how realtimeBalanceOf is implemented for pools, this doesn't create issues. However, the new API should still prevent it from happening.

lib/forge-std
hellwolf, line 1: We should make a separate PR for this.

packages/ethereum-contracts/contracts/agreements/gdav1/GeneralDistributionAgreementV1.sol
```diff
@@ -305,49 +311,102 @@ contract GeneralDistributionAgreementV1 is AgreementBase, TokenMonad, IGeneralDi
...
     }
 
-    // @note setPoolConnection function naming
-    function connectPool(ISuperfluidPool pool, bool doConnect, bytes calldata ctx)
-        public
-        returns (bytes memory newCtx)
+    /// @inheritdoc IGeneralDistributionAgreementV1
+    function connectPool(ISuperfluidPool pool, bytes calldata ctx) external override returns (bytes memory newCtx) {
+        newCtx = ctx;
+        _setPoolConnection(pool, address(0), true /* doConnect */, ctx);
```
hellwolf: Consider to piggyback a small change `_setPoolConnection` to `_setPoolConnectionFor`, the code might read more fluidly gramatically.

packages/ethereum-contracts/contracts/agreements/gdav1/GeneralDistributionAgreementV1.sol
hellwolf: This could be an issue. An option is to supplement with a new event type.

lib/forge-std
d10r: oh, did I commit this?
Apparently it broke only one line, which I fixed in https://github.com/superfluid-org/protocol-monorepo/pull/2093/commits/c5fd4d513d02763b60ca3e828c75c9fba64d080e. Should it be undone anyway?

packages/ethereum-contracts/contracts/agreements/gdav1/GeneralDistributionAgreementV1.sol
```diff
@@ -238,44 +244,43 @@ contract GeneralDistributionAgreementV1 is AgreementBase, TokenMonad, IGeneralDi
...
         returns (ISuperfluidPool pool)
     {
-        return _createPool(
-            token,
-            admin,
-            config,
-            PoolERC20Metadata("", "", 0) // use defaults specified by the implementation contract
-        );
+        string calldata emptyStr;
+        assembly {
```
hellwolf: This may not be safe when createPool is called NOT by an EOA, please don't use it, since there is no guarantee the calldata passed is the end of the ALL calldata. I can construct a bigger calldata and call the createPool with it.

Generally, this is too much of a hack to worth the risk.

d10r: Afaik nothing could go wrong in a dangerous way bcs reading data of length 0.
But we can do without here, so I reverted it.

hellwolf: Reading length itself is 1 WORD.

packages/ethereum-contracts/ops-scripts/deploy-framework.js
```diff
@@ -947,6 +947,31 @@ module.exports = eval(`(${S.toString()})({skipArgv: true})`)(async function (
         if (gdaNewLogicAddress !== ZERO_ADDRESS) {
             agreementsToUpdate.push(gdaNewLogicAddress);
         }
+
+        // check/set ACL role admins
+        const simpleAcl = await SimpleACL.at(simpleAclAddress);
+
+        const aclSuperappRegistrationRoleAdmin = web3.utils.sha3("ACL_SUPERAPP_REGISTRATION_ROLE_ADMIN");
```
hellwolf, line 954: any comment worth having?

hellwolf (approved):

hellwolf:
Suggest for a new PR: Make SimpleACL upgrade robust, by checking if it gets accidentally overriden durign the upgrade process.

majidrazaee2:
دوستان من اشتباه کردم چون کسانی که خودرا همکار شماستند به من گفتن اینکار را انجام دهم

majidrazaee2:
My friend lost 10 wallets to fraud. It's been 4 years since I paid.

majidrazaee2:
I had 3 accounts that I worked with: majidrazaee2@gmail.com razaeem13@gmail.com www.majidrazaee@gmail.com

majidrazaee2:
I lost 10 bags in 4 years.

majidrazaee2:
I lost 10 Trust Wallets in 4 years.

majidrazaee2:
لطفا اگر نمی‌خواهید من با شما ادامه دهم به من بگویید

majidrazaee2:
Thank you to good friends.

PR #2097, opened 2025-08-01
[ETHEREUM-CONTRACTS] Migrate to OpenZeppelin v5

### LINKED ISSUE #2016: [ETHEREUM-CONTRACTS] Upgrade to OpenZeppelin 5 and Rework Its Dependency Management (Breaking Chnage)
# Synopsis

OpenZeppelin was [released](https://blog.openzeppelin.com/introducing-openzeppelin-contracts-5.0) on October 2023. It is a major upgrade, aiming to "modernizing its contracts." Inevitably it contains a few breaking changes, hence making the upgrade to it in the monorepo non-trivial.

Additionally, OpenZeppelin dependency has been managed as part of node dependency in monorepo, while other ethereum contracts dependencies are managed by git submodule directly. It would be a good opportunity to remove this nuisance when upgrading the major version of OpenZeppelin.

Also, monorepo is often distributed to developers as a "git submodule". To use the monorepo, developers usually must "remap" "@openzeppelin/contracts/" to the OpenZeppelin the developer uses. This may introduce a conflict between v4 and v5 OpenZeppelin, although this has a workaround that is noted in the "technical notes" section.

# Acceptance Criteria

This issue aims to upgrade OpenZeppelin to v5 in monorepo and provides a developer smooth breaking-change upgrade while allowing v4/v5 co-existance in developers' repo.

# Technical Notes

**To Remap both v4/v5**

remapping.txt

```
@openzeppelin/contracts/=lib/openzeppelin-contracts-v4/contracts/
@openzeppelin-v5/contracts/=lib/openzeppelin-contracts-v5/contracts/
```

d10r:
## What & Why

see #2016 

## How

Imports and notes about what changed:
```
utils/introspection/IERC1820Registry.sol
  path changed to interfaces/

utils/introspection/ERC1820Implementer
  was removed. Used only in ERC777SenderRecipientMock.t.sol, so I put a copy there.

utils/math/SafeMath.sol
  nomore needed in Solidity 0.8, its functions can be replaced by operators

utils/math/SafeCast.sol

utils/Address.sol (SuperToken.sol)
  isContract() was removed, thus it was replaced by its previous implementation `account.code.length > 0 ?`
  The OZ5 library added a custom error, which caused the ABI to not be empty anymore, which caused typechain to add it, which caused a conflict with the alias in `mappedSubgraphTypes.ts`. Thus added a workaround in `build-types.sh` which removes the entries generated for the Address library.

utils/cryptography/ECDSA.sol

access/IAccessControl.sol

access/AccessControlEnumerable.sol
  path changed to access/extensions

access/AccessControl.sol
  `_setupRole()` was removed. Can be trivially replaced by `_grantRole()`

access/Ownable.sol
  constructor now requires an argument, initial owner needs to be explicitly set (was implicitly set to `_msgSender` before)

token/ERC20/extensions/IERC20Metadata.sol

token/ERC20/extensions/IERC20Permit.sol

token/ERC721/extensions/IERC721Metadata.sol

token/ERC777/IERC777.sol
  path changed to interfaces

token/ERC777/IERC777Recipient.sol
    path changed to interfaces

token/ERC777/IERC777Sender.sol
    path changed to interfaces

token/ERC20/ERC20.sol
  note: SuperToken does NOT inherit this, but did replicate some of the functionality, incl. revert strings. Remaining ones could be replaced by custom errors as specified in [EIP-6093](https://eips.ethereum.org/EIPS/eip-6093), as done in the OZ5 impl.

token/ERC20/utils/SafeERC20.sol (library)
  `safeApprove` has been removed (was already deprecated in v4) because there's nothing safe about it. Was used in `SuperUpgrader`, now replaced with `forceApprove` (which is specifically about ERC20 contracts which requires approval to be set to 0 before it can be set to a new value).

proxy/Proxy.sol
  receive() was removed. It was thus explicitly added to `UUPSProxy` and `FullUpgradableSuperTokenProxy` in order to keep their functionality as is.

proxy/utils/Initializable.sol
  STORAGE LAYOUT CHANGED!
  This is used in `UUPSProxiable` and `BeaconProxiable`. We need to retain the storage layout of those, thus we keep using the v4 version.

proxy/beacon/UpgradeableBeacon.sol

proxy/beacon/BeaconProxy.sol
  v5 added an immutable which stores the beacon address - additionally to storing it in the eip-1967-defined storage slot.
  This is a clever optimization allowing cheaper lookup at runtime. It's safe because the storage layout isn't affected and the beacon address is immutable thus can't get out of sync with the pointer saved in storage.

interfaces/IERC5267.sol
```

openzeppelin-contracts is now a git submodule (foundry style) instead of an npm package.
In order to stay compatible with hardhat (used by legacy tests) and truffle (used by ops scripts), an npm postinstall script places a symlink to that submodule in node_modules. This was found to be the only solution which doesn't cause significant issues with either hardhat or truffle (see dedicated comment).

The openzeppelin imports in the ethereum-contracts package are now qualified by the major version, that is `@openzeppelin-v5/...`. This allows contracts using SF contracts as a dependency to use the default namespace `@openzeppelin` with the OZ version they prefer. The OZ package is now declared a peerDependency.
Hardhat projects need to set up a remapping task in their config file, see CHANGELOG.


packages/ethereum-contracts/contracts/utils/SuperUpgrader.sol
```diff
@@ -60,8 +60,7 @@ contract SuperUpgrader is AccessControlEnumerable {
         IERC20 token = IERC20(superToken.getUnderlyingToken());
         uint256 beforeBalance = token.balanceOf(address(this));
         token.safeTransferFrom(account, address(this), amount);
-        token.safeApprove(address(superToken), 0);
-        token.safeApprove(address(superToken), amount);
+        token.forceApprove(address(superToken), amount);
```
d10r, line 63: This sets to 0 first only if needed (if failing otherwise).
Always first setting to 0 is generally not useful. That makes sense only in the context of a Dapp which wants to apply a delta by calculating the new approval amount non-atomically (outside the applying tx context), risking frontrunning. Indeed `safeApprove` was removed in OZ5 because it was essentially a placebo.
Exception: supposedly there's ERC20 contracts out there not allowing to set allowance from a non-zero to another non-zero value. `forceApprove` takes care of that.

packages/sdk-core/package.json
```diff
@@ -55,7 +55,7 @@
         "pretest": "yarn testenv:start",
         "test": "hardhat test --tsconfig \"tsconfig.test.json\"",
         "dev": "nodemon -e ts -x yarn test",
-        "clean": "rm -rf node_modules; rm -rf dist; rm -rf src/typechain-types; rm -rf src/typechain; rm -rf src/abi; find . -type f -name '*.generated.ts' -exec rm {} +",
+        "clean": "rm -rf node_modules; rm -rf dist; rm -rf typechain-types; rm -rf src/typechain; rm -rf src/abi; find . -type f -name '*.generated.ts' -exec rm {} +",
```
d10r, line 58: this had the wrong path and thus didn't properly clean up

packages/sdk-core/tasks/build-types.sh
```diff
@@ -15,5 +15,11 @@ fi
 # copy the typechain files over from ethereum-contracts
 cp -r ../ethereum-contracts/typechain-types ./src/typechain-types
 
+# Remove the Address export from typechain-types to avoid conflict with mappedSubgraphTypes
+# OpenZeppelin v5 added a custom error to Address library, giving it a non-empty ABI
+# This causes typechain to generate types for it, but it's not needed for the SDK
+sed -i '/export type { Address } from ".\/@openzeppelin\/contracts\/utils\/Address";/d' ./src/typechain-types/index.ts
+sed -i '/export { Address__factory } from ".\/factories\/@openzeppelin\/contracts\/utils\/Address__factory";/d' ./src/typechain-types/index.ts
+
```
d10r, line 23: This became an issue with OZ5 because the Address library now contains a custom error, thus the ABI is nomore empty, thus the hardhat typescript plugin creates a type of it, colliding with the alias to string we have defined in `packages/sdk-core/src/subgraph/mappedSubgraphTypes.ts`.

packages/ethereum-contracts/contracts/upgradability/Initializable.sol
```diff
@@ -0,0 +1,166 @@
...
+     *
+     * Similar to `reinitializer(1)`, except that functions marked with `initializer` can be nested in the context of a
+     * constructor.
+     *
+     * Emits an {Initialized} event.
+     */
+    modifier initializer() {
+        bool isTopLevelCall = !_initializing;
+        require(
+            (isTopLevelCall && _initialized < 1) || (address(this).code.length == 0 && _initialized == 1),
```
hellwolf, line 90: make a comment about this code diff from the original OZ' contract

d10r:
**Making OZ a git submodule**

option 1:
adding in lib/openzeppelin-contracts
foundry: change mappings in all foundry.toml
hardhat.config.js: add a subtask for the remapping:
```
// Remapping for OpenZeppelin contracts
subtask(TASK_COMPILE_GET_REMAPPINGS).setAction(
    async (_, __, runSuper) => {
        const remappings = await runSuper();
        return {
            ...remappings,
            "@openzeppelin/contracts/": "lib/openzeppelin-contracts/contracts/",
        };
    }
);
```
and make the monorepo root the hardhat project root (as is the case in the foundry project)
```
paths: {
        root: "../../",
        sources: "packages/ethereum-contracts/contracts",
        artifacts: "packages/ethereum-contracts/build/hardhat",
        tests: "packages/ethereum-contracts/test",
        cache: "packages/ethereum-contracts/cache",
}
```
Without that, the remapping won't work.
truffle: found no way to get it compiling with OZ outside of the root directory.

option 2:
Possible solution: move lib/ into packages/ethereum-contracts, add 
```
                remappings: [
                    "@openzeppelin/contracts/=lib/openzeppelin-contracts/contracts/",
                ],
```
to truffle-config.js.
Problem: truffle wont create artifact files for imports via remapping. E.g. Ownable and AccessControl are missing. Some truffle script need those. A possible workaround is to place a link to openzeppelin-contracts in node_modules and add create a dummy contract which imports Ownable etc. via that path. But that breaks the hardhat setup (throws an error if it sees a contract references via 2 different paths).

option 3:
as long as we have truffle, place a symlink in node_modules

packages/ethereum-contracts/contracts/upgradability/Initializable.sol
d10r: done

d10r:
We could change EVM version to `cancun`, all SF networks now support it (see https://github.com/hellwolf/eip-canaries/pull/1).
However ganache doesn't support it, so I reverted to `shanghai`.

hellwolf:
@d10r, please cherry-pick https://github.com/superfluid-org/protocol-monorepo/pull/2099

packages/automation-contracts/autowrap/foundry.toml
```diff
@@ -9,7 +9,8 @@ optimizer_runs = 200
 remappings = [
     '@superfluid-finance/solidity-semantic-money/src/=packages/solidity-semantic-money/src/',
     '@superfluid-finance/ethereum-contracts/=packages/ethereum-contracts/',
-    '@openzeppelin/=node_modules/@openzeppelin/',
+    '@openzeppelin-v5/=lib/openzeppelin-contracts/',
```
hellwolf, line 12: ~~for automation contracts, why is openzeppelin-v5 referenced, if it is not used in the code?~~

```
nix-shell$ git grep openzeppelin-v5 packages/automation-contracts/
packages/automation-contracts/autowrap/foundry.toml:    '@openzeppelin-v5/=lib/openzeppelin-contracts/',
packages/automation-contracts/autowrap/hardhat.config.js:            "@openzeppelin/contracts/": "@openzeppelin-v5/contracts/",
packages/automation-contracts/scheduler/foundry.toml:    '@openzeppelin-v5/=lib/openzeppelin-contracts/',
packages/automation-contracts/scheduler/hardhat.config.js:            "@openzeppelin/contracts/": "@openzeppelin-v5/contracts/",
```

Actually, I realized both are v5. So, it is a hack to minimize code changes, I guess. I am not sure it's worth keeping such tech debt anymore. Maybe let's just change the code, it's trivial enough.

packages/automation-contracts/scheduler/contracts/VestingSchedulerV3.sol
```diff
@@ -532,7 +531,7 @@ contract VestingSchedulerV3 is IVestingSchedulerV3, IRelayRecipient {
                 });
             } else {
                 // Linear Default Cliff (calculated based on the overall vesting flow rate)
-                cliffAmount = SafeMath.mul(cliffPeriod, SafeCast.toUint256(flowRate));
+                cliffAmount = cliffPeriod * SafeCast.toUint256(flowRate);
```
hellwolf, line 534: This is a doggy change that needs more explanation.

packages/automation-contracts/scheduler/contracts/VestingSchedulerV2.sol
```diff
@@ -221,10 +220,7 @@ contract VestingSchedulerV2 is IVestingSchedulerV2, SuperAppBase {
                 remainderAmount: remainderAmount
             });
         } else {
-            uint256 cliffAmount = SafeMath.mul(
-                cliffPeriod,
-                SafeCast.toUint256(flowRate)
-            );
+            uint256 cliffAmount = cliffPeriod * SafeCast.toUint256(flowRate);
```
hellwolf, line 223: This is a doggy change that needs more explanation.

packages/automation-contracts/scheduler/test/VestingSchedulerV2.t.sol
```diff
@@ -899,7 +898,7 @@ contract VestingSchedulerV2Tests is FoundrySuperfluidTester {
         );
 
         console.log("Revert with overflow.");
-        vm.expectRevert("SafeCast: value doesn't fit in 96 bits");
+        vm.expectRevert(); // SafeCastOverflowedIntDowncast
```
hellwolf, line 901: there is no reference to it?

packages/automation-contracts/scheduler/test/VestingSchedulerV3.t.sol
```diff
@@ -1105,7 +1105,7 @@ contract VestingSchedulerV3Tests is FoundrySuperfluidTester {
         );
 
         console.log("Revert with overflow.");
-        vm.expectRevert("SafeCast: value doesn't fit in 96 bits");
+        vm.expectRevert(); // SafeCastOverflowedIntDowncast
```
hellwolf, line 1108: ditto

packages/ethereum-contracts/contracts/agreements/gdav1/GeneralDistributionAgreementV1.sol
```diff
@@ -319,7 +319,7 @@ contract GeneralDistributionAgreementV1 is AgreementBase, TokenMonad, IGeneralDi
     {
         newCtx = ctx;
 
-        if (pool.superToken().isPool(this, memberAddr)) {
+        if (memberAddr == address(0) || pool.superToken().isPool(this, memberAddr)) {
```
hellwolf, line 322: sneaky.

packages/ethereum-contracts/contracts/upgradability/SuperfluidUpgradeableBeacon.sol
```diff
@@ -3,15 +3,15 @@ pragma solidity ^0.8.23;
...
+} from "@openzeppelin-v5/contracts/proxy/beacon/UpgradeableBeacon.sol";
 import { BeaconProxiable } from "./BeaconProxiable.sol";
 
 contract SuperfluidUpgradeableBeacon is UpgradeableBeacon {
     error ZERO_ADDRESS_IMPLEMENTATION();    // 0x80883162
     error INCOMPATIBLE_LOGIC();             // 0x5af2144c
     error NO_PROXY_LOOP();                  // 0z66750bca
 
-    constructor(address implementation_) UpgradeableBeacon(implementation_) {}
+    constructor(address implementation_) UpgradeableBeacon(implementation_, _msgSender()) {}
```
hellwolf, line 14: Note:

* Context._msgSender
* Ownable is Context

packages/ethereum-contracts/ops-scripts/deploy-framework.js
```diff
@@ -859,11 +840,9 @@ module.exports = eval(`(${S.toString()})({skipArgv: true})`)(async function (
             "ERC2771_FORWARDER"
         );
 
-        const simpleAclAddress = await getOrDeployHelper(
-            SimpleACL,
-            () => superfluid.getSimpleACL(),
-            "SIMPLE_ACL"
-        );
+        // SimpleACL has now been deployed on all networks.
```
hellwolf, line 843: Okay. What about deploying to a new network, does the bootstrap logic still exist?

packages/ethereum-contracts/test/foundry/agreements/gdav1/GeneralDistributionAgreement.t.sol
```diff
@@ -1093,6 +1093,16 @@ contract GeneralDistributionAgreementV1IntegrationTest is FoundrySuperfluidTeste
             new bytes(0)
         );
         vm.stopPrank();
+
+        // cannot connect the zero address
+        vm.startPrank(alice);
```
hellwolf, line 1098: okay

packages/solidity-semantic-money/foundry.toml
```diff
@@ -11,7 +11,7 @@ optimizer_runs = 200
 via_ir = false
 remappings = [
     '@superfluid-finance/solidity-semantic-money/src/=packages/solidity-semantic-money/src/',
-    '@openzeppelin/=node_modules/@openzeppelin/',
+    '@openzeppelin/=lib/openzeppelin-contracts/',
```
hellwolf, line 14: I will change this to @openzeppelin-v5

packages/ethereum-contracts/contracts/superfluid/SuperToken.sol
```diff
@@ -362,10 +358,9 @@ contract SuperToken is
         _move(operator, holder, recipient, amount, "", "");
 
         if (spender != holder) {
-            _approve(
-                holder,
-                spender,
-                _allowances[holder][spender].sub(amount, "SuperToken: transfer amount exceeds allowance"));
+            require(amount <= _allowances[holder][spender], "SuperToken: transfer amount exceeds allowance");
+            // TODO: this triggers an `Approval` event, which shouldn't happen for transfers.
```
hellwolf, line 362: I can't recall when did this creep in.

packages/automation-contracts/autowrap/hardhat.config.js
```diff
@@ -4,10 +4,22 @@ require("@nomiclabs/hardhat-etherscan");
 require("hardhat-deploy");
 require("hardhat/config");
 require("./script/addStrategy");
+const {TASK_COMPILE_GET_REMAPPINGS} = require("hardhat/builtin-tasks/task-names");
 
 // You need to export an object to set up your config
 // Go to https://hardhat.org/config/ to learn more
 
+// Remapping for OpenZeppelin contracts
+subtask(TASK_COMPILE_GET_REMAPPINGS).setAction(
```
hellwolf, line 12: consider to drop a section in the https://github.com/superfluid-org/protocol-monorepo/blob/dev/packages/ethereum-contracts/README.md

packages/automation-contracts/scheduler/foundry.toml
```diff
@@ -9,7 +9,8 @@ optimizer_runs = 200
 remappings = [
     '@superfluid-finance/solidity-semantic-money/src/=packages/solidity-semantic-money/src/',
     '@superfluid-finance/ethereum-contracts/=packages/ethereum-contracts/',
-    '@openzeppelin/=node_modules/@openzeppelin/',
+    '@openzeppelin-v5/=lib/openzeppelin-contracts/',
```
hellwolf, line 12: why having both versions?

packages/ethereum-contracts/contracts/mocks/ERC777SenderRecipientMock.t.sol
```diff
@@ -1,19 +1,55 @@
...
+ * @dev Implementation of the {IERC1820Implementer} interface.
+ *
+ * Contracts may inherit from this and call {_registerInterfaceForAddress} to
+ * declare their willingness to be implementers.
+ * {IERC1820Registry-setInterfaceImplementer} should then be called for the
+ * registration to be complete.
+ *
+ * CAUTION: This file is deprecated as of v4.9 and will be removed in the next major release.
+ */
+contract ERC1820Implementer is IERC1820Implementer {
```
hellwolf: Consider to move it under contracts/utils/

packages/ethereum-contracts/contracts/superfluid/SuperToken.sol
hellwolf: Added a draft to https://github.com/orgs/superfluid-org/projects/1/views/1 for later to address.

hellwolf (approved):

PR #2104, opened 2025-08-28
[AUTOMATION] deploy autowrap with burned owner keys, switch to hardhat-verify

d10r:
- Autowrap: call `renounceOwnership()` after deployment.

- metadata update of new autowrap deployment on Base

- Replaced @nomiclabs/hardhat-etherscan with @nomicfoundation/hardhat-verify for etherscan v2 compatibility.
Since the v1 etherscan API key becomes defunct anyway, the env vars for v2 api keys was renamed to `ETHERSCAN_API_V2_KEY`.
Also updated all hardhat related packages to the latest version compatible with hardhat v2 and ethers v5.


packages/automation-contracts/autowrap/script/addStrategy.js
d10r, line 1: This was simplified (inlined into the deploy script) because if burning the owner keys, there's no future path for adding more strategies.

package.json
```diff
@@ -23,8 +23,8 @@
         "eslint-plugin-eslint-comments": "^3.2.0",
         "eslint-plugin-import": "^2.29.1",
         "eslint-plugin-prettier": "^5.2.1",
-        "hardhat": "^2.22.9",
-        "hardhat-deploy": "^0.12.4",
+        "hardhat": "^2.26.1",
+        "hardhat-deploy": "^1.0.4",
```
d10r, line 27: The switch from v0 to v1 did not come with API breakage according to the changelog

package.json
```diff
@@ -23,8 +23,8 @@
         "eslint-plugin-eslint-comments": "^3.2.0",
         "eslint-plugin-import": "^2.29.1",
         "eslint-plugin-prettier": "^5.2.1",
-        "hardhat": "^2.22.9",
-        "hardhat-deploy": "^0.12.4",
+        "hardhat": "^2.26.1",
```
d10r, line 26: this implies dropped node18 support. But with ethereum-contracts already having dropped that, should be ok.

package.json
```diff
@@ -4,9 +4,9 @@
     "version": "0.0.0",
     "devDependencies": {
         "@nomicfoundation/hardhat-chai-matchers": "^1.0.6",
-        "@nomicfoundation/hardhat-ethers": "^3.0.7",
-        "@nomiclabs/hardhat-etherscan": "^3.1.7",
-        "@nomiclabs/hardhat-web3": "^2.0.0",
+        "@nomicfoundation/hardhat-ethers": "^3.1.0",
```
d10r, line 7: hardhat-ethers v3 requires hardhat v2 and ethers v6.
There's also hardhat-ethers v4, but it requires hardhat v3.
ethereum-contracts package still uses ethers v5, thus has an overriding import of hardhat-ethers v2 which depends on ethers v5.

packages/automation-contracts/default.nix
```diff
@@ -11,7 +11,6 @@
       ./autowrap/package.json
       ./autowrap/contracts
       ./autowrap/test
```
hellwolf, line 13: what about ./deploy/deploy.js

d10r: my understanding here is that the nix env is only used for building and testing (with foundry), while deploy.js is a hardhat based deploy script. Is that correct or should deploy.js be included (anyway)?

hellwolf: okay, if deploy.js is not used in testing flow, then it's not required.

kasparkallas (approved):

PR #2113, opened 2025-10-13
[METADATA] added VestingSchedulerV3 deployments on eth-sepolia and base-sepolia

d10r:
and bumped metadata version


kasparkallas (approved):

d10r:
pinned postgres version to v16 in the docker-compose for local subgraph testing in order to make behaviour more deterministic.
E.g. v18 changed the default data directory.

PR #2114, opened 2025-10-16
[ETHEREUM-CONTRACTS] emit Transfer event on flow create/delete and units changing from/to 0

### LINKED ISSUE #2115: [PROTOCOL] don't emit Approval in transferFrom
## What & Why

Due to the way it's implemented, `transferFrom` does emit an `Approval` event when subtracting the amount being transferred from the allowance. That seems wrong.

Next: check ERC20 spec and reference implementations if this is indeed unexpected behaviour.
If so, check where this is coming from (ERC777 implementation?). Remove.

d10r:
## What & Why

Many indexers (e.g. explorers) rely on the ERC20 `Transfer` event 
a) for managing the set of token holders,.
b) for triggering a balance calculation
With SuperTokens, that can lead to accounts receiving via flows or distributions to remain _invisible_, and for the balance to never be updated.

## How

By emitting Transfer events, we shall be able to somewhat mitigate this issues:
a) by emitting a Transfer event when creating a flow or assigning pool units, the receiving account is included in the set of token holders to be considered
b) by emitting a Transfer event when deleting a flow or resetting pool units, some indexers may trigger a balance check and thereby update the account balance to the correct number

## Other

Also bumped version to 1.14.0 as this may be the last PR before release.

### DIFF
--- a/packages/ethereum-contracts/contracts/agreements/ConstantFlowAgreementV1.sol
+++ b/packages/ethereum-contracts/contracts/agreements/ConstantFlowAgreementV1.sol
@@ -453,12 +453,7 @@ contract ConstantFlowAgreementV1 is
                 ctx, currentContext);
         }
 
-        if (
-            oldFlowData.flowRate == 0 && flowParams.flowRate != 0 ||
-            oldFlowData.flowRate != 0 && flowParams.flowRate == 0
-        ) {
-            flowVars.token.emitZeroTransfer(flowVars.sender, flowVars.receiver);
-        }
+        flowVars.token.emitZeroTransfer(flowVars.sender, flowVars.receiver);
 
         _requireAvailableBalance(flowVars.token, flowVars.sender, currentContext);
     }
@@ -599,6 +594,8 @@ contract ConstantFlowAgreementV1 is
                     newCtx, currentContext);
             }
         }
+
+        flowVars.token.emitZeroTransfer(flowVars.sender, flowVars.receiver);
     }
 
     /**************************************************************************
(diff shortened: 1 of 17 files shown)


hellwolf:
> Question: why does _transferFrom trigger Approval Event?

Is it from the original OZ old  code base of ERC777/ERC20

packages/ethereum-contracts/contracts/agreements/ConstantFlowAgreementV1.sol
```diff
@@ -453,6 +453,8 @@ contract ConstantFlowAgreementV1 is
                 ctx, currentContext);
         }
 
+        flowVars.token.emitZeroTransfer(flowVars.sender, flowVars.receiver);
```
hellwolf: perhaps `PseudoTransfer`

packages/ethereum-contracts/contracts/agreements/gdav1/SuperfluidPool.sol
```diff
@@ -453,6 +453,11 @@ contract SuperfluidPool is ISuperfluidPool, BeaconProxiable {
             _membersData[memberAddr] = _pdPoolMemberToMemberData(pdPoolMember, claimedValue);
             assert(GDA.appendIndexUpdateByPool(superToken, p, t));
         }
+
+        if (oldUnits == 0 && newUnits != 0 || oldUnits != 0 && newUnits == 0) {
```
hellwolf: to be pedantic, `(oldUnits == 0 || newUnits == 0) && oldUnits != newUnits` has fewer characters. but I had a mental calculation, they probably yield same amoutn of operations in most cases.

hellwolf (approved):
lgtm

hellwolf:
to address #2115 too

d10r:
> > Question: why does _transferFrom trigger Approval Event?
> 
> Is it from the original OZ old code base of ERC777/ERC20

answered [here](https://github.com/superfluid-org/protocol-monorepo/issues/2115#issuecomment-3431363571)

PR #2125, opened 2025-12-11
Super Token Yield Backends: Aave & ERC4626

d10r:


.gitmodules
```diff
@@ -5,3 +5,9 @@
 	path = lib/openzeppelin-contracts
 	url = https://github.com/OpenZeppelin/openzeppelin-contracts.git
 	branch = release-v5.4
+[submodule "aave-v3-origin"]
+	path = aave-v3-origin
+	url = https://github.com/aave-dao/aave-v3-origin
+[submodule "lib/aave-v3"]
```
hellwolf, line 8: clean it up.

packages/ethereum-contracts/contracts/interfaces/superfluid/ISuperToken.sol
```diff
@@ -63,13 +63,16 @@ interface ISuperToken is ISuperfluidToken, IERC20Metadata, IERC777, IERC20Permit
      */
     function changeAdmin(address newAdmin) external;
 
+
```
hellwolf, line 66: It is a breaking change to ISuperToken.

packages/ethereum-contracts/contracts/interfaces/superfluid/ISuperToken.sol
```diff
@@ -63,13 +63,16 @@ interface ISuperToken is ISuperfluidToken, IERC20Metadata, IERC777, IERC20Permit
...
     event AdminChanged(address indexed oldAdmin, address indexed newAdmin);
 
     /**
      * @dev Returns the admin address for the SuperToken
      */
     function getAdmin() external view returns (address admin);
```
hellwolf, line 77: More comments here need to clarify what admin can do, and what is the current admins, and how it can and is expected to evolve over time.

packages/ethereum-contracts/contracts/mocks/SuperTokenMock.t.sol
```diff
@@ -64,8 +64,8 @@ contract SuperTokenStorageLayoutTester is SuperToken {
         require (slot == 18 && offset == 0, "_operators changed location");
         // uses 4 slots
```
hellwolf, line 66: add missing slots testing.

packages/ethereum-contracts/contracts/superfluid/SuperToken.sol
```diff
@@ -839,6 +883,14 @@ contract SuperToken is
         uint256 actualUpgradedAmount = amountAfter - amountBefore;
         if (underlyingAmount != actualUpgradedAmount) revert SUPER_TOKEN_INFLATIONARY_DEFLATIONARY_NOT_SUPPORTED();
 
+        if (address(_yieldBackend) != address(0)) {
```
hellwolf, line 896: move to _burn and _mint, so that selfBurn/selfMint could work too (for SETH).

packages/ethereum-contracts/contracts/superfluid/AaveYieldBackend.sol
```diff
@@ -0,0 +1,79 @@
...
+
+    function withdrawMax() external {
+        // We can delegate the max calculation to the Aave pool by setting amount to type(uint256).max
+        AAVE_POOL.withdraw(address(ASSET_TOKEN), type(uint256).max, address(this));
+    }
+
+    function withdrawSurplus(uint256 totalSupply) external {
+        // totalSupply is always 18 decimals while assetToken and aToken may not
+        (uint256 normalizedTotalSupply,) = ISuperToken(address(this)).toUnderlyingAmount(totalSupply);
+        // decrement by 1 in order to offset Aave's rounding up
```
hellwolf: find a reference to make it more precise.

packages/ethereum-contracts/contracts/superfluid/AaveYieldBackend.sol
```diff
@@ -0,0 +1,79 @@
...
+        uint256 surplusAmount = A_TOKEN.balanceOf(address(this)) - normalizedTotalSupply - 1;
```
hellwolf: Take into account the stray balances of the underlying, forwhatever the reason it came about.

packages/ethereum-contracts/contracts/superfluid/SuperToken.sol
```diff
@@ -195,6 +198,47 @@ contract SuperToken is
         }
     }
 
+    function enableYieldBackend(IYieldBackend newYieldBackend) external onlyAdmin {
+        require(address(_yieldBackend) == address(0), "yield backend already set");
+        _yieldBackend = newYieldBackend;
+        (bool success, ) = address(_yieldBackend).delegatecall(
```
hellwolf: consider an internal library boilerplate to hide the cruft away from the main super token logic.

packages/ethereum-contracts/contracts/superfluid/SuperToken.sol
d10r: 2 reasons against that:
- would need to add an `underlyingAmount` argument (or duplicate the calculation)
- in the selfBurn path we want to set a flag which avoids the withdrawn ETH to be wrapped to ETHx

packages/ethereum-contracts/contracts/superfluid/AaveYieldBackend.sol
```diff
@@ -0,0 +1,121 @@
...
+            IWETH(address(ASSET_TOKEN)).deposit{value: amount}();
+        }
+        // Deposit asset and get back aTokens
+        AAVE_POOL.supply(address(ASSET_TOKEN), amount, address(this), 0);
+    }
+
+    function depositMax() external {
+        uint256 amount = USING_WETH ? 
+            address(this).balance :
+            ASSET_TOKEN.balanceOf(address(this));
```
hellwolf: check scaledBalanceOf

```

  /**
   * @notice Returns the scaled balance of the user.
   * @dev The scaled balance is the sum of all the updated stored balance divided by the reserve's liquidity index
   * at the moment of the update
   * @param user The user whose balance is calculated
   * @return The scaled balance of the user
   */
  function scaledBalanceOf(address user) external view returns (uint256);
```

d10r:
## Aave rounding issue

When depositing X assetToken into an Aave pool, X aToken get minted.
When immediately (no time to accrue interest) withdrawing X assetToken from an Aave pool, the burned aToken amount can be X (most of the time) or X+1 (sometimes) or X+2 (rarely).
Here a log of a fork test trying random numbers, using exponential distribution, depositing between 2 (wei) and 100M USDC:
```
  === Iteration 366 ===
  assetAmount requested: 100000000 0
  assetAmount received: 100000000 0
  aTokenAmount decrease: 100000000 1
  diff (requested - received):      0
  diff (received - aTokenDecrease): 1
  Bound result 44
  === Iteration 367 ===
  assetAmount requested: 21572363 539494
  assetAmount received: 21572363 539494
  aTokenAmount decrease: 21572363 539495
  diff (requested - received):      0
  diff (received - aTokenDecrease): 1
  Bound result 1
  === Iteration 368 ===
  assetAmount requested: 0 2
  assetAmount received: 0 2
  aTokenAmount decrease: 0 2
  diff (requested - received):      0
  diff (received - aTokenDecrease): 0
  Bound result 7
  === Iteration 369 ===
  assetAmount requested: 0 174
  assetAmount received: 0 174
  aTokenAmount decrease: 0 175
  diff (requested - received):      0
  diff (received - aTokenDecrease): 1
  Bound result 36
  === Iteration 370 ===
  assetAmount requested: 77869 556422
  assetAmount received: 77869 556422
  aTokenAmount decrease: 77869 556422
  diff (requested - received):      0
  diff (received - aTokenDecrease): 0
  Bound result 19
  === Iteration 371 ===
  assetAmount requested: 0 940007
  assetAmount received: 0 940007
  aTokenAmount decrease: 0 940007
  diff (requested - received):      0
  diff (received - aTokenDecrease): 0
  Bound result 35
  === Iteration 372 ===
  assetAmount requested: 49850 230849
  assetAmount received: 49850 230849
  aTokenAmount decrease: 49850 230849
  diff (requested - received):      0
  diff (received - aTokenDecrease): 0
  Bound result 34
  === Iteration 373 ===
  assetAmount requested: 18569 528966
  assetAmount received: 18569 528966
  aTokenAmount decrease: 18569 528966
  diff (requested - received):      0
  diff (received - aTokenDecrease): 0
  Bound result 8
  === Iteration 374 ===
  assetAmount requested: 0 368
  assetAmount received: 0 368
  aTokenAmount decrease: 0 368
  diff (requested - received):      0
  diff (received - aTokenDecrease): 0
  Bound result 10
  === Iteration 375 ===
  assetAmount requested: 0 1782
  assetAmount received: 0 1782
  aTokenAmount decrease: 0 1784
  diff (requested - received):      0
  diff (received - aTokenDecrease): 2
```

(18569 528966 means 18569.528966 USDC)

### Why does this happen?

Aave uses continuously compounding interest for deposits.
That's a problem not too dissimilar from what the GDA is doing.
While the GDA keeps track of per-user and total `units`, and derives distributed amounts/flowrates proportionally, Aave pools keep track of `shares` and of a `liquidityIndex`.

When an Aave asset pool is initialized, 1 share = 1 assetToken.
Then with time going on and interest accruing, this relationship keeps shifting: 1 share = r*1 assetTokens.
r is the multiplier expressing the principal + interest somebody depositing at day one would have accrued. E.g. the USDC pool on Base now has r = 1.1179...
This number is stored with 27 digits and named `liquidityIndex`. It can be queried via `IPool.getReserveNormalizedIncome()`. E.g.
```
$ cast call --rpc-url https://mainnet.base.org 0xA238Dd80C259a72e81d7e4664a9801593F98d1c5 "getReserveNormalizedIncome(address)" 0x833589fCD6eDb6E08f4c7C32D4f71b54bdA02913 | xargs hex2dec 
1117935072554780784462200675
```
Divided by 1e27 (named `RAY`), this is 1.1179...

The user facing unit is not shares (which keep depegging further and further from the assetToken, but aTokens.
aTokens are always pegged to the assetToken 1:1. That's achieved by continuously scaling (aka rebasing) them with r.

When a user requests withdrawal of x assetTokens, Aave calculates the corresponding shares by dividing by r, rounding up (favoring the protocol).
The rounding error introduced here is currently capped at 1 (wei), but can grow with r.
With r <= 1, the max rounding error is 1, with r <= 2 it's 2 etc.
For USD at current yields, it would take over a decade to transition to an r which allows this rounding error to become 2 (meaning 100% compounded yield since pool creation).

So, why do we already see some cases of the diff between aToken amount and assetAmount being 2?
That's because of a 2nd source of rounding error: when subtracting shares from the total shares held by the user, the protocol is again rounding such that it favours the protocol. This error is capped at 1, regardless of r.

### Economics of griefing attack

So, a user could grief SF by continuously downgrading/withdrawing choosing the amounts such that the (current) max rounding error of 2 wei per operation is lost.
The cost of doing so:
```
> gasPerIteration = 200000
200000
> gasPrice = 0.0005 * 1e9
500000
> ethPrice=3000
3000
> usdPricePerIteration = gasPerIteration * gasPrice * ethPrice / 1e18
0.0003
> lossPerIteration = 2/1e6
0.000002
> costPerUsdLoss = 1/lossPerIteration * usdPricePerIteration
150
> maxGasPerHour = 62500000 * 3600 / 2
112500000000
> maxLossPerHour = maxGasPerHour / gasPerIteration * lossPerIteration
1.125
> yieldPerHour = curTotalSupply * avgAPR / (365 * 24)
1.36986301369863
> minTxCostPerHour = maxGasPerHour * gasPrice * ethPrice / 1e18
168.75
```

**In order to cause a loss of 1 USDC, a griefer would need to spend min 150$ worth of tx fees** with the gas price at 0.0005 gwei ([current floor price of Base](https://base-l2.statuspage.io/incidents/lyhrg67gqn38) - assuming no blob cost).

Note that the griefer would not get the 1 USDC, it would just withhold it from the SF protocol, at a high cost.

Even if somebody would go and do that, this could happen in the following unrealistic, but theoretically possible worst case scenario:
- utilize the full sustained (EIP-1559-targeted utilization) Base gas capacity of 62,500,000 gas per block
- pay just the floor gas price
This assumes there's ZERO other transactions on Base.

In this scenario the griefer could withhold ~1.125 USDC from SF protocol by using all Base block space to accumulate rounding errors with repeated `downgrade` transactions.

And even if this were to happen, it would still not be enough to offset the yield generated by Aave for SF with the current USDCx supply of ~400k.

For asset tokens with 18 decimals (e.g. WETH) the amounts would be smaller by many orders of magnitude.

packages/ethereum-contracts/.solhint.json
```diff
@@ -16,6 +16,16 @@
         "constructor-syntax": "error",
         "func-visibility": ["error", { "ignoreConstructors": true }],
         "quotes": ["error", "double"],
-        "max-line-length": ["error", 120]
+        "max-line-length": ["error", 120],
+	"use-natspec": "off",
```
hellwolf: tab?!

packages/ethereum-contracts/contracts/interfaces/superfluid/IYieldBackend.sol
```diff
@@ -0,0 +1,54 @@
...
+    function enable() external;
+
+    /// Invoked by `SuperToken` as delegatecall.
+    /// Restores the prior state, e.g. by revoking given approvals
+    function disable() external;
+
+    /// Invoked by `SuperToken` as delegatecall.
+    /// Deposits the given amount of the underlying asset into the yield backend.
+    function deposit(uint256 amount) external;
+    /// Invoked by `SuperToken` as delegatecall.
```
hellwolf: mising a new line

packages/ethereum-contracts/contracts/interfaces/superfluid/IYieldBackend.sol
```diff
@@ -0,0 +1,54 @@
...
+
+    /// Invoked by `SuperToken` as delegatecall.
+    /// tranfers the deposited asset exceeding totalSupply of the SuperToken to the preset receiver account
+    function withdrawSurplus(uint256 totalSupply) external;
+
+    /// Invoked by `SuperToken` as delegatecall.
+    /// Returns the amount of the underlying asset currently managed by the yield backend.
+    /// This shall reflect the amount which could currently be withdrawn from the yield backend,
+    /// including the generated yield.
+    function getManagedAmount() external view returns (uint256);
```
hellwolf: this doesn't seem to be used anymore

packages/ethereum-contracts/contracts/interfaces/superfluid/IYieldBackend.sol
```diff
@@ -0,0 +1,54 @@
...
+    /// Restores the prior state, e.g. by revoking given approvals
+    function disable() external;
+
+    /// Invoked by `SuperToken` as delegatecall.
+    /// Deposits the given amount of the underlying asset into the yield backend.
+    function deposit(uint256 amount) external;
+    /// Invoked by `SuperToken` as delegatecall.
+    /// Deposits the maximum amount of the underlying asset into the yield backend.
+    /// Maximum is defined by the underlying asset balance of the SuperToken and the yield backend capacity.
+    function depositMax() external;
```
hellwolf: depositMax/withdrawMax logic is repeated in yield backends. we should rather keep it in the super token logic directly.

packages/ethereum-contracts/contracts/superfluid/ERC4626YieldBackend.sol
```diff
@@ -0,0 +1,67 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.23;
+
+import { IYieldBackend } from "../interfaces/superfluid/IYieldBackend.sol";
+import { IERC20, ISuperToken } from "../interfaces/superfluid/ISuperfluid.sol";
+import { IERC4626 } from "@openzeppelin-v5/contracts/interfaces/IERC4626.sol";
```
hellwolf, line 6: Since it is not SparkSavingsProgram specific, we should name it to ERC2626YieldBackend (Tokenized Vaults).

packages/ethereum-contracts/contracts/interfaces/superfluid/IYieldBackend.sol
d10r: removed `depositMax`.
We should imo keep `withdrawMax` because the way max withdrawable amount is determined is backend specific.

packages/ethereum-contracts/.solhint.json
hellwolf: It still uses tab.

packages/ethereum-contracts/truffle-config.js
```diff
@@ -389,7 +389,7 @@ const E = (module.exports = {
                     runs: 200,
                 },
                 // see https://docs.soliditylang.org/en/latest/using-the-compiler.html#target-options
-                evmVersion: "shanghai",
+                evmVersion: "cancun",
```
hellwolf, line 392: This is a important change that must be stated in the ChangeLog.

packages/ethereum-contracts/contracts/interfaces/superfluid/ISuperToken.sol
hellwolf: Taking a note: we should start to track SuperToken's public API versioning history in place (in this file, directly)

hellwolf: We will start to track super token public api version in another PR.

hellwolf (approved):

PR #2138, opened 2026-01-27
Upload sherlock audit report 2026-01-27

hellwolf:


d10r (approved):
nitpicking: the other audit reports names start with just yyyy-mm (omitting the day), would be more consistent to name this file using the same pattern.

PR #2190, opened 2026-06-15
[WIP] [ETHEREUM-CONTRACTS] v1.15.2

d10r:
- [x]  tracking foundry/forge update: stable version at 1.7.1
- [x] solc update
- [x] forge std lib update

PR #2196, opened 2026-06-25
[ETHEREUM-CONTRACTS] Remove unused canonical wrapper factory APIs

d10r:
The methods for _canonical_ SuperToken wrapper creation are removed because they keep confusing devs (and probably also agents).
This was added with the intention of creating an onchain registry of canonical tokens, potentially alternative to the resolver listings.

Using `createCanonicalERC20Wrapper()` would be possible only after gov owner calling `initializeCanonicalWrapperSuperTokens`. Of the mainnets, this was done only on Ethereum because that happened to be a fresh deployment which didn't require creation of a list of previously deployed SuperTokens considered canonical (grandfathering).
While work on creating such a list for other mainnets did take place, the process eventually stalled because of several reasons, e.g. requirements for a SuperToken to be "canonical" not being well enough defined, and eventually because the feature simply wasn't anticipated enough to justify further investment of resources.

Since the presence of this methods as part of the API continues to mislead devs (_canonical_ sounds like something you may want to use after all), we decided to remove them, which this PR does.
No applications or components relying on this part of the API were found.


hellwolf:
shall we get this merged before #2190 2190

hellwolf (approved):

PR #2227, opened 2026-09-16
[ETHEREUM-CONTRACTS] Cap SuperApp callback returndata and harden ABI bytes parsing

d10r:
Unbounded or invalid returnData of SuperApp callbacks could cause transactions to fail in a way which circumvents the jailing rules.

This PR adds the following safeguards:
* returnData is capped to `CALLBACK_RETURNDATA_CAP` (set to 128 kB). Unless it was caused by the provided `userData` already exceeding it, the App is jailed if exceeding it.
* Added a preliminary check to the input to  `padLength32` which since Solidity 0.8 could panic if invoked with an invalid value.


d10r:
userData affects execution gas throughout the agreement call because it is embedded in ctx:

  ctx.length = 448 + roundUpTo32(userData.length)

  For example, 32,320 bytes of userData produce a 32,768-byte context—the callback input limit.

  During deleteFlow, that payload passes through several encoding and copying steps:

  Host:
    decode userData → encode inner context → encode outer ctx
    → encode CFA call arguments

  CFA:
    copy ctx into memory → encode isCtxValid arguments
    → encode decodeCtx arguments → decode returned Context
    → copy calldata ctx into newCtx

  Nested Host calls:
    copy ctx into memory for validation/hashing
    → decode nested context → encode returned Context

  Callback dispatch and the eventual agreement return add further copies.

  These operations incur copying, ABI-processing, hashing, and memory-expansion costs. EVM memory expansion is charged incrementally according to:

  M(w) = 3w + floor(w² / 512)       // w = memory size in 32-byte words
  expansion cost = M(new) - M(old)



Measurements from a test case doing `deleteFlow` to a SuperApp with ctx pass-through in the callback:
image

Thus the choice of 32k for `CALLBACK_CONTEXT_CAP`.

lib/forge-std
hellwolf, line 1: These changes are slipped in?

hellwolf (approved):

lib/forge-std
d10r: yes, restored now.

PR #2228, opened 2026-09-16
[ETHEREUM-CONTRACTS] Share SuperApp callback gas across before and after hooks

d10r:
## Summary

`CALLBACK_GAS_LIMIT` is currently a **fresh stipend on every hook**. A hostile SuperApp can spend the full budget in `beforeAgreementTerminated` and again in `afterAgreementTerminated`, so a liquidator that provisioned one callback's worth of gas OOGs after the app has already returned success — no jail, stream stays open. That is worse under [EIP-7825](https://eips.ethereum.org/EIPS/eip-7825) (~16.78M tx cap) when the Host default is 15M.

Stacked on https://github.com/superfluid-org/protocol-monorepo/pull/2227 (private review: https://github.com/d10r/sf-protocol-monorepo/pull/6).

- One stipend per SuperApp **before/after pair**. `callAppBeforeCallback` returns unused gas; `callAppAfterCallback` takes that stipend (capped by the Host).
- AgreementLibrary threads the remainder next to `cbdata`. A skipped after-hook drops a local; the next operation starts from a full budget.
- NOOP befores pass `type(uint256).max`; after caps that to `CALLBACK_GAS_LIMIT`. Exhausted remainder is `0`.
- Host, CFA, and IDA must be upgraded together (`ISuperfluid` callback signatures change).


packages/ethereum-contracts/contracts/agreements/AgreementLibrary.sol
```diff
@@ -77,13 +82,14 @@ library AgreementLibrary {
...
         internal
-        returns(bytes memory cbdata)
+        returns(bytes memory cbdata, uint256 remainingCallbackGas)
     {
         bool isSuperApp;
         bool isJailed;
         uint256 noopMask;
         (isSuperApp, isJailed, noopMask) = ISuperfluid(msg.sender).getAppManifest(ISuperApp(inputs.account));
         if (isSuperApp && !isJailed) {
+            remainingCallbackGas = type(uint256).max;
```
hellwolf: Try to remove the local assumption of how before/after callback should work.

e.g. using the pattern of passing current/new values:
```
    function callAppBeforeCallback(
        CallbackInputs memory inputs,
        uint256 currentRemainingCallbackGas,
        bytes memory ctx
    ) returns(bytes memory cbdata, uint256 newRemainingCallbackGas)
```

hellwolf: ~~On second thought, adding another field in ctx will remove a lot of mental overhead for the new code.~~

hellwolf (approved):

